PR #95, opened 2015-01-31
@dzucconi: Admin can link an article to a fair

craigspaeth:
Also some minor cleanups including adding README.mds to components, which I think could be a nice pattern to adopt in Force and others.

cap

### DIFF
--- a/client/apps/edit/components/admin/index.coffee
+++ b/client/apps/edit/components/admin/index.coffee
@@ -4,29 +4,30 @@ request = require 'superagent'
 async = require 'async'
 featuredListTemplate = -> require('./featured_list.jade') arguments...
 sd = require('sharify').data
-Autocomplete = require '../../../../components/autocomplete/index.coffee'
-AutocompleteSelect = require '../../../../components/autocomplete_select/index.coffee'
 
 module.exports = class EditAdmin extends Backbone.View
 
   initialize: ({ @article }) ->
     @article.on 'open:tab2', @onOpen
-    @article.on 'change:fair_id', @renderFairInput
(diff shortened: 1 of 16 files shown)


client/apps/edit/components/admin/index.coffee
```diff
@@ -1,29 +1,48 @@
...
 
-  onAutocompleteSelect: (e, item) =>
+  onAuthorSelect: (e, item) =>
     return unless confirm "Are you sure you want to change the author?"
     @article.trigger('finished').save(author_id: item.id)
 
+  setupFairAutocomplete: ->
+    AutocompleteSelect = require '../../../../components/autocomplete_select/index.coffee'
+    select = AutocompleteSelect @$('#edit-admin-fair .edit-admin-right')[0],
+      url: "#{sd.ARTSY_URL}/api/v1/match/fairs?term=%QUERY"
```
craigspaeth: Unfortunately no fair resource in v2 at the moment so just directly interacting with v1 here :-/

client/components/autocomplete_select/README.md
```diff
@@ -0,0 +1,19 @@
...
+````coffeescript
+select = AutocompleteSelect @$('#input-container')[0],
+  url: "#{sd.ARTSY_URL}/api/v1/match/fairs?term=%QUERY"
+  filter: (res) -> for r in res
+    { id: r._id, value: r.name }
+  selected: (e, item) ->
+    article.save fair_id: item.id
+  cleared: ->
+    article.save fair_id: null
+````
```
dzucconi: :100:

PR #96, opened 2015-02-09
@broskoski: Allow querying articles by fair id + -published_at & save published_at on articles

craigspaeth:
This will allow the year round page in Force to pull articles that are truly featured to the The Armory Show instead of just hardcoding the author they came from.


broskoski:
yeauhhh :+1:

PR #118, opened 2015-02-18
@dzucconi: Denormalize author data into the article on publish

craigspaeth:
It's alright for the client to fetch the author on the fly for an individual article, but for list views like /magazine it would be crazy to blast out a bunch of Gravity fetches to render the author's name. This saves denormalized author data to the article when publishing. I'll follow up with a migration strategy for old posts/articles.


dzucconi:
Sweet; very nice to have indeed.

PR #151, opened 2015-02-25
@dzucconi: Admin can add a user to Writer

craigspaeth:
This gives Editorial the ability to add contributor to Writer. Which will fill a current need for Editorial to author articles under new Artsy users they create (often the case with new fairs). Eventually I want to take a step back and consider a more elegant solution to this user-creating/syncing process.

cap


dzucconi:
I think I remember you telling me about this but remind me—what is the logic behind having a separate users database for writer. Decorating the user records?

craigspaeth:
@dzucconi: Yep exactly. For instance a user might have `writes_for_fairs: [id, id]` to toggle UI in Writer for managing a fair's magazine page. Also in general I've found myself wanting a database to decorate users in other client apps so seemed like a good idea to set out with (e.g. Force could benefit from things like `seen_set_management_banner_times: 4`).

However, in the end these smaller use cases may not be enough to justify the edge cases involved in keeping them in sync with Gravity. Something I have to ponder on more :thought_balloon:

Issue #122, opened 2015-02-18
Can associate partners to articles


craigspaeth:
In prep for this: https://github.com/artsy/force/issues/2060

Might not be as straightforward as assigning a fair in admin settings b/c we want partner users to be able to do this too.

craigspaeth:
Chatted with @halleyjohnson. For v1 keep it minimal w/ small feature for admin.

* On publish: determine the user's ACL and tie the article to the PP
* Admin UI that allows admins to explicitly tie articles to PPs

Other notes:
* Partner articles are often 1 or 2 recent/relevant articles and a backlog of really old
  * Treat most recent article as tier 1 no matter what?
  * Generally partner wants to promote the most recent article over anything
* Currently PP is tied to user
* Would like more control over order in fairs, maybe PP eventually
* Removing from PP is a pain currently (not an issue with Writer)

craigspaeth:
- [ ] Implement Admin UI to associate articles to a partner
- [ ] Refactor users to use API v1 until [this](https://groups.google.com/forum/#!topic/artsy-api-developers/nmRcYjVFOnI) is addressed
- [ ] Clean up user schema `{ _id, name, email, type, access_to_partner_ids, profile_handle, profile_id, profile_icon_url, access_token: (encrypted) }`
- [ ] Refactor user API to allow an admin to search for a user from v1. On switch user add them to the database.
- [ ] Drop sync users script for using the migration script to copy users from the gravity database

PR #155, opened 2015-02-26
@broskoski: An admin can add an article to a partner's profile

craigspaeth:
For now just re-uses the fair assigning UI to associate articles to partners. I need to address [some](https://github.com/artsy/positron/issues/122#issuecomment-76294392) larger refactors for Positron users so that when a user with access to a partner publishes an article, it automatically get associate to that partner's profile.

For the time being, admins do this featuring manually for the most part anyways, so this will suffice for shipping Magazine.

cap


broskoski:
:+1: Positron is looking so slick!

PR #164, opened 2015-03-02
@broskoski: Sync to Post features to artist & artwork pages

craigspaeth:
Currently Editorial has to "Sync to Post" then go and feature through the old UI. This moves more of their workflow exclusively inside Writer. Also adds some random cleanups here and there.


broskoski:
Nice :ok_hand:

PR #209, opened 2015-03-12
@craigspaeth use quicker bytitle artwork search

kanaabe:
I wanted to point out that there are no thumbnails in this search since it looks like typeahead.js doesn't have the ability to edit the html of the results.

Results are MUCH quicker but I am also doing a few strange things that could probably be refactored :sweat_smile: 

artwork_search_positron2


kanaabe:
Right, tests need to be added/fixed too

Also, I'm getting a bunch of errors like this: 
`Uncaught Error: Invariant Violation: ReactMount: Two valid but unequal nodes with the same data-reactid: .2.0.1:6:0.1.0.1.1.1.0`

I think it's a matter of removing the selection once it's added but I can't seem to debug it. If you have any ideas, please feel free to chime in ;)

client/apps/edit/components/section_artworks/index.coffee
```diff
@@ -27,13 +27,35 @@ module.exports = React.createClass
...
 
+  setupAutocomplete: ->
+    Autocomplete = require '../../../../components/autocomplete/index.coffee'
+    @autocomplete = new Autocomplete
+      url: "#{sd.ARTSY_URL}/api/search?q=%QUERY"
+      el: $(@refs.autocomplete.getDOMNode())
+      filter: (res) ->
+        vals = []
+        for r in res._embedded.results
+          if r.type == 'Artwork'
```
craigspaeth: Oh nice! I didn't even think v2's search would work here given it returns mixed results. That's exciting!

client/apps/edit/components/section_artworks/index.coffee
```diff
@@ -27,13 +27,35 @@ module.exports = React.createClass
...
+            id = r._links.self.href.substr(r._links.self.href.lastIndexOf('/') + 1)
```
craigspaeth: Oh this is one reason why I didn't want to use v2... you can't get the actual ids, or any of the first class artwork data, from the results. Although I see that you've cleverly worked around this by fetching the artwork in `onSelect`.

This works well for me but an alternative is to just use v1's `https://api.artsy.net/api/v1/match/artworks?term=`

client/apps/edit/components/section_artworks/index.coffee
```diff
@@ -123,10 +148,10 @@ module.exports = React.createClass
           (@props.section.artworks.map (artwork, i) =>
             li { key: i },
               div { className: 'esa-img-container' },
-                img { src: artwork.get('image_urls')?.large }
+                img { src: artwork.get('image_urls')?.large or artwork.attributes.images[0]?.image_urls.large }
```
craigspaeth: In general you should use `set` and `get` instead of `.attributes` so `artwork.get('images')[0]...`. In this case I don't mind b/c this illuminates a larger refactor I want to do which is drop the orchestration stuff in favor of just v1 data models.

craigspaeth:
I think you can actually use html in the results of typeahead, but it's not well documented—not necessary for now though. I'll come by and help debug the tests and then this will be ready to merge.

craigspaeth:
w00t w00t :green_heart:

kanaabe:
Thanks for your patience and help on this one!

Issue #216, opened 2015-03-16
Unable to edit article in writer


mollygottschalk:
I'm experiencing an issue where I'm unable to add new items to an article in writer.

When I got to add a new item (text or an artwork, for example) I hover my cursor over the place where I want the plus sign to appear. When I click the plus sign, the drop down menu shows up and then I choose, say, "text." However, at that point, the text box that already exists below pops up, giving me the option to edit the existing text, rather than add new text in a new text box. 

I have tried logging in and out, and working in an incognito window, but have not had any luck in solving the issue.

mollygottschalk:
@craigspaeth

craigspaeth:
Thanks @mollygottschalk. I think I've actually seen this before too, can you paste a screenshot of what it looks like when it focuses on the text section that already exists.

mollygottschalk:
https://writer.artsy.net/articles/5506e5febea07e060091c0ea/edit
![screen shot 2015-03-16 at 4 07 54 pm](https://cloud.githubusercontent.com/assets/3266511/6675421/a6321816-cbf6-11e4-8253-abb5674615aa.png)

mollygottschalk:
@craigspaeth

Issue #214, opened 2015-03-13
Using H2 with a click doesn't save


craigspaeth:
It updates the UI but what actually gets saved isn't wrapped in an H2. Strangely using the cmd + 2 shortcut works.

PR #217, opened 2015-03-16
@broskoski: Re-render text sections only when not editing

craigspaeth:
I set `shouldComponentUpdate` to return false for text sections b/c React & Scribe don't play very nicely together, so I though I would just take a sledge hammer to the re-renders in text sections. This fixed a lot of issues with weird things jumping around while trying to edit text sections, but unsurprisingly caused other bugs with surrounding text sections, see: https://github.com/artsy/positron/issues/216.

After toying around a bit the secret sauce seems to be to not let the component re-render while you're in "editing" mode. So now `shouldComponentUpdate` checks against whether you're in that editing mode.


craigspaeth:
Also PR creeped with this https://github.com/artsy/positron/issues/214, which was just a matter of adding another event handler.

PR #218, opened 2015-03-17
@kanaabe: Ensure profile icon

craigspaeth:
This adds some logic to better choose a potentially working profile icon, as well as replaces the icon on the client-side if the image url is broken. Like I mentioned in the issue, ideally we just rely on the API for a consistently working image url but I have other user refactors I want to address with that later.


kanaabe:
eek jpg -> png

Issue #221, opened 2015-03-17
Indeterminate editing issues


halleyjohnson:
Had strange syncing issues with a couple of articles in the last two days. Currently working on this post: https://www.artsy.net/article/artsy-editorial-the-top-10-chinese-artists-not-named-ai (in top of Editorial's Writer account) and artworks are not showing up the same in Writer and on the live page. Either they're not appearing in Writer but show up on the front end (like the first module here) or they are not appearing at all (like the second). Yesterday had similar issues with this post: https://www.artsy.net/article/editorial-a-supercharged-art-dubai-becomes-the-hotbed and also had a lot of text blocks duplicating/switching places or disappearing. Sorry this isn't more specific, just not sure what exactly is the trigger here. A lot of artworks, maybe?

in Writer:

![screen shot 2015-03-17 at 17 30 34](https://cloud.githubusercontent.com/assets/10982707/6698069/d4646d9e-cccb-11e4-89b9-3f7cf7c286e1.png)

on site:

![screen shot 2015-03-17 at 17 30 42](https://cloud.githubusercontent.com/assets/10982707/6698071/d7556198-cccb-11e4-95a6-198c7db61896.png)

craigspaeth:
This is fixed in production now.

PR #250, opened 2015-03-20
@kanaabe: Drop orchestration stuff for artworks and just use v1

craigspaeth:
Fixes this https://github.com/artsy/positron/issues/221 by just punting and using API v1 for artworks.

Reasoning:
* Ideally this orchestration layer stuff exists in it's own app so other web apps can use it anyways
* We already had to drop v2 for autocomplete b/c it was too slow without getting more structured data returned from the search endpoint (https://github.com/artsy/positron/pull/209)
* We're also already using v1 for admin autocompletes that don't exist in v2 (e.g. match endpoints)
* In the end v2 is just not up to speed yet to try to solely depend on it in Writer, but it was a nice experiment while it lasted.


kanaabe:
Alright!

dblock:
Thanks for trying out v2, Craig! Lots of features were put in there because of this effort, so it's all a plus. Please open anything from here that you think V2 needs.

Issue #231, opened 2015-03-18
Thumbnail view updates


craigspaeth:
Will follow up with the browser graphic asset. For now just use what's already there.

* Remove teaser & tag inputs
* Title textarea font-size: 24px line-height: 28px

![image](https://cloud.githubusercontent.com/assets/555859/6718324/4d40bce6-cd89-11e4-8e90-df0976e77e05.png)

![image](https://cloud.githubusercontent.com/assets/555859/6718335/61e4a1da-cd89-11e4-84b4-f3c71ac60206.png)

PR #251, opened 2015-03-20
@craigspaeth Updates thumbnail editor

kanaabe:
Also removed a couple tests and code around 'tags' and 'teasers'. 

positron-input

### DIFF
--- a/client/apps/edit/components/layout/index.coffee
+++ b/client/apps/edit/components/layout/index.coffee
@@ -19,7 +19,6 @@ module.exports = class EditLayout extends Backbone.View
     @article.on 'missing', @highlightMissingFields
     @article.on 'finished', @onFinished
     @article.on 'loading', @showSpinner
-    @article.on 'finished open:tab1', @syncTitleTeaser
     @article.once 'sync', @onFirstSave if @article.isNew()
     @article.sections.on 'change:layout', => _.defer => @popLockControls()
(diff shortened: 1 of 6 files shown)


kanaabe:
Fixes https://github.com/artsy/positron/issues/231

craigspaeth:
Awesome thanks!

client/apps/edit/components/layout/index.coffee
```diff
@@ -73,8 +64,8 @@ module.exports = class EditLayout extends Backbone.View
     }
 
   toggleAstericks: =>
-    @$('.edit-required + :input').each (i, el) =>
-      $(el).prev('.edit-required').attr 'data-hidden', !!$(el).val()
+    @$('.edit-required ~ :input').each (i, el) =>
+      $(el).prevAll('.edit-required').attr 'data-hidden', !!$(el).val()
```
kanaabe: Do you see this being a problem at all? The "Use article title" button was added between the `edit-required` and `input` elements which caused this part to break unless I used prevAll. Not sure if there's a better solution or if this affects anything else.

craigspaeth: I don't see it being a problem. If it causes issues down the line we can be more explicit about targeting these asterisks rather than the magical searching nearby that this does.

PR #252, opened 2015-03-21
@kanaabe: Use cid as the "key" and start updating for the latest React

craigspaeth:
There's a bug where adding an image section between other images would render the state of an adjacent component. I eventually deduced it's because I wasn't using the `key` property to identify the components (something React was throwing a warning for). So turns out we should be taking warning from React seriously :-P... this also spurred my attempt at addressing this warning `Warning: Something is calling a React component directly. Use a factory or JSX instead. See: http://fb.me/react-legacyfactory` but it seems there's still some place where `React.createFactory` needs to be used.


client/apps/edit/components/section_list/index.coffee
```diff
@@ -44,6 +44,7 @@ module.exports = React.createClass
               index: i
               editing: @state.editingIndex is i
               ref: 'section' + 1
+              key: section.cid
```
craigspaeth: Initially I tried using `i` but that didn't help as it wasn't a property that carried with the section as it moved around. So good to know that this key needs to be unique among the array, _as well_ as specific to the data you're passing into the component.

Issue #275, opened 2015-03-23
"use article title" not working


halleyjohnson:
I've tried to run this through a few times with no success: clicking "use article title" on the thumbnail page for this post (https://www.artsy.net/article/editorial-tantalizing-fashion-with-an-ominous-twist-sergey) and it doesn't show up upon saving.

![screen shot 2015-03-23 at 12 27 16](https://cloud.githubusercontent.com/assets/10982707/6784756/f78c82a8-d157-11e4-96b9-7afcea7fbe6b.png)

![screen shot 2015-03-23 at 12 27 38](https://cloud.githubusercontent.com/assets/10982707/6784761/03ff6d34-d158-11e4-9a84-fd3040ec396c.png)

It worked when I pasted in the title, thinking it may need the cursor to be in the box to jog the autosave? Does that make sense?

PR #278, opened 2015-03-23
@craigspaeth Fixing use article button in thumbnail view

kanaabe:
Closes https://github.com/artsy/positron/issues/275

Autosaves after clicking the Use Article Title button.


craigspaeth:
Nice! :+1:

Issue #215, opened 2015-03-16
Associate Articles with Auction Pages


katarinabatina:
Related to https://github.com/artsy/force/issues/2180

cc @dzucconi

craigspaeth:
We can just add another admin autocomplete-select like fairs and partners

![image](https://cloud.githubusercontent.com/assets/555859/6789043/d4dd0194-d173-11e4-902a-adebb5c7e51d.png)

Down the line maybe we can consider having Ohm/Waves hook into the Writer API to centralize the "fair/auction" building experience.

craigspaeth:
@kanaabe: You will also need to support this on the Writer API level which should be as simple as adding a Joi validation param [here](https://github.com/artsy/positron/blob/master/api/apps/articles/model.coffee#L66) and [here](https://github.com/artsy/positron/blob/master/api/apps/articles/model.coffee#L79).

PR #279, opened 2015-03-23
@craigspaeth Allow admin to add auction_id to an article

kanaabe:
Autosaves when you click on the auction and retrieves the auction when you come back to page. 

positron_auction_autocomplete

### DIFF
--- a/api/apps/articles/model.coffee
+++ b/api/apps/articles/model.coffee
@@ -65,6 +65,7 @@ schema = (->
   featured_artwork_ids: @array().items(@objectId()).allow(null)
   fair_id: @objectId().allow('', null)
   partner_ids: @array().items(@objectId()).allow(null)
+  auction_id: @objectId().allow('', null)
   featured: @boolean().default(false)
 ).call Joi
 
@@ -77,6 +78,7 @@ querySchema = (->
   artwork_id: @objectId()
   fair_ids: @array()
   partner_id: @objectId()
+  auction_ids: @array()
   sort: @string()
   tier: @number()
   featured: @boolean()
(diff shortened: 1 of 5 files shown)


api/apps/articles/model.coffee
```diff
@@ -65,6 +65,7 @@ schema = (->
   featured_artwork_ids: @array().items(@objectId()).allow(null)
   fair_id: @objectId().allow('', null)
   partner_ids: @array().items(@objectId()).allow(null)
+  auction_id: @objectId().allow('', null)
```
kanaabe: Better to rename to sale_id?

kanaabe: Also, a user should only be allowed to add it to a single auction, correct?

client/apps/edit/components/admin/test/index.coffee
```diff
@@ -22,6 +22,7 @@ describe 'EditAdmin', ->
         EditAdmin::setupAuthorAutocomplete = sinon.stub()
         EditAdmin::setupFairAutocomplete = sinon.stub()
         EditAdmin::setupPartnerAutocomplete = sinon.stub()
+        EditAdmin::setupAuctionAutocomplete = sinon.stub()
```
kanaabe: I didn't see tests for the other ones, but I'm happy to write them if you see fit!

kanaabe:
Closes https://github.com/artsy/positron/issues/215

api/apps/articles/model.coffee
craigspaeth: Good point... I'm not strongly opinionated on that naming. I like to optimistically think we'll some day refactor auctions to a first class thing, so I'm leaning towards `auction_id`.

For now yeah single auction is fine. We may need to make this a many to many later down the line... but I'll also probably have to address that with fairs—I'll cross that bridge when we get to it.

client/apps/edit/components/admin/test/index.coffee
craigspaeth: Nah in this case we'd just be testing a bunch of configuration options—not a lot of code to test there.

craigspaeth:
Sweet, thanks! Pretty straight forward :smile_cat:

PR #286, opened 2015-03-27
@craigspaeth Published_at and slug changer

kanaabe:


api/apps/articles/model.coffee
```diff
@@ -21,13 +21,15 @@ request = require 'superagent'
 schema = (->
   author_id: @objectId().required()
   tier: @number().default(2)
-  slug: @string().allow(null)
+  slug: @string().allow('', null)
+  slugs: @array().items(@string()).allow('',null,[])
```
craigspaeth: We shouldn't be allowing the client to change `slugs`. B/c this retains history I'd rather this be protected behind the API. (e.g. Orta might want make a native Writer app and we don't want him abusing this power :smiley_cat: )

api/apps/articles/model.coffee
```diff
@@ -161,24 +163,26 @@ validate = (input, callback) ->
   Joi.validate whitelisted, schema, callback
 
 update = (article, input, callback) ->
-  input.published_at = new Date if input.published and not article.published
-  article = _.extend article, input, updated_at: new Date
-  User.find article.author_id, (err, author) ->
+  input.published_at = new Date if(input.published and not article.published and not input.published_at)
```
craigspaeth: Nice!

api/apps/articles/model.coffee
```diff
@@ -161,24 +163,26 @@ validate = (input, callback) ->
...
+    slug = input.slug
   else if author
     slug = _s.slugify(author.user.name) + '-' + titleSlug
   else
     slug = titleSlug
-  article.slugs ?= []
-  article.slugs.push slug unless slug in article.slugs
+  article.slugs.push slug
+  article.slugs = _.uniq(article.slugs.reverse())
+  article.slugs.reverse()
```
craigspaeth: Not sure reversing twice here is necessary. Seems like you add the new slug to the end, then reverse the whole thing so it's added to the beginning. Might be more straightforward to just `article.slugs = [slug].concat _.uniq(article.slugs)`.

client/apps/edit/components/admin/index.coffee
```diff
@@ -152,3 +157,42 @@ module.exports = class EditAdmin extends Backbone.View
...
+    else
+      $('.edit-admin-slug-generate').show()
+
+    $('.edit-admin-slug-input').on 'keyup', _.debounce( =>
+      @slugify()
+      @article.save slug: $('.edit-admin-slug-input').val()
+    , 1000 )
+
+  slugify: ->
+    t = $('.edit-admin-slug-input')
```
craigspaeth: Nitpick, but I prefer to preface jQuery elements with dollar sign, e.g. `$t = ` for clarity.

client/apps/edit/components/admin/index.coffee
```diff
@@ -152,3 +157,42 @@ module.exports = class EditAdmin extends Backbone.View
...
+    slug = _s.slugify(t.val())
+    t.val slug
+
+  generateSlugFromTitle: ->
+    cat = [@article.get('author').name, @article.get('title')].join('-')
```
craigspaeth: Hmm seeing this duplicated logic server/client makes me wonder if the API should just let the client decide the slug and throw a validation error if a client tries to save an article without one... :smile_cat: :thought_balloon:  (out of scope for this PR)

craigspaeth:
Awesome! Looks good! The only comment that's a blocker for me is the first one b/c I wouldn't want the client accidentally clobbering slug history.

kanaabe:
One sec, just want to change a couple details you pointed out

craigspaeth:
:green_heart:

kanaabe:
YES! :green_heart: :green_heart: :green_heart: 

It should be noted that the "main" slug will change if the title is changed. Of course we still save old slugs so they are reachable. Only an Admin can customize the slug. If a title is changed after an admin customizes, the slug will change back to the artist + new title slug

PR #290, opened 2015-03-29
@craigspaeth (cc @kanaabe): New Thumbnail view image + copy

craigspaeth:
image


craigspaeth:
Note to self: Clean this image up. For one, it's anti-aliasing is raggedy and probably an opportunity to better line things up or take a step back and think of a different layout.

PR #325, opened 2015-04-07
@kanaabe: Remove migration & sync to post junk

craigspaeth:
Now that we're switched over this stuff isn't necessary.


client/lib/middleware.coffee
```diff
@@ -17,9 +18,9 @@ uaParser = require('ua-parser')
 @locals = (req, res, next) ->
   res.locals.sd.URL = req.url
   res.locals.sd.USER = req.user?.toJSON()
-  res.locals.sd.USER_HASH = require('crypto')
-    .createHmac('sha256', INTERCOM_SECRET).update(req.user.get('id'))
-    .digest('hex') if req.user
+  console.log INTERCOM_SECRET
```
kanaabe: Stray log

craigspaeth: Whoops! Good catch, looks like some of my fussing with intercom snuck in here.

kanaabe:
Awesome! Great to see this stuff finally moved out.

Issue #312, opened 2015-04-01
Click 'Save' button on article doesn't save slug and published date


kanaabe:
It auto-saves correctly but clicking Save again after the auto-save happens reverts it back to whatever value it previously was.

PR #326, opened 2015-04-07
@craigspaeth Fixes a small case in slug saving

kanaabe:
Closes https://github.com/artsy/positron/issues/312

And generally: 
Slug is changed if Admin changes it
If the article is unpublished, the slug will be generated by the author and title (unless an Admin input is present as well)
Slug does not change automatically if article is published

### DIFF
--- a/api/apps/articles/model.coffee
+++ b/api/apps/articles/model.coffee
@@ -181,12 +181,13 @@ update = (article, input, callback) ->
 addSlug = (article, input, author, callback) ->
   titleSlug = _s.slugify(article.title).split('-')[0..7].join('-')
   article.slugs ?= []
+  #Don't change the article slug unless it's unpublished or a new slug is added
   if input.slug? and (input.slug != _.last(article.slugs))
     slug = input.slug
-  else if author
-    slug = _s.slugify(author.user.name) + '-' + titleSlug
+  else if article.published is false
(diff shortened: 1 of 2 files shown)


api/apps/articles/model.coffee
```diff
@@ -181,12 +181,13 @@ update = (article, input, callback) ->
...
     slug = input.slug
-  else if author
-    slug = _s.slugify(author.user.name) + '-' + titleSlug
+  else if article.published is false
+    slug = if author then _s.slugify(author.user.name) + '-' + titleSlug else titleSlug
   else
-    slug = titleSlug
+    return article
   article.slugs = _.unique(article.slugs).concat [slug]
   article
```
kanaabe: Previously, when an Admin tried to click the 'Save' button in a published article after 'autosave' happened, it would generate and set a slug from the title because this method thought the second save was a duplicate from `(input.slug != _.last(article.slugs)` which caused it to generate the slug from author/title and didn't consider that it was a published article. 

Sorry, that was a mouthful.

craigspaeth:
w00t, thanks! I'm thinking of working out a simpler slug system with Halley but I'm glad you addressed this for the time being :)

kanaabe:
Ok cool, was worth diving into, such a crazy-world-of-slugs

Issue #282, opened 2015-03-25
artworks not appearing in feature section automatically


halleyjohnson:
It seems that this is working on some posts but not others? A few examples below.

Here I had to insert them manually:
https://writer.artsy.net/articles/5512c58b94e4f206008ae18d/edit

Here nothing is loading:
![screen shot 2015-03-25 at 10 42 21 am](https://cloud.githubusercontent.com/assets/10982707/6827072/d086c30a-d2db-11e4-9d44-13cb5ef21c51.png)
https://writer.artsy.net/articles/5511588280e2a60600108c0a/edit

Here everything loaded fine:
![screen shot 2015-03-25 at 10 43 30 am](https://cloud.githubusercontent.com/assets/10982707/6827118/0cbbb254-d2dc-11e4-95cc-a3aabe7bc266.png)
https://writer.artsy.net/articles/551051e6cf9092060094c0c7/edit

kanaabe:
The issue is that some links in text sections (possibly from a Google doc?) do not have pure Artsy hrefs. They are Google redirects like this: 
https://www.google.com/url?q=https%3A%2F%2Fwww.artsy.net%2Fartist%2Ftrenton-doyle-hancock&sa=D&sntz=1&usg=AFQjCNHh8QsVzsZi0sKx8thbzm-GFPq2Mw

The method `slugsFromHTML` in the [section model](https://github.com/artsy/positron/blob/master/client/models/section.coffee#L17) sees that this link contains the string 'artist' or 'artwork' and then assumes it is formatted and incorrectly sets the slug as some encoded url. 

For the immediate fix I'll get the redirect from these Google links so they can be parsed correctly. Going forward though, is it possible to stop adding these ugly links when we convert Google docs to Positron? Or is it something we just have to deal with? @craigspaeth

halleyjohnson:
Hey guys, just wanted to note that this doesn't make sense for the artworks not working. I pulled the links directly from Artsy for the ones in this post: https://writer.artsy.net/articles/552810d1157367060053eea7/edit

And they're not showing up in the Admin page:
![screen shot 2015-04-10 at 14 22 04](https://cloud.githubusercontent.com/assets/10982707/7094137/fd06d678-df8c-11e4-98b5-de287e3eb014.png)

kanaabe:
Hi Halley, sorry, I'm not sure I follow. This is what I see on the Admin panel of this article: 

[...]

PR #330, opened 2015-04-09
@craigspaeth Handles Google redirect for featured/mentioned section

kanaabe:
Closes https://github.com/artsy/positron/issues/282


craigspaeth:
Good catch! I think it would be good to have a test here. If it's a PITA don't worry about it, but hopefully should be straightforward.

kanaabe:
Oh true, definitely!

craigspaeth:
:+1:

Issue #341, opened 2015-04-16
artist_statement.pdf link mis-interpreted as artist url


craigspaeth:
http://people.reed.edu/~miyos/artist_statement.pdf was attempting to be fetched as an artist id.

craigspaeth:
[This](https://github.com/artsy/positron/blob/63d77bdc74c76b29bd2b8bef4cd6c5f51d866249/client/models/section.coffee#L20) is too loose, should check for artsy.net in there.

craigspaeth:
To reproduce, add a link with this url in a text section: http://people.reed.edu/~miyos/artist_statement.pdf then go to the admin tab and watch the mentioned stuff just spin b/c it tries to request `id[]=artist_statement.pdf`.

craigspaeth:
Should also ideally refactor this out of the Writer API/orchestration logic and just fetch directly from v1 like we did with artworks, then if one of the parallel fetches 404s we just ignore it and render anyways.

PR #345, opened 2015-04-17
@craigspaeth Fixes mis-interpreted URLs and rips out Artist orchestration layer

kanaabe:
Closes https://github.com/artsy/positron/issues/341
Goodbye artist orchestration layer, now using v1 only
Fixes small bug with loading spinner on featured artists/artworks section


craigspaeth:
Wohoo! Very nice! Thanks!

client/apps/edit/components/admin/test/index.coffee
```diff
@@ -64,7 +64,7 @@ describe 'EditAdmin', ->
       @view.featureFromInput('Artists')(
         currentTarget: $ "<input value='/andy-warhol' />"
       )
-      Backbone.sync.args[0][2].data.ids[0].should.equal 'andy-warhol'
+      Backbone.sync.args[0][1].id.should.equal 'andy-warhol'
```
kanaabe: We aren't using the `ApiCollection` anymore for artist so I'm just testing against the request's `id` param

craigspaeth: :+1: works for me.

client/collections/artworks.coffee
```diff
@@ -18,6 +18,7 @@ module.exports = class Artworks extends Backbone.Collection
       new Artwork(id: id).fetch
         error: options.error
         success: (artwork) -> cb null, artwork
+        complete: options.complete
     , (err, artworks) =>
```
kanaabe: The `complete` method was missing here which allows the `featureFromInput` in the mentioned/featured artworks section to be called back. Was previously holding on to the loading spinner.

craigspaeth: Ah ha! Good catch!

PR #350, opened 2015-04-21
@kanaabe: Accommodate query params when extracting artist/artwork slugs from text sections 

craigspaeth:
@halleyjohnson Found another bug where the mentioned stuff would just spin b/c of it getting caught on fair microsite query params.

### DIFF
--- a/client/collections/mixins.coffee
+++ b/client/collections/mixins.coffee
@@ -3,6 +3,7 @@
 #
 
 _ = require 'underscore'
+async = require 'async'
 
 @ApiCollection =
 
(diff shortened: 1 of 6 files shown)


client/collections/mixins.coffee
```diff
@@ -10,18 +11,18 @@ _ = require 'underscore'
...
\ No newline at end of file
+    @reject (model) => model.id in col.pluck 'id'
+
+  getOrFetchIds: (ids, options = {}) ->
+    async.map ids or [], (id, cb) =>
+      new @model(id: id).fetch
+        error: (m, err) -> cb null, err
+        success: (model) -> cb null, model
+    , (err, models) =>
+      options.complete?()
```
craigspaeth: I refactored `getOrFetchIds` to DRY it up into the mixin. I think the way async was being used in tandem with options was causing the `complete` callback not get called properly, so I moved the callbacks until after all of the fetched/failed calls.

kanaabe: :+1:

Issue #156, opened 2015-02-27
Positron user refactors


craigspaeth:
- [x] Refactor users to use API v1
- [x] Clean up user schema `{ _id, name, email, type, access_to_partner_ids, profile_handle, profile_id, profile_icon_url, access_token: (bcrypted) }`
- [x] Save a user's `access_to_partner_ids.0` to an article's `partner_ids.0` on publish
- [x] Remove contributors panel in favor of autocomplete just pulling from Artsy v1 and allowing impersonate

craigspaeth:
````
{
  _id
  name
  email
  type
  access_to_partner_ids
  profile_handle
  profile_id
  profile_icon_url
  access_token: (bcrypted)
  access_token_expires_at:
}
````
* Client gets access token
* Client requests /users/me and gets back schema
* Server
  find user by bcrypted token
  if now > access_token_expires_at
    fetch /api/v1/user/me and update record
  else
    asks v1 for data and stores schema
* Migration:
  run script that iterates through user records and saves as new schema

PR #366, opened 2015-05-01
@craigspaeth (cc @kanaabe): Large user refactor

craigspaeth:
Closes https://github.com/artsy/positron/issues/156

The Positron API now does less stuff with users and holds a smaller + flatter user document while the Positron client is responsible for all authing and holding access tokens (essentially making the Positron users collection just a cache, but still allows us to decorate user docs in the future).

Along-side this shift of responsibilities we're dropping v2 orchestration-ey things in favor of the client searching API v1 for users. This combined with the fact that "impersonate" will sync a user to Positron on the fly allows us to drop the silly contributors panel b/c for all intents and purposes Artsy users are Positron users.

In short I'm hoping these changes will make the whole Positron user situation simpler and safer.


kanaabe:
Awesome!

PR #372, opened 2015-05-03
@craigspaeth (cc @kanaabe): Can add cover images to videos

craigspaeth:
cap


api/apps/verticals/model.coffee
```diff
@@ -24,6 +24,7 @@ schema = (->
   description: @string().allow('', null)
   slug: @string().allow('', null)
   partner_logo_url: @string().allow('', null)
+  partner_website_url: @string().allow('', null)
```
craigspaeth: UBS logos need to link somewhere.

Issue #388, opened 2015-05-08
Allow access_token in query params


craigspaeth:
It's just a PITA to play with the API otherwise.

Issue #153, opened 2015-02-26
Refactor article routes so that auth isn't necessary for public articles


craigspaeth:
For /articles?published=true and for `req.article.published` we don't need to find the user whether they pass an access token or not.

PR #389, opened 2015-05-12
@kanaabe: Random auth things

craigspaeth:
https://github.com/artsy/positron/issues/388 and https://github.com/artsy/positron/issues/153


kanaabe:
Ok cool, this looks good!

PR #405, opened 2015-05-27
@dzucconi: Associate articles as artist bios

craigspaeth:
Another little admin UI autocomplete to indicate an article is the bio of an artist.

image


dzucconi:
Sweet. Are there any real bios actually ready to go?

craigspaeth:
Actually not sure @halleyjohnson do we have any artist bios in Writer yet?

halleyjohnson:
Just Andy Warhol, but I need to double-check before we associate that one as a bio.

Issue #408, opened 2015-05-27
Clicking autocomplete select seems to fail


craigspaeth:
I noticed a React error in the JS console here.

PR #413, opened 2015-05-29
@craigspaeth Defers before setting state

kanaabe:
A heavy single line of code indeed. Thanks @craigspaeth for the help! 

Closes https://github.com/artsy/positron/issues/408

### DIFF
--- a/client/components/autocomplete_select/index.coffee
+++ b/client/components/autocomplete_select/index.coffee
@@ -25,7 +25,7 @@ module.exports.AutocompleteSelect = AutocompleteSection = React.createClass
     @autocomplete = new Autocomplete _.extend _.pick(@props, 'url', 'filter'),
       el: $(@refs.input?.getDOMNode())
       selected: (e, item) =>
-        @setState value: item.value, id: item.id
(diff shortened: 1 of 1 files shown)


craigspaeth:
All you @kanaabe! You found the issue—I just happen to know that when in doubt `_.defer`! ( -___- )

client/components/autocomplete_select/index.coffee
```diff
@@ -25,7 +25,7 @@ module.exports.AutocompleteSelect = AutocompleteSection = React.createClass
     @autocomplete = new Autocomplete _.extend _.pick(@props, 'url', 'filter'),
       el: $(@refs.input?.getDOMNode())
       selected: (e, item) =>
-        @setState value: item.value, id: item.id
+        _.defer => @setState value: item.value, id: item.id
```
craigspaeth: Would probably be good to leave a comment here—something simple like # Defering because of click race condition to avoid accidentally removing it later.

craigspaeth:
w00t w00t! Epic detective work!

PR #414, opened 2015-05-29
@kanaabe: Two minor bugs

craigspaeth:
* autocomplete selects in admin would jump b/c forgot to nudge the selected state too
* Defer focus so when you close a selected autocomplete it focuses back into the input
* Fix CSS of "Featured Artists & Artworks"


kanaabe:
Ah nice call.

PR #417, opened 2015-06-02
@craigspaeth Cleans featured links

kanaabe:
Closes https://github.com/artsy/positron/issues/416


craigspaeth:
Nice! :+1: However, I think you should add a test [here](https://github.com/artsy/positron/blob/master/client/apps/verticals/test/routes.coffee#L18) to ensure there's no regressions.

kanaabe:
Ah yes, adding

PR #435, opened 2015-06-08
@kanaabe: Can explicitly indicate a Vertical as "featured" to call it out in Magazine

craigspaeth:
Halley mentioned that the Gallery Insights vertical accidentally appeared called out instead of the Venice Biennale. So we need the ability to be able to explicitly only call out certain verticals, hence the checkbox. Also re-arranged some of the admin inputs to be clearer.

image


kanaabe:
Feature all the things!

PR #436, opened 2015-06-09
@gristleism Uses git link to update google-sanitize

kanaabe:
Directly include the repo in package.json so we can see the changes made in scribe-plugin-sanitize-google-doc


gristleism:
Looks good!

Issue #434, opened 2015-06-08
Captions disappearing from some uploaded images


halleyjohnson:
Captions have disappeared from some uploaded images in the two articles below—sourcing more. 
https://www.artsy.net/article/artsy-editorial-venice-biennale-2015-21-top-museum-shows-events
https://www.artsy.net/article/artsy-editorial-at-the-venice-biennale-balancing-hope-with-national-interests

Notice that the ones in the second article are not _all_ missing. The general behavior for captions is to disappear if the caption field is engaged with after the initial session (can explain more if that's not clear).

craigspaeth:
@halleyjohnson: You mentioned this happens when messing with the image module in the Writer edit UI right? Can you please log steps to reproduce this so we can try to track down the bug.

halleyjohnson:
@craigspaeth I meant this could be related; no idea how exactly these disappeared, we just stumbled upon them. But the bug I am aware of happens this way:

Enter artwork caption.
Save draft or article.
Reopen article.
Engage with caption field.
Caption disappears upon engage.

There may be a time factor—I'm not sure that every time you reopen the article this happens, but may lapse after a while of not editing it.

PR #443, opened 2015-06-11
@craigspaeth Fixes missing captions

kanaabe:
Closes https://github.com/artsy/positron/issues/434


craigspaeth:
Nice find! Ideally this comes with a test, but the trickiness tradeoff might not be worth it. I'll go ahead and merge, thanks!

Issue #444, opened 2015-06-12
Writer will log you in to Artsy when previewing a draft


craigspaeth:
~~Right now you can log in to Force by passing in the access token [through a header](https://github.com/artsy/artsy-passport/blob/master/index.coffee#L289). We may need to support logging in by passing the token through a query param as well.~~

We support logging in to Force by passing the access token in the query parmas. Positron should just be able to append `?access_token=...` to [the link](https://github.com/artsy/positron/blob/master/client/apps/edit/components/header/index.jade#L15).

PR #445, opened 2015-06-15
@craigspaeth Adds access_token to preview link

kanaabe:
Closes https://github.com/artsy/positron/issues/444

Appends `?access token=xx` to the preview link so we can log into Force when a user tries to preview an article.


craigspaeth:
:+1:

PR #456, opened 2015-07-01
@craigspeth - adds ability for an admin to add keywords to an article (closes #452)

### LINKED ISSUE #452: Admins should be able to define Article keywords
From Taktical, regarding Google News:
"Google has guidelines on how to set up Meta tags for google news.  https://support.google.com/news/publisher/answer/68297?hl=en  A standard Meta tag should look like this; <meta name="news_keywords" content="World Cup, Brazil 2014, Spain vs Netherlands">"

Admins should be able to add custom keywords to an article.

broskoski:
screen shot 2015-07-01 at 10 54 52 am

### DIFF
--- a/api/apps/articles/model.coffee
+++ b/api/apps/articles/model.coffee
@@ -40,7 +40,6 @@ schema = (->
   thumbnail_image: @string().allow('', null)
   tags: @array().items(@string())
   title: @string().allow('', null)
-  keywords: @string().allow('', null)
   published: @boolean().default(false)
   published_at: @date()
   lead_paragraph: @string().allow('', null)
--- a/api/apps/articles/model.coffee
+++ b/api/apps/articles/model.coffee
@@ -40,6 +40,7 @@ schema = (->
   thumbnail_image: @string().allow('', null)
   tags: @array().items(@string())
   title: @string().allow('', null)
+  keywords: @string().allow('', null)
   published: @boolean().default(false)
   published_at: @date()
   lead_paragraph: @string().allow('', null)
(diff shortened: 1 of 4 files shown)


craigspaeth:
Nice! Does it make more sense for this to be an array or would we only use this to Interpolate comma separated keywords directly into a meta tag? If the latter :+1: ready to merge!

broskoski:
Actually have another branch that does that but it felt pretty cumbersome tbh. This would copy the keywords string directly to the tag, which I admit could be more trouble than an array. Your call!

api/apps/articles/model.coffee
```diff
@@ -40,6 +40,7 @@ schema = (->
   thumbnail_image: @string().allow('', null)
   tags: @array().items(@string())
   title: @string().allow('', null)
+  keywords: @string().allow('', null)
```
craigspaeth: Going for array on the data end, this would just need to be `@array().items(@string())` like `tags` above.

client/apps/edit/components/layout/index.coffee
```diff
@@ -65,6 +65,7 @@ module.exports = class EditLayout extends Backbone.View
       lead_paragraph: @$('#edit-lead-paragraph').html()
       thumbnail_title: @$('#edit-thumbnail-title :input').val()
       thumbnail_teaser: @$('#edit-thumbnail-teaser :input').val()
+      keywords: @$('#edit-admin-keywords input').val()
```
craigspaeth: And here we'd just add a `.split(',')`

broskoski:
Ok, removed new tags attr, should now use existing tags array instead. No migration needed.

craigspaeth:
Nice!

Issue #460, opened 2015-07-06
Link not saving for specific article


craigspaeth:
1. Visit this article https://writer.artsy.net/articles/559aed4e2cb64006000a5db3/edit
2. Try to link "Come as you are" to http://nomadicdivision.org/exhibition/come-as-you-are/ with just clicking (no shortcut keys)
3. Click save draft
4. Reopen to find it's not linked

PR #461, opened 2015-07-06
@kanaabe: Save text more often

craigspaeth:
Solves https://github.com/artsy/positron/issues/460 —a Writer user was having trouble saving a link in text. Turns out we're probably just simply not updating the model often enough (was on key up and clicking certain elements—but if you hyperlink something with just clicks it could avoid these events). This updates the model on clicking off of the section—which is a good action to "lock in some changes".


kanaabe:
Ah, nice catch!

Issue #168, opened 2015-03-03
"Organizations" e.g. Artsy Editorial = Organization, Tess = User


halleyjohnson:
for linking articles to their authors, for example an Artsy Editorial piece Halley wrote gets reposted to her personal profile but lives on Editorial's profile. currently doing this via old admin

craigspaeth:
@halleyjohnson: I've thought it over and I'd like it if we held off on blindly porting this functionality over from the old posting system. At-least for the immediate term I would like to not make this a blocker for launching Writer/Magazine.

In the past we had a more social posting platform and the idea of a "repost" (kinda like a fancier form of "Like"ing) a la Tumblr made sense. Now the primary use case, as you explain, is to indicate that someone is a part of a "Artsy Editorial" or is writing on behalf of "Artsy staff". I don't think the concept of "reposting" fits this well.

I don't know the right answer yet, but as an example, what Medium does would be more appropriate. Medium essentially has two concepts: "Users" (or authors) and what I'm going to call "Organizations"

e.g.

User: https://medium.com/@mwichary
Organization: https://medium.com/designing-medium

As you can see on a user's profile it will either show the article by their name

![image](https://cloud.githubusercontent.com/assets/555859/6587762/de7eb6a8-c75e-11e4-9cce-cac7c956e2fb.png)

or by the organization name + the author name

![image](https://cloud.githubusercontent.com/assets/555859/6587773/f8bb0fe4-c75e-11e4-9249-a6595d767c6d.png)

I could imagine Artsy adopting a similar data model. Or maybe something as simple as a checkbox indicating this article was written by staff. Either way I don't think we should be able to post things on other user's profiles, nor should we be porting "reposting" functionality now that we're aiming for a less social-network-like editorial platform.

halleyjohnson:
@craigspaeth As mentioned, this sounds great to me. If you're comfortable enough with using that Medium format, I can run this by Editorial at standup this afternoon and get some general feedback?

halleyjohnson:
I think this can also solve our author issue on the landing page, right? So you'd get both "Artsy Editorial" and the writer's name on Magazine?

craigspaeth:
@halleyjohnson That's the general idea, but that level of detail I still need to work out with Robert.

craigspaeth:
[...]

PR #465, opened 2015-07-10
@kanaabe: Begin Organization support

craigspaeth:
This starts to add CRUD support for editorial "organizations". For more details on this see https://github.com/artsy/positron/issues/168

cap


client/components/admin_form/README.md
```diff
@@ -0,0 +1,37 @@
+# Admin Form
```
craigspaeth: This is used to DRY up the generic admin form UI verticals and organizations share... if more of these conventional admin-only UIs appear I might even go a step further and make a configurable app that can be mounted b/c they share similar routing code too.

kanaabe: Great, thanks for adding the readme.md!

PR #473, opened 2015-07-17
@broskoski: An admin can add authors to an Organization

craigspaeth:
This creates a "autocomplete list" component that should be useful for any admin list UIs in Writer (will maybe refactor the artist & artwork page featuring to use this). That in turn made it easy to build associating authors to an organization.

cap

### DIFF
--- a/client/components/autocomplete_list/README.md
+++ b/client/components/autocomplete_list/README.md
@@ -5,15 +5,19 @@ An autocomplete box that when selected adds to a list. Useful for "arrays" of da
 ## Example
 
 ````coffeescript
-select = AutocompleteList @$('#input-container')[0],
+list = AutocompleteList @$('#input-container')[0],
   label: 'Authors'
   name: 'author_ids'
   url: "#{sd.ARTSY_URL}/api/v1/match/users?term=%QUERY"
+  selected: (e, item, items) ->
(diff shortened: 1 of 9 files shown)


client/components/autocomplete_list/index.coffee
```diff
@@ -0,0 +1,48 @@
...
+module.exports.AutocompleteList = AutocompleteList = React.createClass
+
+  getInitialState: ->
+    loading: true, items: @props.items or []
+
+  componentDidMount: ->
+    @addAutocomplete()
+
+  addAutocomplete: ->
+    Autocomplete ?= require '../autocomplete/index.coffee'
```
broskoski: why require autocomplete like this?

craigspaeth: I was having some trouble with typeahead in tests... there's definitely a better solution to this though.

client/components/autocomplete_list/README.md
```diff
@@ -0,0 +1,23 @@
+# Autocomplete List
```
broskoski: :+1:

broskoski:
Very nice, sir!

craigspaeth:
Note to self: This PR is lacking tests, and I need to follow up on that.

PR #481, opened 2015-07-23
@craig use updated url tooltip plugin

jorystiefel:


craigspaeth:
w00t w00t! :sparkles: Open Source :sparkles: Looks good!

PR #482, opened 2015-08-03
@kanaabe: Can query articles by organization_id

craigspaeth:
This will make it easier to create an organization page by allowing a `organizaiton_id=` param to the articles API.

Screenshot of Force PR below (coming soon). Keeping it super minimal b/c I think this page deserves a real design pass before I go replicating the user profile page.

image


kanaabe:
v. nice! GR:eight:

PR #489, opened 2015-08-21
@craigspaeth Adds contributing authors

kanaabe:
Also adds a `all_by_author` query param which finds articles that an author has been the main author AND contributing author of. Will use in Force for adding articles to a contributor's profile page. 

contributing

### DIFF
--- a/api/apps/articles/model.coffee
+++ b/api/apps/articles/model.coffee
@@ -82,6 +82,7 @@ schema = (->
   biography_for_artist_id: @objectId().allow(null)
   featured: @boolean().default(false)
   exclude_google_news: @boolean().default(false)
+  contributing_authors: @array().items(@objectId()).allow(null)
 ).call Joi
 
 querySchema = (->
@@ -103,6 +104,7 @@ querySchema = (->
   exclude_google_news: @boolean()
   q: @string()
   organization_id: @objectId()
+  all_by_author: @objectId()
 ).call Joi
 
 #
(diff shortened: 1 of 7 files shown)


api/apps/articles/model.coffee
```diff
@@ -140,6 +142,12 @@ toQuery = (input, callback) ->
     query.auction_id = ObjectId input.auction_id if input.auction_id
     query.vertical_id = ObjectId input.vertical_id if input.vertical_id
     query.biography_for_artist_id = ObjectId input.biography_for_artist_id if input.biography_for_artist_id
+
+    query.$or = [
+      { author_id: ObjectId(input.all_by_author) }
+      { contributing_authors: ObjectId(input.all_by_author) }
```
craigspaeth: Niice!

client/apps/edit/components/admin/index.coffee
```diff
@@ -145,6 +147,36 @@ module.exports = class EditAdmin extends Backbone.View
...
+      placeholder: 'Search by user name or email...'
+      filter: (users) -> for user in users
+        { id: user.id, value: _.compact([user.name, user.email]).join(', ') }
+      selected: (e, item, items) =>
+        @article.save contributing_authors: _.pluck items, 'id'
+      removed: (e, item, items) =>
+        @article.save contributing_authors: _.without(_.pluck(items, 'id'),item.id)
+    if ids = @article.get 'contributing_authors'
+      @authors = []
+      async.each ids, (id, cb) =>
```
craigspaeth: Nice! Looks like you grasped async pretty easily! Let me know if anything is unclear with how that control flow lib works.

craigspaeth:
Solid work! Code is great!

kanaabe:
@craigspaeth Do you think we should also be saving user `name` and profile `hrefs` ? I realize we'll need to fetch users on the Force/MG to show the bylines. Not sure what is cleaner/best practice here for API design.

craigspaeth:
@kanaabe Denormalizing some user data into the articles e.g. `article.contributing_authors: [{ id: ObjectId(), name: '', profile_id: '', profile_handle: '' }]` is probably a good idea. We already do this for the primary author [see here](https://github.com/artsy/positron/blob/master/api/apps/articles/model.coffee#L224). So :+1: from me. Either way if it's just an array of ids the field should probably be renamed to `contributing_author_ids` to be specific.

Issue #490, opened 2015-08-21
An admin can curate an overview page of "featured links" for a "BrandPartner"


owendodd:
Moved Owen's comments to this issue https://github.com/artsy/force/issues/3316

> The main change to the current page will be the fact that the overview page will be a set of featured links rather than an automated feed of current / past shows. Each link will have a thumbnail, subheadline (describing the content type), headline, and desc (which will be used as either an author byline, date range, or location).

I can follow up with a comp for the Writer aspect of this, but basically we'll want to reuse generic admin UI to allow "10" featured links. You can reference how we do this in Verticals.

![image](https://cloud.githubusercontent.com/assets/555859/9445087/44ee9698-4a58-11e5-88d1-aee8f103f6e8.png)

So a `brand_partner_overview` API app that models the data something like...

```
{
  partner_id: ObjectId() // References gravity partner `_id` field
  featured_links: [
    {
      thumbnail_url: '',
      headline: '',
      subheadline: '',
      description: ''
    }
  ]
}
```

alexanderfabry:
Should this story be in force?

I think we'll want a story in positron around managing the featured links for the BPP, though.

owendodd:
@craigspaeth mentioned it would be better here. Maybe since the majority of the work will be around the featured links it seems. Happy to break stories out if there is a way that makes more sense though Craig!

alexanderfabry:
Would personally find it clearer to have the work items broken out, since there's definitely stuff that needs to happen on Force, but happy to do whatever @craigspaeth and @kanaabe like best!

craigspaeth:
:-P in my attempt to save you some time Owen, I've probably ended up causing us to spend more time discussing this. Sorry about that! I went ahead and broke up the issues https://github.com/artsy/force/issues/3316

owendodd:
Haha, much appreciated Craig!

PR #491, opened 2015-08-25
@craigspaeth Brand partnerships

kanaabe:
Closes https://github.com/artsy/positron/issues/490

brandpartner

### DIFF
--- a/api/apps/brand_partners/index.coffee
+++ b/api/apps/brand_partners/index.coffee
@@ -4,8 +4,8 @@ routes = require './routes'
 
 app = module.exports = express()
 
-app.get '/brand-partners', routes.index
-app.get '/brand-partners/:id', routes.find, routes.show
-app.post '/brand-partners', setUser, authenticated, adminOnly, routes.save
-app.put '/brand-partners/:id', setUser, authenticated, adminOnly, routes.find, routes.save
-app.delete '/brand-partners/:id', setUser, authenticated, adminOnly, routes.find, routes.delete
+app.get '/brand_partners', routes.index
+app.get '/brand_partners/:id', routes.find, routes.show
+app.post '/brand_partners', setUser, authenticated, adminOnly, routes.save
+app.put '/brand_partners/:id', setUser, authenticated, adminOnly, routes.find, routes.save
+app.delete '/brand_partners/:id', setUser, authenticated, adminOnly, routes.find, routes.delete
--- a/api/apps/brand_partners/index.coffee
+++ b/api/apps/brand_partners/index.coffee
@@ -1,0 +1,11 @@
+express = require 'express'
+routes = require './routes'
+{ setUser, authenticated, adminOnly } = require '../users/routes'
+
+app = module.exports = express()
+
+app.get '/brand_partners', routes.index
+app.get '/brand_partners/:id', routes.find, routes.show
+app.post '/brand_partners', setUser, authenticated, adminOnly, routes.save
+app.put '/brand_partners/:id', setUser, authenticated, adminOnly, routes.find, routes.save
+app.delete '/brand_partners/:id', setUser, authenticated, adminOnly, routes.find, routes.delete
--- a/api/apps/brand_partners/index.coffee
+++ b/api/apps/brand_partners/index.coffee
@@ -4,8 +4,8 @@ routes = require './routes'
 
 app = module.exports = express()
 
-app.get '/brand_partners', routes.index
-app.get '/brand_partners/:id', routes.find, routes.show
-app.post '/brand_partners', setUser, authenticated, adminOnly, routes.save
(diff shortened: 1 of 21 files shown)


craigspaeth:
Amazing! That was so fast! Impressive! :sparkles: :boom: :fireworks: :sparkles:

client/components/layout/templates/sidebar.jade
```diff
@@ -25,11 +25,11 @@
...
         br
         | Verticals
-      a( href='/organizations'
-         class=(sd.URL.match('/organizations') ? 'is-active' : '') )
+      a( href='/brand-partners'
+         class=(sd.URL.match('/brand-partners') ? 'is-active' : '') )
         include ../public/icons/layout_organizations.svg
         br
-        | Orgs
+        | Brands
```
kanaabe: Might not be the best icon to use -- let me know if you have a better one! @owendodd

api/apps/brand_partners/model.coffee
```diff
@@ -0,0 +1,83 @@
...
+{ ObjectId } = require 'mongojs'
+{ ARTSY_URL, API_MAX, API_PAGE_SIZE } = process.env
+
+#
+# Schemas
+#
+schema = (->
+  id: @objectId()
+  partner_id: @objectId()
+  slug: @string().allow('', null)
```
craigspaeth: I don't think this will be necessary b/c the slug is determined by the partner model.

kanaabe: Good point, so make it searchable [here?](https://github.com/kanaabe/positron/blob/brandpartnership/api/apps/brand_partners/model.coffee#L41)

api/apps/brand_partners/index.coffee
```diff
@@ -0,0 +1,11 @@
+express = require 'express'
+routes = require './routes'
+{ setUser, authenticated, adminOnly } = require '../users/routes'
+
+app = module.exports = express()
+
+app.get '/brand-partners', routes.index
```
craigspaeth: I think this should be underscore case to keep consistency with our main API & json fields.

kanaabe: Ah yes, I had it underscored previously, should have gone with my gut. Thanks!

client/components/admin_form/index.styl
```diff
@@ -39,6 +39,13 @@ gutter-size = 30px
 .admin-form-right-col + .admin-form-right-col
   padding-left gutter-size
 
+.admin-form-input-section
+  display inline-block
+  width 50%
+
+.admin-form-input-section + .admin-form-input-section
```
craigspaeth: :+1:

api/apps/brand_partners/index.coffee
kanaabe: Should we keep the client url as `brand-partners`?

test/helpers/fixtures.coffee
```diff
@@ -78,6 +78,17 @@ module.exports = ->
     'name': 'Artsy Editorial'
     'slug': 'editorial'
     'author_ids': null
+  fixtures.brandPartners =
+    id: '559ff9706b69f6a086a65633'
+    slug: 'VersaceVersace'
```
craigspaeth: I do like this slug though!

api/apps/brand_partners/index.coffee
craigspaeth: Hmm no strong opinion here, keeping it dashes on the client works for me.

In general dashes are blessed as the more "SEO friendly" urls b/c Google used to not treat underscores as word separators, but that 1. Obviously doesn't matter for a non-website app like Writer and 2. I think Google treats both as separators  these days anyways.

api/apps/brand_partners/model.coffee
craigspaeth: So in that case we can just do `query = { _id: ObjectId(id) }` b/c we don't need to support `/api/brand_partners/:slug` we should instead just query based on the partner id (which you support [here](https://github.com/kanaabe/positron/blob/brandpartnership/api/apps/brand_partners/model.coffee#L34) but I think you need to add a line that converts string -> object id [here](https://github.com/kanaabe/positron/blob/brandpartnership/api/apps/brand_partners/model.coffee#L48)) e.g. `/api/brand_partners?partner_id=:id`.

craigspaeth: Example of converting string -> objectId in where function: https://github.com/kanaabe/positron/blob/master/api/apps/articles/model.coffee#L138 ...going to submit that patch to Joi-objectid some day!

craigspaeth:
Looks good! Mainly concerned about the comments on snake_case and ensuring querying `/api/brand_partners?partner_id=` works. Then it's good to merge!

craigspaeth:
Well done!

Issue #494, opened 2015-08-25
Denormalize contributing authors


kanaabe:
incl. name and profile href
For less fetches in Force/MG

PR #497, opened 2015-08-26
@craigspaeth Denormalize contributing author data

kanaabe:
Closes https://github.com/artsy/positron/issues/494

### DIFF
--- a/api/apps/articles/model.coffee
+++ b/api/apps/articles/model.coffee
@@ -211,10 +211,10 @@ sortParamToQuery = (input) ->
           featured_artist_ids: article.featured_artist_ids.map(ObjectId) if article.featured_artist_ids
           featured_artwork_ids: article.featured_artwork_ids.map(ObjectId) if article.featured_artwork_ids
           biography_for_artist_id: ObjectId(article.biography_for_artist_id) if article.biography_for_artist_id
-          contributing_authors: article.contributing_authors.map( (id)->
-            id.id = ObjectId(id.id)
-            id
-            ) if article.contributing_authors
+          contributing_authors: article.contributing_authors.map( (author)->
+            author.id = ObjectId(author.id)
+            author
+          ) if article.contributing_authors
         ), callback
 
 validate = (input, callback) ->
--- a/api/apps/articles/model.coffee
+++ b/api/apps/articles/model.coffee
@@ -82,7 +82,12 @@ schema = (->
   biography_for_artist_id: @objectId().allow(null)
   featured: @boolean().default(false)
   exclude_google_news: @boolean().default(false)
-  contributing_authors: @array().items(@objectId()).allow(null)
+  contributing_authors: @array().items([
+    @object().keys
+      id: @objectId().allow(null)
+      name: @string().allow('', null)
+      profile_id: @string().allow('', null)
+    ]).allow(null)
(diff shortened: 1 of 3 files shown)


api/apps/articles/model.coffee
```diff
@@ -145,7 +150,7 @@ toQuery = (input, callback) ->
 
     query.$or = [
       { author_id: ObjectId(input.all_by_author) }
-      { contributing_authors: ObjectId(input.all_by_author) }
+      { contributing_authors: { $elemMatch: { id: ObjectId input.all_by_author} } }
```
craigspaeth: `$elemMatch` nice!

api/apps/articles/model.coffee
```diff
@@ -206,7 +211,10 @@ sortParamToQuery = (input) ->
           featured_artist_ids: article.featured_artist_ids.map(ObjectId) if article.featured_artist_ids
           featured_artwork_ids: article.featured_artwork_ids.map(ObjectId) if article.featured_artwork_ids
           biography_for_artist_id: ObjectId(article.biography_for_artist_id) if article.biography_for_artist_id
-          contributing_authors: article.contributing_authors.map(ObjectId) if article.contributing_authors
+          contributing_authors: article.contributing_authors.map( (id)->
```
craigspaeth: I think it would be clearer to name this `(author) ->` and `author.id = ObjectId(author.id)`

craigspaeth:
So looking at this it appears the client saves the denormalized author data, and the server accepts whatever it sends down. The way the primary author is denormalized is accepting just the author's id and the server fills in the denormalized `author` from it's own data.

At first this seemed odd to me, but then I thought about it actually seems like a simpler + fine way to do it. I guess one issue is if a user changes their name and the data gets out of sync... but that's a problem to be addressed separately with either approach. Another issue might be if this API were used by various 3rd parties that would allow for irresponsible clients to save junk data, but also not a use case we need to address right now.

So in conclusion, this works for me! Just wanted to muse on the tradeoffs involved in this approach.

craigspaeth:
Good work! We can expand on the potential edge case issues with denormalizing when/if they arrive.

kanaabe:
Good point, I guess one way to solve that would be to fetch and denormalize author data from Gravity on find/present and just save IDs. Seems problematic though. Do people do that kind of stuff?

craigspaeth:
Sure do. Technically that's an application-level join—just instead of within the database it's joining a query from the database with a query to the Gravity API. However, that kinda defeats the purpose of denormalizing this data which is to avoid joining API requests like that (it would be just as easy, and probably preferable, to implement that on the Force-side).

My initial thought was to fetch & save the denormalized author data upon [saving the article](https://github.com/artsy/positron/blob/master/api/apps/articles/model.coffee#L198) or [upon updating the author](https://github.com/artsy/positron/blob/master/api/apps/users/model.coffee#L65). Of course that will still get out of sync if the user data is updated in Gravity—but the only way to avoid that is to do the join described above... so tradeoffs ¯\_(シ)_/¯

Again, no need to over-think this now. I think what's in place is a solid solution!

PS. This is an awesome related read from Mongo's blog: http://blog.mongodb.org/post/88473035333/6-rules-of-thumb-for-mongodb-schema-design-part-3

PR #503, opened 2015-09-03
@craigspaeth Removes failed fetches from adding to featured collection

kanaabe:
The JS on the admin panel was breaking (prevented being able to save because of failed validation) for artworks/artists that are removed from Artsy. This just makes sure an artwork/artist is there before adding it to a collection.


craigspaeth:
:+1

PR #507, opened 2015-09-10
@craigspaeth Adds [] default contributing authors

kanaabe:
Cleaner on the front-end to deal with `.length` instead of checking null.

### DIFF
--- a/api/apps/articles/model.coffee
+++ b/api/apps/articles/model.coffee
@@ -255,4 +255,3 @@ addSlug = (article, input, author) ->
     _id: undefined
     slug: _.last article.slugs
     slugs: undefined
-    contributing_authors: article.contributing_authors or []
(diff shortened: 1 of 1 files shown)


api/apps/articles/model.coffee
```diff
@@ -255,3 +255,4 @@ addSlug = (article, input, author) ->
     _id: undefined
     slug: _.last article.slugs
     slugs: undefined
+    contributing_authors: article.contributing_authors or []
```
kanaabe: Wondering if instead I should update this in the DB.

craigspaeth: I think ideally the database should try to reflect the json responses as much as possible so update in the db preferred. Use `db.articles.update({ contributing_authors: null }, { contributing_authors: [] }, { multi: true })` but make sure to try it on local/staging first. You can do this with robomongo or `heroku run coffee --app=positron-production` (the former will be easier).

craigspaeth:
:+1: Definitely better to default to `[]` here. I feel a little ashamed for crutching on allowing null in so many places.

PR #508, opened 2015-09-16
@craigspaeth Verticals -> Sections

kanaabe:
Changed:
- New endpoint for `sections` which is identical to the current `verticals`
- Section UI edits `sections` endpoint and replaces vertical editing
- Admin->Edit replaces ability to add a single "vertical" to use AutocompleteList to select multiple "sections"

Side note: DB is now updated to include `sections` collection and `section_ids` in articles based on a query which copies any existing `vertical_id` to `section_ids`. 
 
From here the plan is to: 
1. PR Force and MG to use new endpoints - in the meantime this means editing a vertical or adding articles to sections will do nothing cc @halleyjohnson 
2. Delete Verticals
3. Add other features: 
>If the user clicks an article from a section, that section's stickies & such should display. If they click in from an external source or a page on the site that's not a section, could we set a "default" or "primary" section on a per-article basis? Maybe it defaults to the first section it's added to but can be overridden?


test/helpers/fixtures.coffee
```diff
@@ -73,6 +73,14 @@ module.exports = ->
     'partner_logo_url': 'http://gemini.herokuapp.com/123/miaart-banner.jpg'
     'thumbnail_url': 'http://gemini.herokuapp.com/123/miaart-banner.jpg'
     'featured_article_ids': [ '5522d03ae8e369060053d953' ]
+  fixtures.sections =
```
craigspaeth: Probably don't need the verticals fixture here anymore.

client/apps/edit/components/admin/index.coffee
```diff
@@ -93,23 +93,30 @@ module.exports = class EditAdmin extends Backbone.View
...
-      selected: (e, item) =>
-        @article.save vertical_id: item.id
-      cleared: =>
-        @article.save vertical_id: null
-    if id = @article.get 'vertical_id'
-      request
-        .get("#{sd.API_URL}/verticals/#{id}").end (err, res) ->
-          select.setState value: res.body.title, loading: false
+  setupSectionAutocomplete: ->
+    AutocompleteList = require '../../../../components/autocomplete_list/index.coffee'
```
craigspaeth: `AutocompleteList` w00t w00t!

test/helpers/fixtures.coffee
kanaabe: Yeah def, I'll remove it at the "Delete Vertical" step since the `/vertical` endpoint still exists and uses the fixture

craigspaeth:
Awesome! Great work making this migration smooth! #3 on your list sounds like an interesting feature... I would log that in the Publishing team Trello and hold off until it gets prioritized next Publishing meeting, but ultimately defer that decision to you.

Great job! These things are always quite tricky and you've made it look effortless! :100: :confetti_ball: Looking forward to the Force/MG PRs.

PR #509, opened 2015-09-17
@craigspaeth Deletes verticals

kanaabe:
Should also get rid of the collection and the article's `vertical_id` reference in the DB.


craigspaeth:
Nice!

craigspaeth:
Be sure to use the surgical `{ $unset }` query when removing vertical_id and of course always test on local/staging first!

PR #511, opened 2015-09-18
@craigspaeth Adds autocomplete list to partners admin

kanaabe:
Articles can be associated to multiple partners. The article will display in all the partner profiles selected.


craigspaeth:
Easy peasy! Nice work!

PR #513, opened 2015-09-21
@craigspaeth Fix contributing author fetch bug

kanaabe:
Separated the `ids` assignment from the if statement.


kanaabe:
Merging for now since we can't access contributing authors ATM.

PR #520, opened 2015-09-28
@craigspaeth Email Metadata

kanaabe:
Uploading a thumbnail image resizes and crops using Embedly. Hoping to use our own tools soon. https://github.com/artsy/gemini/issues/166


client/apps/edit/components/thumbnail/index.coffee
```diff
@@ -43,3 +43,22 @@ module.exports = class EditThumbnail extends Backbone.View
...
+      remove: =>
+        emailMetadata = @article.get('email_metadata') || {}
+        emailMetadata.large_image_url = ''
+        emailMetadata.small_image_url = ''
+        $('.edit-email-large-image-url span').text ''
+        $('.edit-email-small-image-url span').text ''
+        @article.save email_metadata: emailMetadata
+      done: (src) =>
+        emailMetadata = @article.get('email_metadata') || {}
+        emailMetadata.large_image_url = crop(src, { width: 1280, height: 960 } )
```
craigspaeth: Currently Positron just stores original urls and lets the client-app use Embedly/Gemini for resizing. I think it would be better to consolidate this to just `image_url` rather than save embedly urls to the article. That is unless I'm missing some key piece of context like the "client app" in this case is Mailchimp which can't easily integrate embedly—in which case I still think it would be good to store the original, pre-cropped image under `image_url` along-side this.

craigspaeth:
Looks good! I think we need to still store the pre-cropped image url in case we stop using Embedly and lose data. Otherwise feel free to merge when :green_heart:

client/apps/edit/components/thumbnail/index.coffee
kanaabe: Ok cool! I had a feeling this was a better way to do it. :smile:

PR #524, opened 2015-09-29
@craigspaeth Lock express-force-ssl version

kanaabe:
:lock:


craigspaeth:
:+1: Could also use `app.set('trustXFPHeader', true)` ([related](https://github.com/battlejj/express-force-ssl#as-of-v030-there-are-some-configuration-options)). But I'm totally down with :lock: too

PR #525, opened 2015-09-30
@craigspaeth Upload to S3 script

kanaabe:
Creates a CSV out of positron's article data and then pushes it to the appropriate S3 bucket.


scripts/daily_upload_s3.coffee
```diff
@@ -0,0 +1,49 @@
...
+writableStream = fs.createWriteStream('scripts/tmp/' + filename)
+
+csvStream.pipe(writableStream)
+csvStream.write(["id", "author_id", "auction_id", "contributing_authors", "fair_id", "featured", "featured_artist_ids", "featured_artwork_ids", "partner_ids", "primary_featured_artist_ids", "slugs", "tags", "title", "tier", "published_at","show_ids","section_ids","thumbnail_image","thumbnail_title"])
+
+db.articles.find({ published: true })
+  .on('data', (doc) ->
+    if doc
+      published_at = if doc.published_at then moment(doc.published_at).format('YYYYMMDDhhmmss') + " EST" else ''
+      csvStream.write([doc._id, doc.author_id, doc.auction_id, doc.contributing_authors, doc.fair_id, doc.featured, doc.featured_artist_ids, doc.featured_artwork_ids, doc.partner_ids, doc.primary_featured_artist_ids, doc.slugs, doc.tags, doc.title, doc.tier, published_at, doc.show_ids, doc.section_ids, doc.thumbnail_image, doc.thumbnail_title])
```
craigspaeth: Minor style comment: Would prefer to have this break around line 80 than trail off.

.gitignore
```diff
@@ -15,4 +15,5 @@ client/public/assets
 dump.rdb
 .env
 .env.ignore
-node_modules
\ No newline at end of file
+node_modules
+scripts/tmp
```
kanaabe: The generated CSVs go in a tmp folder -- I added this just in case someone accidentally runs the script and the CSV didn't delete itself for whatever reason.

scripts/daily_upload_s3.coffee
```diff
@@ -0,0 +1,49 @@
...
+
+# Connect to database
+db = mongojs(process.env.MONGOHQ_URL, ['articles'])
+
+# Setup CSV
+filename = "export_" + moment().format('YYYYMMDDhhmmss') + ".csv"
+csvStream = csv.createWriteStream { headers: true }
+writableStream = fs.createWriteStream('scripts/tmp/' + filename)
+
+csvStream.pipe(writableStream)
```
craigspaeth: Very nice use of streams going on here!

scripts/daily_upload_s3.coffee
```diff
@@ -0,0 +1,49 @@
...
+    # End Streaming
+    csvStream.end()
+
+    # Setup S3 Client
+    client = knox.createClient
+      key: process.env.S3_KEY
+      secret: process.env.S3_SECRET
+      bucket: process.env.FULCRUM_BUCKET
+
+    client.putFile 'scripts/tmp/' + filename, "reports/positron_articles/#{filename}", {
```
craigspaeth: Writing to a local file on disk and then uploading that file all at once is legit, but if you're curious you could probably avoid the extra local writing and reading by piping your `csvStream` directly to knox's put API (related https://github.com/Automattic/knox#examples). e.g. something like this:

```coffeescript
s3PutStream = client.put('scripts/tmp/' + filename)
csvStream.pipe(s3PutStream)
```

scripts/daily_upload_s3.coffee
```diff
@@ -0,0 +1,49 @@
...
+  ).on 'end', ->
+
+    # End Streaming
+    csvStream.end()
+
+    # Setup S3 Client
+    client = knox.createClient
+      key: process.env.S3_KEY
+      secret: process.env.S3_SECRET
+      bucket: process.env.FULCRUM_BUCKET
```
craigspaeth: Would be nice to add a `FULCRUM_BUCKET=REPLACE` in the .env.example here so we keep track of all the of config used. Also I like to use destructuring to clean this up e.g. `{ FULCRUM_BUCKET } = process.env`—but just another stylistic choice.

scripts/daily_upload_s3.coffee
kanaabe: Ah! This is definitely the route to go. I'll commit that.

.gitignore
craigspaeth: Very good thinking ahead on this!

craigspaeth:
No blockers on my end! Well done :clap: Analytics folks will be thrilled to have this! I assume this is run daily through heroku scheduler, but lmk if you need advice on that end.

PR #538, opened 2015-10-15
@kanaabe: HTML XSS sanitizing at API level

craigspaeth:
This uses [an xss sanitization module](https://github.com/leizongmin/js-xss) to escape bad HTML from articles at the API level and prevent haxorz from trying to put `alert('hi')`s in their articles via the API—saving those users who accidentally wander on to 1337fr33kn00bkiller's Artsy profile. Ideally these "text modules" have really stripped down html anyways, so I actually wish this were in place sooner if nothing else but to further clean up copypasta-ed HTML. I know some have abused the looseness of Scribe's sanitization—so hopefully this doesn't shock ppl and cause any headaches.

What's kinda neat is about this being an :sparkles: isomorphic :sparkles: javascript app is that, if necessary, we could extract this lib into a module that can be used in our [client-side Scribe sanitization plugin](https://github.com/artsy/positron/blob/master/client/apps/edit/lib/sanitizer.coffee) and give immediate feedback—e.g. avoiding someone being surprised the garbage they pasted into Writer didn't stick when they refreshed the page.

### DIFF
--- a/api/apps/articles/model.coffee
+++ b/api/apps/articles/model.coffee
@@ -11,6 +11,9 @@ async = require 'async'
 Joi = require 'joi'
 Joi.objectId = require('joi-objectid') Joi
 moment = require 'moment'
+xss = require 'xss'
+cheerio = require 'cheerio'
+url = require 'url'
 { ObjectId } = require 'mongojs'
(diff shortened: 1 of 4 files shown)


kanaabe:
Nice!

PR #539, opened 2015-10-16
@kanaabe: Fix filtering out non-text sections

craigspaeth:
Whoops!


kanaabe:
Oh yes. There we go, I should have caught that.

PR #540, opened 2015-10-19
@craigspaeth Fixes fulcrum script 

kanaabe:
The Heroku instance doesn't have the tmp folder created, so the script was failing. This creates the tmp folder if it doesn't exist yet. 

I tried doing a direct mongo read to S3 stream but it kept failing, I think because of the unknown size. Maybe we can do a pair refactor sometime!

### DIFF
--- a/scripts/daily_upload_s3.coffee
+++ b/scripts/daily_upload_s3.coffee
@@ -19,7 +19,12 @@ db = mongojs(process.env.MONGOHQ_URL, ['articles'])
 # Setup CSV
 filename = "export_" + moment().format('YYYYMMDDhhmmss') + ".csv"
 csvStream = csv.createWriteStream { headers: true }
-writableStream = fs.createWriteStream('scripts/tmp/' + filename)
+dir = 'scripts/tmp/'
+if !fs.existsSync dir
+  fs.mkdir dir
+
+writableStream = fs.createWriteStream( dir + filename)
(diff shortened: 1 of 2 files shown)


craigspaeth:
Hmm yeah an unknown content size would be tricky to work around with streams—this might be the only reasonable solution after all.

scripts/daily_upload_s3.coffee
```diff
@@ -19,7 +19,12 @@ db = mongojs(process.env.MONGOHQ_URL, ['articles'])
 # Setup CSV
 filename = "export_" + moment().format('YYYYMMDDhhmmss') + ".csv"
 csvStream = csv.createWriteStream { headers: true }
-writableStream = fs.createWriteStream('scripts/tmp/' + filename)
+dir = 'scripts/tmp/'
+if !fs.existsSync dir
+  fs.mkdir dir
```
craigspaeth: `fs.mkdir` requires a callback, and as it stands you'll like have a race condition where the script might try to write to a non-existent directory b/c `mkdir` didn't finish. You could use `fs.mkdirSync` to avoid this or better yet you can just check in the tmp folder by creating the directory and adding an empty .gitkeep file inside of it.

kanaabe: Ah good call, update to come.

PR #552, opened 2015-10-28
@jtotoole: Pluralize some admin stuff

craigspaeth:
I used the singular label for two form fields that are actually "has many" associations.


jtotoole:
:+1: :+1:

PR #553, opened 2015-10-28
@jtotoole: Fix for new backbone

craigspaeth:
Related to 1.2.3 update from Backbone that no longer allows `null` `at:` properties http://backbonejs.org/#changelog


jtotoole:
:100: :100: :100:

PR #568, opened 2015-11-06
@craigspaeth: get twitter + facebook profile links for admins

jtotoole:
Is there a good way of testing to ensure that we're not exposing non-admin social-media UIDs? I was stumbling on that because [the data](https://github.com/artsy/antigravity/blob/master/node/fabricate.coffee#L175) Antigravity returns from the `/user/:id` endpoint that [the findOrInsert method hits](https://github.com/artsy/positron/blob/master/api/apps/users/model.coffee#L50) is set to 'Admin'


craigspaeth:
Yes. Like I mentioned irl you can short-circuit antigravity to add your own routes b/c it's just an express app [around here](https://github.com/jtotoole/positron/blob/facebook-id/api/apps/users/test/model.coffee#L10).

e.g.

````coffeescript
it 'doesnt save non-admin data', (done) ->
  app.get '/api/v1/user/:id', (req, res, next) ->
    res.send { type: 'User' } 
  app.use '/__gravity', require('antigravity').server
  ...
````

I'll take a look at that when I pull and see what's up with simple modal

File: api/apps/users/model.coffee [hunk 1]
```diff
@@ -65,10 +65,17 @@ save = (user, accessToken, callback) ->
...
+    if user.type is "Admin"
+      (cb) ->
+        request
+        .get("#{ARTSY_URL}/api/v1/me/authentications")
+        .set('X-Access-Token': accessToken).end cb
   ], (err, results) ->
     return callback err if err
     encryptedAccessToken = results[0]
     profile = results[1].body
+    facebook_uid = results[2].body[0].uid if results[2].body[0].uid?
```
craigspaeth: Code golf tip you can assign a var in your if like so `facebook_uid = uid if uid = results[2].body[0].uid?`

craigspaeth:
:+1: Well done! I'll merge and see about that extra test.

PR #569, opened 2015-11-06
@jtotoole: Add test & update modules

craigspaeth:
I couldn't repro the issue you're having with simple-modal... I'll have to see what's going on in your branch.

### DIFF
--- a/api/apps/users/test/model.coffee
+++ b/api/apps/users/test/model.coffee
@@ -48,6 +48,14 @@ describe 'User', ->
         user.twitter_uid.should.equal '321'
         done()
 
+    it 'does not get user social media UIDs', (done) ->
+      user = _.extend fixtures().users, type: 'Admin'
+      db.users.insert user, ->
+        User.findOrInsert user._id, 'foobar', (err, user) ->
+          (user.facebook_uid?).should.be.false()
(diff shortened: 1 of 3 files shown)


api/apps/users/test/model.coffee
```diff
@@ -48,6 +48,14 @@ describe 'User', ->
         user.twitter_uid.should.equal '321'
         done()
 
+    it 'does not get user social media UIDs', (done) ->
+      user = _.extend fixtures().users, type: 'Admin'
```
craigspaeth: I was wrong about having to change antigravity. Because `findOrInsert` looks up users in Positron's db, we can just insert a stub user. (The other approach could work too, but this is simpler).

craigspaeth: Hmm `type: 'Admin` is not what we want to test! One sec lemme investigate.

craigspaeth: Lolz ignore everything I said—I do need to intercept antigravity ༼;´༎ຶ ۝ ༎ຶ༽

craigspaeth:
Ok know things are in order.

jtotoole:
:+1: :+1: :+1:

PR #570, opened 2015-11-09
@craigspaeth Adds a descending counter for recommended number of chars in a thumbnail title

kanaabe:
Warns the user when they are nearing over 97 chars which would push them to the undesired four lines seen here (added text to the original title): 

Happy Title

Unhappy Title

You can successfully save an article that goes over that limit, it's more of a warning for now.

Closes https://github.com/artsy/force/issues/3629


owendodd:
:raised_hands:

craigspaeth:
Well done! To the point w/ tests :+1:

PR #572, opened 2015-11-10
@jtotoole: Point back to non-hosted simple-modal

craigspaeth:
I needed to do this before b/c of some weird issue with `peerDependencies` and browserify. Seems to have maybe resolved itself now and this should fix the deploy build.


jtotoole:
:fist: Simple Modal!

PR #573, opened 2015-11-11
@kanaabe Adds autocomplete list to shows and fixes selection

kanaabe:

### DIFF
--- a/client/apps/edit/components/admin/index.jade
+++ b/client/apps/edit/components/admin/index.jade
@@ -48,18 +48,18 @@ section#edit-admin.admin-form-container.max-width-container
     .admin-form-right
       .admin-form-right-col
         label Fair
-          #edit-admin-fair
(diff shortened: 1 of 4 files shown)


client/apps/edit/components/admin/index.jade
```diff
@@ -48,18 +48,18 @@ section#edit-admin.admin-form-container.max-width-container
...
       .admin-form-right-col
         label Editorial Section
-          #edit-admin-section
+        #edit-admin-section
         label Show
-          #edit-admin-shows
+        #edit-admin-shows
         label Auction
-          #edit-admin-auction
+        #edit-admin-auction
```
kanaabe: Turns out that autocomplete did not like the that the parent element was a label.

kanaabe: cc @craigspaeth

PR #581, opened 2015-11-15
@craigspaeth: Unlinked artists script

jtotoole:


scripts/unlinked_artists.coffee
```diff
@@ -2,52 +2,75 @@
...
-  for artist, i in artists
-    name = (artist.first?.trim() or '') + ' ' + (artist.last?.trim() or '')
-    if name.trim() and articlesHtml.indexOf(name) >= 0
-      slugs.push artist._slugs[0]
-  slugs
+findUnlinked = ($) ->
+  namesAndCounts = {}
+  html = $.html().toLowerCase()
+  for name in ARTISTS
+    nameLink = 'artsy.net/artist/' + name.split(' ').join('-')
```
craigspaeth: I think you will want to do `name.toLowerCase().split(' ').join('-')`. However, this won't be as reliable as testing if any of the links have the artist name inside of them because some slugs aren't 1:1 with the name e.g. there might be slugs like `andy-warhol1`. However, I think this good enough and will match most of the time.

The other way might look something like...

````coffeescript
linkExists = name.toLowerCase() in $('a').map(-> $(this).text().toLowerCase().trim())
````

PR #583, opened 2015-11-16
@jtotoole: Minor design tweaks on publish scheduling

craigspaeth:
B/c I'm a pixel pusher I made some small design tweaks to the schedule to publish feature.
1. Changed scheduled to be published to an Artsy green b/c to me red seemed like it was an error that needed to be handled

image
1. Moved forms to same row and adjusted the button state to look more like our follow buttons on the site.

To be scheduled

image

Scheduled

image

Hover

image

### DIFF
--- a/client/apps/article_list/index.styl
+++ b/client/apps/article_list/index.styl
@@ -104,6 +104,9 @@
   vertical-align middle
   margin 0 23px 0 15px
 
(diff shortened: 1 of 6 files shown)


kanaabe:
Pixel pusher 😂

PR #585, opened 2015-11-17
@jtotoole Render time and use scheduled_publish_at when setting publishing

kanaabe:


scripts/scheduled_posts.coffee
```diff
@@ -15,21 +15,18 @@ switch process.env.NODE_ENV
...
+  if !doc
+    debug 'Scheduled publication finished'
     process.exit()
-)
+  if err
+    exit err
+  doc.published = true
+  doc.published_at = moment(doc.scheduled_publish_at).toDate()
+  doc.scheduled_publish_at = null
+  db.articles.save doc
```
kanaabe: Using `$set` with `Date()` sets the value of the document to the JS Date format. However in mongo, dates should be stored as the ISO string format. The problem was that some articles that were scheduled couldn't be queried by date because their `published_at` was in JS Date format. 

If the mongojs wrapper is used, it's fine to save using JS Date and the wrapper will convert and save it for us, but if you're using $set, it just sets it to the string without a transform, it seems. 

Another small change is to use `doc.scheduled_publish_at` as the new `published_at` date. Unfortunately with mongo, you can't perform an `update` while using the values of the current doc, you have to use a function transform instead which is why i'm doing `forEach`.

jtotoole:
:+1: :+1:  Thanks Kana, glad to have your sharp eye on this!

scripts/scheduled_posts.coffee
craigspaeth: :-o that's odd that it won't typecast using `$set` but will using `save` that seems like a bug in mongojs or the Node mongodb driver. Given the latter issue though, I think this is a great way to kill two birds with one stone.

Solid! :moyai: :boom: :bird: :bird: :dash: :ghost: :ghost:

craigspaeth: Also food for thought: Given the non-blocking nature of Node we could also just wrap this script in a function call and use setInterval (essentially what [node-cron](https://github.com/ncb000gt/node-cron) does):

````
setInterval publishScheduled, 10000
````

To run this every 10 seconds (or w/e tighter interval) and that might allow using straight `update` call. I'm not recommending that—just wanted to point out how you can think about asynchronous operations outside of the typical request-response code.

jtotoole: A plus of that approach would be to allow for more precise scheduled publication (up to the minute rather than just :10 past the hour), correct?

craigspaeth: Yep yep—and less :dollar: from using Heroku scheduler.

kanaabe: Oh that's a cool idea!

PR #588, opened 2015-11-24
@kanaabe: let admins set published_at date

jtotoole:


kanaabe:
Looks good! Thanks for chatting it over with me.

PR #601, opened 2015-11-25
@craigspaeth Rich Embed for Admin

kanaabe:
Paste a link and it will first try to use one of Embedly's [content providers](http://embed.ly/providers) (Google Maps, Slideshare, Soundcloud + 300 others) and if it doesn't match any, then it will try to use the link as an iframe src. The latter is ideal for handling interactive graphics or special html pages we create with [Vanity](vanity.artsy.net). All of this is admin only. Other users won't see the Embed icon in the toolbox. 

embed

### DIFF
--- a/client/apps/edit/components/section_embed/index.coffee
+++ b/client/apps/edit/components/section_embed/index.coffee
@@ -16,50 +16,41 @@ module.exports = React.createClass
 
   getInitialState: ->
     errorMessage: ''
-    src: ''
+    url: ''
     embeddable: ''
     iframe: ''
     height: ''
 
   componentDidMount: ->
-    src = @props.section.get('src')
+    url = @props.section.get('url')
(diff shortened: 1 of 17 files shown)


client/apps/edit/components/section_embed/index.coffee
```diff
@@ -0,0 +1,123 @@
+#
+# Artworks section that shows artwork images in various layouts. A user can
```
craigspaeth: Looks like some copypasta'ed comments you might want to update.

craigspaeth:
Very nicely done! This is super slick with the Embedly integration! Code looks clean and clear, no blockers there. :clap:

craigspaeth:
Oh and note for myself: I should probably follow up with a better icon fort that overflow width layout.

PR #602, opened 2015-11-30
@craigspaeth Adds Super Article admin - partnerlogo/link/related articles

kanaabe:
#### TODO 
- Still need a way to relate the related articles back to the super article. When a Super Article is saved, we should run some background job that updates the related articles since they will need TOC info and partner logos. These related articles should probably also have some additional marker such as `has_super_article: '5515bea68ab80c060003f630'`. Feel free to comment on what kind of data would be helpful to you!

- When enable is clicked, the content section should turn red if the fullscreen background is not yet added. Will do this when implementing FS-BG

#### Below is how the data in this section is saved
```
"super_article": {
   "partner_link": "htphldfgs",
   "partner_logo": "https://artsy-media.png",
   "secondary_partner_logo": "https://artsy-media.png",
   "related_articles": [
      "5515bea68ab80c060003f630",
      "55021d69e877fa0600014e1d"
   ]
},
"is_super_article": false
```

`is_super_article` might not be necessary as a boolean, but I think it will make the front-end easier to deal with. 

#### UI Example
super_article_admin

### DIFF
--- a/api/apps/articles/model.coffee
+++ b/api/apps/articles/model.coffee
@@ -98,6 +98,12 @@ inputSchema = (->
     author: @string().allow('',null)
     credit_line: @string().allow('',null)
     credit_url: @string().allow('',null)
+  is_super_article: @boolean().default(false)
+  super_article: @object().keys
+    partner_link: @string().allow('',null)
+    partner_logo: @string().allow('',null)
+    secondary_partner_logo: @string().allow('',null)
+    related_articles: @array().items(@objectId()).allow(null)
 ).call Joi
 
 querySchema = (->
--- a/api/apps/articles/test/model.coffee
+++ b/api/apps/articles/test/model.coffee
@@ -343,7 +343,7 @@ describe 'Article', ->
         article.published_at.should.be.an.instanceOf(Date)
         moment(article.published_at).format('YYYY').should
           .equal moment().format('YYYY')
-        done() 
+        done()
 
     it 'updates published_at when admin changes it', (done) ->
       Article.save {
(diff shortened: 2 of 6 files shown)


craigspaeth:
Nice!! Clean & clear code and whipped up pretty fast to boot! I have no comments on the code itself, I think the data modeling makes sense and neat to see how little code it is overall. Well done!

I actually think your TODOs are just nice-to-haves, and you can feel free to move on to the next item.

To the first point about denormalizing super article data back on to the related articles:

I think this might actually a premature optimization. Denormalizing like this makes sense for cases where the related data is needed more often than not, e.g. authors & contributing authors whose names show up in every view of an article—thumbnail or otherwise. However in this case we will likely only access the super article data in two ways 1. From the view of the super article itself and 2. In the header/footer of a related article page. In both cases I think we can start by having the client conditionally fetch [like we do for related editorial sections](https://github.com/artsy/force/blob/master/models/article.coffee#L53) and if that turns out to be too slow we can decide to do a Positron-api-level join or background denormalize like you mentioned. It's typically better to avoid denormalizing one-to-N relationships at first b/c it can be harder to maintain and is essentially a form of caching—and mo' cache mo' problems as I always say... or [as MongoDB themselves say:](http://blog.mongodb.org/post/88473035333/6-rules-of-thumb-for-mongodb-schema-design-part-3)

> You’d do this only for fields that are frequently read, get read much more often than they get updated, and where you don’t require strong consistency, since updating a denormalized value is slower, more expensive, and is not atomic.

To the second point about providing UI feedback validating saving a super article:

I like the UX consideration here! This is definitely a nice feature to save editorial from saving an incomplete super article, but IMO not worth it since super articles are admin only features anyways. That said, more power to you if you find a solution quick—I just wouldn't spend too much time on it.

:clap: Nice work!

PR #603, opened 2015-12-02
@zamiang Adds is_super_article query param

kanaabe:
with mongo update
`db.getCollection('articles').update({is_super_article: null},{$set:{is_super_article:false}},{multi:true})`


zamiang:
:+1:

kanaabe:
merging this!

Issue #604, opened 2015-12-03
Query for all articles apart of super article


craigspaeth:
Following up on our chat, I think it might be most useful and still a little more REST-kosher to add an endpoint to the root articles route that returns all of the articles that are a part of the super article. It should act something like this:

```
GET /api/articles?super_article_family_for=<Any of the following article IDs>
=> Returns...
[
  { is_super_article: true, title: 'The Year in Art' ... },
  { is_super_article: false, title: 'Top 20 Living Artists Today' ... },
  { is_super_article: false, title: 'The 20 Top Emerging Artists' ... },
  { is_super_article: false, title: '...t' ... },
  { is_super_article: false, title: '...' ... },
  { is_super_article: false, title: '...' ... }
]
```

That would involve some fancy query building like we [do for `all_by_author`](https://github.com/artsy/positron/blob/master/api/apps/articles/model.coffee#L168). Maybe something like...

````coffeescript
query.$or ?= []
query.$or.push {...}, {...} if input.all_by_author
query.$or.push(
  { _id: ObjectId(input.super_article_family_for) }
  { 'super_article.related_articles': ObjectId(input.super_article_family_for) }
) if input.super_article_family_for
````

@zamiang @kanaabe

**Comment by kanaabe** [comment]
Super :muscle: Family :family:

**Comment by zamiang** [comment]
@kanaabe mind if I take this one? Curious about adding to the positron backend.

**Comment by zamiang** [comment]
@craigspaeth I don't really see a good way to get the whole super family without multiple queries. But! Do have an idea that is kinda the…family spirit.

```
query.$or.push(
  { _id: ObjectId(input.super_article_family_for) }
  { 'super_article.related_articles': ObjectId(input.super_article_family_for) }
```
Will return the searched for article (via _id) and any super articles with the searched article as a related article.

Once we have that super article, we can go and fetch each individual article -- but that would have to be another query.

Personally, I feel like we should just go with 
```
query.$or.push(
  { 'super_article.related_articles': ObjectId(input.super_article_for) }
```
And then have clients fetch the related articles by passing all of the related articles in one query. Perhaps adding `ids` (an array of ids) to the `inputSchema`.

craigspaeth:
[...]

PR #605, opened 2015-12-03
@craigspaeth Super article for article

zamiang:
Discussion in #604 

This simply returns the super article for a given article. 

I really don't see a clear path for getting the whole family without a big if-else within article.where (https://github.com/artsy/positron/blob/master/api/apps/articles/model.coffee#L135) -- which is pretty nice and readable. Up for keeping it pretty simple/basic first? If this has any performance issues in practice, we certainly know what to do.

### DIFF
--- a/api/apps/articles/model.coffee
+++ b/api/apps/articles/model.coffee
@@ -165,10 +165,14 @@ toQuery = (input, callback) ->
     query.auction_id = ObjectId input.auction_id if input.auction_id
     query.section_ids = ObjectId input.section_id if input.section_id
     query.biography_for_artist_id = ObjectId input.biography_for_artist_id if input.biography_for_artist_id
+    query.featured_artwork_ids = ObjectId input.artwork_id if input.artwork_id
     query.tags = { $in: input.tags } if input.tags
 
+    # Convert query for super article for article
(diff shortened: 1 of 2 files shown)


api/apps/articles/model.coffee
```diff
@@ -165,17 +166,28 @@ toQuery = (input, callback) ->
...
-    query.$or = [
+    query.$or.push(
       { primary_featured_artist_ids: ObjectId(input.artist_id) }
       { featured_artist_ids: ObjectId(input.artist_id) }
       { biography_for_artist_id: ObjectId(input.artist_id) }
-    ] if input.artist_id
+    ) if input.artist_id
+
+    # Convert query for super article for article
+    query.$or.push(
```
craigspaeth: :+1: works for me—but now that you're just doing one condition I don't think you need to loop this in to $or, you can build on the query similar to how `biography_for_artist_id` does it.

`query['super_article.related_articles']= ObjectId(input.super_article_for) if input.super_article_for`

craigspaeth:
:+1: Happy to merge this, but would appreciate taking a stab at the comment above first.

zamiang:
Cleaned up! I did keep the line that breaks out an array for $or since it seems to make the following two cases (articles by author, articles feature to) a bit easier to read.

PR #606, opened 2015-12-04
@kanaabe Fix super_article_for query to work in production as well as in tests

zamiang:
So, the previous solution only actually worked in tests :embarrassed:. I did not realize that `related_articles` is actually an array of strings -- not an array of ObjectIds. So, by using an objectId in the query and the test, I was tricking myself into thinking that this worked. This pull request makes the feature actually work.

Also want to flag in case the ObjectId vs String isn't intentional.


kanaabe:
`related_articles` _should_ be an array of ObjectIds..https://github.com/artsy/positron/blob/master/api/apps/articles/model.coffee#L106

kanaabe:
Closing this -- related_articles aren't currently ObjectIds but they should be, sorry about that! Fix is coming soon.

Issue #592, opened 2015-11-24
Super Article Logo(s)


owendodd:
Super Articles should always have the full Artsy logo, and can optionally have a second logo for partnerships. This logo should be scaled to a max height of 80px and max width of 200px. 

![screen shot 2015-11-24 at 6 45 54 pm](https://cloud.githubusercontent.com/assets/3196536/11384171/a009fbb4-92db-11e5-9185-09b91389dfd8.png)

![screen shot 2015-12-02 at 2 36 35 pm](https://cloud.githubusercontent.com/assets/3196536/11542346/fa05b69c-9904-11e5-94cd-6f0999ca33f4.png)

<img width="649" alt="screen shot 2015-11-29 at 6 55 12 pm" src="https://cloud.githubusercontent.com/assets/3196536/11461269/96cb4436-96cd-11e5-96e5-e3b2b6461bb0.png">

owendodd:
This is controlled in the Admin panel of an article:

<img width="862" alt="screen shot 2015-11-26 at 10 33 47 am" src="https://cloud.githubusercontent.com/assets/3196536/11426784/5118f4d6-942c-11e5-8d6b-25b3fa4c5027.png">
<img width="863" alt="screen shot 2015-11-26 at 10 37 08 am" src="https://cloud.githubusercontent.com/assets/3196536/11426785/511ba8e8-942c-11e5-908f-ade02c12a30a.png">

owendodd:
Updated slightly. (2 additional fields)

![screen shot 2015-12-02 at 2 36 11 pm](https://cloud.githubusercontent.com/assets/3196536/11542339/f1d55a54-9904-11e5-9d13-5d7596330806.png)

owendodd:
Also should note that I have been imagining partner logo 1 as an svg so we can use it in white/black. can we currently do that?

zamiang:
Should the secondary partner logo also have a link?

owendodd:
:)

owendodd:
haha, they should actually both have links as I think the partner logo would link to somewhere different than the partner link.... so many fields!

owendodd:
Ok this should actually be all the fields. It would be nice if we could hide these until you hit the 'Enable' toggle.

![screen shot 2015-12-02 at 3 59 50 pm](https://cloud.githubusercontent.com/assets/3196536/11544121/29f94694-990e-11e5-8a13-937d81774e70.png)

Issue #596, opened 2015-11-25
Super Article Partner Link


owendodd:
Super articles should have an optional partner link field, which would show up in the expanded header (https://github.com/artsy/force/issues/3818 )

![screen shot 2015-11-24 at 7 02 10 pm](https://cloud.githubusercontent.com/assets/3196536/11384539/9a288262-92de-11e5-9d6c-cbabed492d51.png)

owendodd:
Partner Links would input in the admin panel:

<img width="862" alt="screen shot 2015-11-26 at 10 33 47 am" src="https://cloud.githubusercontent.com/assets/3196536/11426798/5ff0941e-942c-11e5-85dc-e253fe66f019.png">
<img width="863" alt="screen shot 2015-11-26 at 10 37 08 am" src="https://cloud.githubusercontent.com/assets/3196536/11426797/5fee7404-942c-11e5-8428-02994c0872cb.png">

craigspaeth:
I think this also needs a "Partner Link Copy" field to fill in the "Download the Planet Art App" part.

zamiang:
Couple additions to make it more reusable

- The + should be included in the Planet Art logo image (it is styled very specific to the style of the Planet Art logo)
- The secondary partner logo should have some text associated with it like "In partnership with" since that may not be something consistent across future logos in super posts.

![screenshot 2015-12-02 12 40 26](https://cloud.githubusercontent.com/assets/110563/11538886/104941f2-98f2-11e5-83f7-0d68ad26c889.png)
)

owendodd:
Here is the admin implications of both of the above comments. Two new fields "Partner Link Title" and "Secondary Logo Text"


![screen shot 2015-12-02 at 2 36 11 pm](https://cloud.githubusercontent.com/assets/3196536/11541965/244ca138-9903-11e5-8b87-476145d7490b.png)

Issue #593, opened 2015-11-24
Super Article TOC


owendodd:
SuperArticles should have the ability to have "sub-articles" that would show up in a table of contents (however, super articles aren't required to have a TOC/sub articles and can live as a single page)
This table of contents would show up at the top of the overview:
<img width="949" alt="screen shot 2015-11-29 at 7 17 34 pm" src="https://cloud.githubusercontent.com/assets/3196536/11461282/e37d97e8-96cd-11e5-9e41-9f096bf88e51.png">

In the header on the overview and all sub articles:
![screen shot 2015-11-24 at 6 48 28 pm](https://cloud.githubusercontent.com/assets/3196536/11384256/6c2cb100-92dc-11e5-908a-ed72f84a529a.png)

And in the footer of the header and all sub articles: 
![screen shot 2015-12-02 at 2 36 35 pm](https://cloud.githubusercontent.com/assets/3196536/11542403/2f78b1e4-9905-11e5-9451-57d7012bade2.png)

owendodd:
This is controlled in the Admin panel of an article:

<img width="862" alt="screen shot 2015-11-26 at 10 33 47 am" src="https://cloud.githubusercontent.com/assets/3196536/11426794/5bbdee6e-942c-11e5-9415-e36b7619c20c.png">
<img width="863" alt="screen shot 2015-11-26 at 10 37 08 am" src="https://cloud.githubusercontent.com/assets/3196536/11426793/5ba01b50-942c-11e5-9804-2113071eaa92.png">

owendodd:
Updated here:

![screen shot 2015-12-02 at 3 59 50 pm](https://cloud.githubusercontent.com/assets/3196536/11544177/7b8fc91a-990e-11e5-8f73-a9f3566fdf74.png)

PR #608, opened 2015-12-04
@zamiang Adds additional super article partner fields

kanaabe:
Closes https://github.com/artsy/positron/issues/592 https://github.com/artsy/positron/issues/593 https://github.com/artsy/positron/issues/596


zamiang:
Looks good to me :+1:

Issue #607, opened 2015-12-04
Fix related article types


kanaabe:
Looks like the related_articles array is saving as strings instead of ObjectIds which is problematic for querying and finding article families :family: :broken_heart: 

![2015-12-04_1100](https://cloud.githubusercontent.com/assets/2236794/11594323/5675d506-9a77-11e5-8dd6-7f0e3d30d24d.png)

PR #610, opened 2015-12-04
@zamiang Type cast objectid on related_articles

kanaabe:
Closes https://github.com/artsy/positron/issues/607
Totally forgot to do this... -.- 

Anyway this will map the ObjectId over `super_article.related_articles` when saving.


zamiang:
Ah! Interesting. Should it also have a test :grin: ?

zamiang:
But ya looks good to go - seems like this was just a regular thing for adding models that was missed.

kanaabe:
Yep, we do the type casting because of this funny little issue: https://github.com/pebble/joi-objectid/issues/2#issuecomment-75189638

Good call, test coming!

zamiang:
Seems like the build is stuck or something?

PR #626, opened 2015-12-14
@craigspaeth: [WIP/don't merge] script to add artist links

jtotoole:
Still need to test locally but would appreciate some quick feedback if you've got it!


scripts/link_artists.coffee
```diff
@@ -0,0 +1,61 @@
...
+      # replace first mention of name in text w/ name + artsy link
+      if findUnlinked article.lead_paragraph, artist
+        addLink article.lead_paragraph, artist, article
+      else
+        for section in article.sections when section.type is 'text'
+          if findUnlinked section.body, artist
+            addLink section.body, artist, article
+
+findUnlinked = (text, artist) ->
+  nameLink = artist.name + '</a>' #any way to exclude google hrefs from check?
```
jtotoole: In a number of cases there are links from artist names that go to Google search results for that artist on Artsy as opposed to the Artsy page itself; wasn't quite sure of how to exclude those links. We don't want to override any other kinds of existing links.

scripts/link_artists.coffee
```diff
@@ -0,0 +1,61 @@
...
+  return true if (_s.count text, nameLink) is 0 && (_s.count html, artist.name) > 0
+
+addLink = (text, artist, article) ->
+  # an absurd means of attempting to ignore possessive mentions
+  text.replace (artist.name if text.charAt(text.indexOf(artist.name) + (artist.name).length) is not '\''), "<a href=\"https://artsy.net/artist/#{artist.id}\">" + artist.name + "</a>"
```
jtotoole: Trying to ignore possessive mentions but not certain this will work

craigspaeth: Hmm hard to tell what's happening here, but you could try `artistName.split("'").split("'")[0]`

scripts/link_artists.coffee
```diff
@@ -0,0 +1,61 @@
...
+      else
+        for section in article.sections when section.type is 'text'
+          if findUnlinked section.body, artist
+            addLink section.body, artist, article
+
+findUnlinked = (text, artist) ->
+  nameLink = artist.name + '</a>' #any way to exclude google hrefs from check?
+  return true if (_s.count text, nameLink) is 0 && (_s.count html, artist.name) > 0
+
+addLink = (text, artist, article) ->
```
craigspaeth: I like the splitting out of functions going on here!

scripts/link_artists.coffee
craigspaeth: I would try extracting the arguments of this function into variables to make this more readable for starters.

```coffeescript
artistName = artist.name if text.charAt(text.indexOf(artist.name) + (artist.name).length) is not '\''
link = "<a href=\"https://artsy.net/artist/#{artist.id}\">" + artist.name + "</a>"
text.replace artistName, link
```

Then to try an only wrap before the apostrophe of artist mentions you can work on `artistName`. `.replace` is only going to replace the first instance of the string if you pass a string as the first argument e.g. `"foo foo".replace('foo', 'bar') => "bar foo"`. What you probably want here is to replace _all_ occurrences of the artist name, which will require a greedy regex e.g.

````coffeescript
artistNameRegex = /// #{artist.name} ///g
link = "<a href=\"https://artsy.net/artist/#{artist.id}\">" + artist.name + "</a>"
text.replace artistNameRegex, link
````

..although still need to figure out how to make that regex not match on names with a following apostrophe

jtotoole: Thanks, good thought re: extraction. Actually do just want the first instance of the name.

scripts/link_artists.coffee
jtotoole: :haircut:

scripts/link_artists.coffee
craigspaeth: Oh cool, in that case :+1:

craigspaeth:
Going to close, feel free to reopen later.

PR #627, opened 2015-12-14
@zamiang More options on video component

kanaabe:
video

### DIFF
--- a/client/apps/edit/components/section_tool/index.coffee
+++ b/client/apps/edit/components/section_tool/index.coffee
@@ -41,6 +41,8 @@ module.exports = React.createClass
         @props.sections.add {
           type: 'video'
           url: ''
+          background_color: ''
+          layout: 'column_width'
         }, at: @props.index + 1
       when 'slideshow'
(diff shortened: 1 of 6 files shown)


zamiang:
:sparkles: :tada: :sparkles: :

client/apps/edit/components/section_tool/index.coffee
```diff
@@ -41,6 +41,8 @@ module.exports = React.createClass
         @props.sections.add {
           type: 'video'
           url: ''
+          background_color: ''
+          layout: 'column_width'
```
zamiang: So `column_width` is the new default - what is the new fullscreen one called?

zamiang: Ah! Looks like `overflow_fillwidth`

kanaabe: Indeed, we use it in at least one other place too, I think artwork sections

zamiang:
Cool! Good to merge and deploy :+1:

PR #630, opened 2015-12-15
@zamiang Adds fullbleed option to embed

kanaabe:

### DIFF
--- a/client/apps/edit/components/section_embed/index.styl
+++ b/client/apps/edit/components/section_embed/index.styl
@@ -85,3 +85,14 @@
     opacity 1
   .embed-container
     width 1060px
+
+.edit-section-container[data-layout=overflow_fillwidth][data-type=embed]
+  width 1100px
+  left -280px
+  input
+    width 519px
+  iframe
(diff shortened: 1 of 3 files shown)


client/apps/edit/components/section_embed/index.styl
```diff
@@ -85,3 +85,14 @@
...
     width 1060px
+
+.edit-section-container[data-layout=overflow_fillwidth][data-type=embed]
+  width 1100px
+  left -280px
+  input
+    width 519px
+  iframe
+    width calc(100vw - 110px)
+    margin-left calc(1220px + ((100vw - 1270px)/2) - 100vw)
```
kanaabe: So this calculation is kind of insane and I think the real solution will be to separate each of the containers to control its own margins like what you did with Force but since this needs to be done today, it's basically finding the negative distance here:

zamiang: haha yea that makes sense. is cray tho!

zamiang:
ok! Looks great :+1:

PR #636, opened 2015-12-18
@jtotoole Checks if name exists before trying to render it

kanaabe:
When `user.get('name')` was null, it would break the page. :0


jtotoole:
:white_check_mark: :white_check_mark:

PR #643, opened 2015-12-29
@craigspaeth Check that video URL is properly formatted when saving

kanaabe:
We need to add `http://` to URLs if the user gives us a link like `youtube.com?v=123`. The reason being -- `embed-video` uses `URL.parse.hostname` which apparently is `null` when there is no protocol provided. 

Related issue here: https://github.com/artsy/force/issues/4020#event-501401794

### DIFF
--- a/client/apps/edit/components/section_video/index.coffee
+++ b/client/apps/edit/components/section_video/index.coffee
@@ -41,7 +41,9 @@ module.exports = React.createClass
       @props.section.set cover_image_url: @state.coverSrc
 
   onChangeUrl: (e) ->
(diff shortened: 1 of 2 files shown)


craigspaeth:
We do this [for links at the api-level](https://github.com/artsy/positron/blob/master/api/apps/articles/model.coffee#L355), I wonder if it makes sense to do the same here. I'll leave that as a maybe-nice-to-have though. :+1: works for me!

kanaabe:
Oh!! Much better idea. I'll revise soon.

Issue #624, opened 2015-12-14
Handle adding a section above a callout better


kanaabe:
Difficult to add a section above a callout section, nav gets pushed down

![2015-12-14_1313](https://cloud.githubusercontent.com/assets/2236794/11789456/7fe32912-a264-11e5-94c2-513bd2efc755.png)

Issue #622, opened 2015-12-12
Clean up section toolbar


kanaabe:
![](https://cloud.githubusercontent.com/assets/3196536/11646354/3b11a0fe-9d2b-11e5-836e-f7af2f803dd4.png)

![](https://cloud.githubusercontent.com/assets/3196536/11646355/3b11eb54-9d2b-11e5-9e47-1b817600b850.png)

![](https://cloud.githubusercontent.com/assets/3196536/11646353/3ad37a86-9d2b-11e5-822a-b0764286ee40.png)

PR #647, opened 2016-01-08
@jtotoole Cleaner toolbar and handle adding a section above callout better

kanaabe:
Closes https://github.com/artsy/positron/issues/624 https://github.com/artsy/positron/issues/622


jtotoole:
:dizzy:

PR #649, opened 2016-01-13
@craigspaeth Be able to select fullscreen image

kanaabe:
fs-image


client/apps/edit/components/section_fullscreen/index.coffee
```diff
@@ -66,7 +68,10 @@ module.exports = React.createClass
       add: (src) =>
         @setState progress: 0.1
       done: (src) =>
-        @setState background_url: src, progress: null
+        if src.indexOf('.mp4') > 0
```
craigspaeth: :+1: there's also [`path.extname`](https://nodejs.org/api/path.html#path_path_extname_p) when you need to be real sure.

craigspaeth:
Solid!

PR #652, opened 2016-01-20
@craigspaeth Send article updates to Sailthru

kanaabe:
If an editor wants to change the email section's credit line, headline, author, or whatever, we want to send those updates to Sailthru for emails. 

I'll comment below on questions I have!

### DIFF
--- a/api/apps/articles/model.coffee
+++ b/api/apps/articles/model.coffee
@@ -15,8 +15,9 @@ xss = require 'xss'
 cheerio = require 'cheerio'
 url = require 'url'
 request = require 'superagent'
+debug = require('debug') 'api'
 { ObjectId } = require 'mongojs'
-{ ARTSY_URL, API_MAX, API_PAGE_SIZE, SAILTHRU_KEY, SAILTHRU_SECRET, API_URL, EMBEDLY_KEY, FORCE_URL } = process.env
+{ ARTSY_URL, API_MAX, API_PAGE_SIZE, SAILTHRU_KEY, SAILTHRU_SECRET, API_URL, EMBEDLY_KEY, FORCE_URL, ARTSY_EDITORIAL_ID, NODE_ENV } = process.env
 sailthru = require('sailthru-client').createSailthruClient(SAILTHRU_KEY,SAILTHRU_SECRET)
 { crop } = require('embedly-view-helpers')(EMBEDLY_KEY)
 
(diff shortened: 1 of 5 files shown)


api/apps/articles/model.coffee
```diff
@@ -385,9 +387,40 @@ typecastIds = (article) ->
...
+sendArticleToSailthru = (article, cb) =>
+  images = {}
+  tags = article.keywords.concat ['article']
+  tags = tags.concat ['artsy-editorial'] if article.author_id is "503f86e462d56000020002cc"
+  tags = tags.concat ['magazine'] if article.featured is true
+  if article.email_metadata?.image_url
+    images =
+      full: url: crop(article.email_metadata.image_url, { width: 1200, height: 706 } )
+      thumb: url: crop(article.email_metadata.image_url, { width: 900, height: 530 } )
+  if NODE_ENV is 'production'
```
kanaabe: Want to limit this to production so `http://staging.artsy.net/article/*` doesn't get added on saves.

api/apps/articles/model.coffee
```diff
@@ -385,9 +387,40 @@ typecastIds = (article) ->
...
+      return cb(err) if err
+      cb(null)
+
 sanitizeAndSave = (callback) -> (err, article) ->
   return callback err if err
-  db.articles.save sanitize(typecastIds article), callback
+  # Send new content call to Sailthru on any published article save
+  if article.published
+    sendArticleToSailthru article, (err) =>
+      console.log 'Log an error' if err
```
kanaabe: How would I log the error here?

client/apps/edit/components/admin/index.coffee
```diff
@@ -457,8 +457,8 @@ module.exports = class EditAdmin extends Backbone.View
     clientFormat = moment(date).format('L')
     publishTime = moment(date).format('HH:mm')
     @saveFormat = moment(date).toDate()
-    $('.edit-admin-input-date').val(clientFormat)
-    $('.edit-admin-input-time').val(publishTime)
+    $('#edit-admin-publish-date input').val(clientFormat)
+    $('#edit-admin-schedule-publish-time input').val(publishTime)
```
kanaabe: This stuff is unrelated - I'm trying to track down a bug where an 'Invalid Date' gets appended to the tags/social share section. Trying to be more specific with the inputs.

api/apps/articles/model.coffee
craigspaeth: Yeah I think this is the way to go then. Ideally there'd be a "staging" version of Sailthru and you'd set the keys above to be different. However, without that I think this is next best approach.

api/apps/articles/model.coffee
```diff
@@ -385,9 +387,40 @@ typecastIds = (article) ->
     biography_for_artist_id: ObjectId(article.biography_for_artist_id) if article.biography_for_artist_id
     super_article: if article.super_article?.related_articles then _.extend article.super_article, related_articles: article.super_article.related_articles.map(ObjectId) else {}
 
+sendArticleToSailthru = (article, cb) =>
+  images = {}
+  tags = article.keywords.concat ['article']
+  tags = tags.concat ['artsy-editorial'] if article.author_id is "503f86e462d56000020002cc"
```
craigspaeth: Let's set this artsy editorial id as a config variable. Not that I expect it to change, but it'll be easier to read as a constant. (Also should do this in Force/MG if we haven't).

api/apps/articles/model.coffee
craigspaeth: We use the [debug](https://github.com/visionmedia/debug) module to do better logging. [Here's an example](https://github.com/artsy/positron/blob/22dab049e10a68ce3515866fb33f2cee6a9e1eb6/api/index.coffee#L7).

craigspaeth: We actually have a `onPublish` callback already and I think this [belongs there](https://github.com/artsy/positron/blob/master/api/apps/articles/model.coffee#L284). That would allow you to leverage the [`if article.published`](https://github.com/artsy/positron/blob/master/api/apps/articles/model.coffee#L251) logic that already exists.

client/apps/edit/components/admin/index.coffee
craigspaeth: :+1:

craigspaeth:
Looks good! Plz address the comments above and write a test for `sendArticleToSailthru` (you'll want to use [rewire](https://github.com/jhnns/rewire) to stub `sailthru`), then I'm ready to merge when :green_heart: !

api/apps/articles/model.coffee
kanaabe: Initially that's where I wanted to hook into but often times an editor will save article data after it's already been published. We want to send the article to Sailthru anytime it's being saved, as long as its published.

api/apps/articles/model.coffee
kanaabe: G8 idea - it definitely should be configurable.

kanaabe:
Testing on my local now and getting some error so please hold on merging!

api/apps/articles/test/integration.coffee
```diff
@@ -1,18 +1,22 @@
...
-app = require '../../../'
+sinon = require 'sinon'
 request = require 'superagent'
 { ObjectId } = require 'mongojs'
 
 describe 'articles endpoints', ->
 
   beforeEach (done) ->
+    @Article = require '../model'
+    sinon.stub(@Article, 'sendArticleToSailthru').yields()
```
craigspaeth: Nice! I hardly use `yields`, I need to step up my sinon game.

craigspaeth:
:clap: well done on the refactors + tests!

PR #656, opened 2016-01-22
@craigspaeth Fixes sendArticleToSailthru and modify tests

kanaabe:


api/apps/articles/model.coffee
```diff
@@ -390,7 +390,7 @@ typecastIds = (article) ->
     super_article: if article.super_article?.related_articles then _.extend article.super_article, related_articles: article.super_article.related_articles.map(ObjectId) else {}
 
 @sendArticleToSailthru = (article, cb) =>
-  cb null unless NODE_ENV is 'production'
+  return cb() unless NODE_ENV is 'production'
```
craigspaeth: Whoops! Missed that, good catch—always remember to `return callback` to short circuit functions.

kanaabe: Ha so all those errors we were seeing previously had to do with this line. I was just calling back but the rest of the method was still executing. Updated tests so we don't need to actually stub in the integration tests since they have NODE_ENV 'test'

kanaabe:
Thanks!

api/apps/articles/model.coffee
craigspaeth: Ah ha! :-P that makes sense... I though `NODE_ENV` was set to `test`

Issue #658, opened 2016-01-22
Sailthru Content Call QA


kanaabe:
- send after generating slugs
- send after generating keywords
- undefined slug issue (`_.last(article.slugs)` perhaps )

PR #659, opened 2016-01-24
@craigspaeth Fix url on sailthru content call

kanaabe:
Closes https://github.com/artsy/positron/issues/658

The `article.slug` isn't created at this point yet so it was sending 'undefined' for articles that are newly published.


craigspaeth:
:+1:

Issue #666, opened 2016-02-03
Get rid of Brands in writer


owendodd:
![screen shot 2016-02-03 at 1 48 34 pm](https://cloud.githubusercontent.com/assets/3196536/12793145/15aa7eba-ca7d-11e5-982a-5acd28b4264f.png)

thoughts? cc @alexanderfabry @halleyjohnson

craigspaeth:
:+1: We can always dig up deleted code thanks to git. IMO we should never keep something around if the only reason is because "we might need to bring it back later".

PR #667, opened 2016-02-03
@craigspaeth Removes Brand Partners

kanaabe:
Closes https://github.com/artsy/positron/issues/666


craigspaeth:
Nice! Kill it with :fire: :fire: :fire:

PR #668, opened 2016-02-03
@craigspaeth: James' linking script

craigspaeth:


scripts/artist_ids.coffee
```diff
@@ -0,0 +1,612 @@
+module.exports = 
+
+ARTISTIDS = [
```
dylanfareed: I think we can prolly just re-use the one file between this and https://github.com/artsy/positron/pull/668/files#diff-9910aafd47cba1f0a8336ed3b16f537cR3

craigspaeth: :+1: good catch!

PR #674, opened 2016-03-01
@craigspaeth Checks that href exists before trying to sanitize it

kanaabe:
This caused Positron to crash at 2:36PM today. `url.parse` doesn't fail gracefully so we should check that `$(this).attr 'href'` exists before trying to sanitize it. Weird that a link in the HTML didn't have an `href` field in the first place though.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,10 @@ BIN = node_modules/.bin
 s:
 	$(BIN)/coffee index.coffee
 
+# Start the server using forever
+sf:
+	$(BIN)/forever $(BIN)/coffee --nodejs --max_old_space_size=960 index.coffee
+
 # Run all of the project-level tests, followed by app-level tests
 test: assets
 	$(BIN)/mocha $(shell find api/test -name '*.coffee' -not -path 'test/helpers/*')
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ s:
 
 # Start the server using forever
 sf:
-	$(BIN)/forever $(BIN)/coffee --nodejs --max_old_space_size=960 index.coffee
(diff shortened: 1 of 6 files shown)


craigspaeth:
We should also definitely put [forever](https://github.com/foreverjs/forever) in our `make s` command like we do [on Force](https://github.com/artsy/force/blob/master/Makefile#L17). I'm kinda shocked it's taken this long to notice we need that—it's pretty standard for any Node production setup.

craigspaeth:
Code looks good! However, I'd prefer we write a test for this to prevent a regression.

kanaabe:
Added a test and make sf...oy it really is amazing how long we have not had this for.

craigspaeth:
Solid! One last thing—make sure to add `make sf` [to the Procfile](https://github.com/artsy/positron/blob/master/Procfile#L1) so heroku acknowledges it as the command to run the production server with. Then good to merge!

kanaabe:
Oh oops! Done.

Makefile
```diff
@@ -12,6 +12,10 @@ BIN = node_modules/.bin
 s:
 	$(BIN)/coffee index.coffee
 
+# Start the server using forever
+sf:
+	$(BIN)/forever $(BIN)/coffee --nodejs --max_old_space_size=960 index.coffee
```
craigspaeth: You can take the `--max_old_space_size=960` flag off here. That's an artifact of debugging the memory leak on Force. Luckily Positron averages less than a quarter of this memory use anyways so it wouldn't affect anything.

craigspaeth:
Sorry :-P one more thing about the `max_old_space_size`... then it's good to merge!

kanaabe:
Ok removed, merging!

craigspaeth:
:+1:

Issue #655, opened 2016-01-22
Invalid Date appearing on share modules


kanaabe:

kanaabe:
Looks like this was fixed by being more specific by using id [here](https://github.com/artsy/positron/pull/652/files#diff-669449640a696d2950b15f75d7ad5fd5L461)

I haven't seen any `share_description: 'Invalid Date' ` in Robomongo:

![2016-01-25_2107](https://cloud.githubusercontent.com/assets/2236794/12570663/c13b1abe-c3a7-11e5-8628-ba2749596d14.png)


So closing for now but please keep me posted @halleyjohnson if you see this creep up again!

kanaabe:
Now appearing in the tags section as well.

PR #680, opened 2016-03-08
@craigspaeth Validate date on client before trying to save it

kanaabe:
(Hopefully) Closes https://github.com/artsy/positron/issues/655

### DIFF
--- a/client/apps/edit/components/admin/index.coffee
+++ b/client/apps/edit/components/admin/index.coffee
@@ -447,18 +447,20 @@ module.exports = class EditAdmin extends Backbone.View
   setupPublishDate: ->
     $('.edit-admin-input-date, .edit-admin-input-time').on 'blur', =>
       dateAndTime = @$('.edit-admin-input-date').val() + ' ' + @$('.edit-admin-input-time').val()
-      @formatAndSetPublishDate dateAndTime
-      @article.save published_at: @saveFormat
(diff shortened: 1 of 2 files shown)


client/apps/edit/components/admin/index.coffee
```diff
@@ -447,18 +447,20 @@ module.exports = class EditAdmin extends Backbone.View
...
 
-  formatAndSetPublishDate: (date) ->
-    date = new Date(date)
-    clientFormat = moment(date).format('L')
-    publishTime = moment(date).format('HH:mm')
-    @saveFormat = moment(date).toDate()
-    $('#edit-admin-publish-date input').val(clientFormat)
-    $('#edit-admin-schedule-publish-time input').val(publishTime)
+  formatAndSetPublishDate: (date, cb) ->
+    return cb null unless moment(date).isValid()
```
craigspaeth: Doesn't seem like there's anything async happening here, so you could get away with `return unless moment...`. Not a blocker though.

craigspaeth:
Looks good! Appreciate the test.

Might be time to think about breaking this admin view up now that it's reaching 500 lines of code. Feel free to follow up PR with these suggestions, or just keep it in mind for next time you touch this code.

:+1:

client/apps/edit/components/admin/index.coffee
kanaabe: Ah I was overthinking that callback for sure :)

PR #686, opened 2016-03-20
@craigspaeth Adds process.exit instead of db.close in s3 script

kanaabe:
Whether or not this is what's causing the closed connections, we should exit the script with `process.exit`

### DIFF
--- a/scripts/daily_upload_s3.coffee
+++ b/scripts/daily_upload_s3.coffee
@@ -51,6 +51,6 @@ db.articles.find({ published: true })
       'Content-Type': 'text/csv'
     }, (err, result) ->
 
-      # Delete filename and close db
+      # Delete filename and exit
       fs.unlink(dir + filename)
-      db.close()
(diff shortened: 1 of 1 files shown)


scripts/daily_upload_s3.coffee
```diff
@@ -51,6 +51,6 @@ db.articles.find({ published: true })
       'Content-Type': 'text/csv'
     }, (err, result) ->
 
-      # Delete filename and close db
+      # Delete filename and exit
       fs.unlink(dir + filename)
-      db.close()
+      process.exit(1)
```
craigspaeth: This should just be `process.exit()`—add the 1 indicates it's exiting with an error, which in this case we want to indicate an intentional exit.

craigspaeth:
I'm actually [pretty sure this is unrelated](https://artsy.slack.com/archives/web/p1458494059000124), let's not bother with it. Thanks for the PR anyways!

Issue #683, opened 2016-03-15
Featured Artists prevents save when artist name can't be truncated


kanaabe:
![image](https://cloud.githubusercontent.com/assets/2236794/13764238/ed79213c-ea23-11e5-983f-f9aa164c3bab.png)

PR #688, opened 2016-03-22
@craigspaeth Bug fixes for automated emails prep

kanaabe:
Closes https://github.com/artsy/positron/issues/683 https://trello.com/c/snJPic41/123-bugs-to-fix-for-automating-emails

- New text field for custom text that gets sent to ST
- Refactor Scheduled posts
- Use Xapp token for Gravity fetches
- Fixes label in featured artists/artworks in admin section 
- Saves email metadata on publish instead of autopopulating it on client


api/apps/articles/model/retrieve.coffee
```diff
@@ -47,6 +48,10 @@ schema = require './schema'
 
     # Allow regex searching through the q param
     query.thumbnail_title = { $regex: new RegExp(input.q, 'i') } if input.q
+
+    # Look for articles with scheduled dates before the given date
+    query.scheduled_publish_at = { $lt: moment(input.scheduled_publish_at).toDate() } if input.scheduled_publish_at
```
craigspaeth: Adding this to `where` exposes it in the API—which is fine, but I might call this `scheduled_publish_before` for clarity of end-user sake. You can also just use `db.find({ $lt ... })` instead of `@where` inside of `publishScheduledArticles` instead.

api/apps/articles/model/index.coffee
```diff
@@ -37,27 +39,46 @@ retrieve = require './retrieve'
...
+      do (article) =>
+        callbacks.push (callback) =>
+          @save {
+            id: article.id.toString()
+            author_id: article.author_id.toString()
+            published: true
+            published_at: moment(article.scheduled_publish_at).toDate()
+            scheduled_publish_at: null
+          }, 'foo', callback
+    async.parallel callbacks, (err, results) ->
```
craigspaeth: Might be more useful to use `async.map` e.g.

````coffeescript
async.map results.results, (article, callback) ->
, (err, results) ->
````

File: api/apps/articles/model/save.coffee [hunk 3]
```diff
@@ -31,11 +32,20 @@ sailthru = require('sailthru-client').createSailthruClient(SAILTHRU_KEY,SAILTHRU
...
-@onPublish = (article, author, accessToken, cb) =>
-  if not article.published_at
-    article.published_at = new Date
+@onPublish = (article, author, cb) =>
+  unless article.published_at
+    article.published_at = new Date()
+  setEmailFields article, author
   @generateSlugs article, author, cb
 
+setEmailFields = (article, author) =>
```
craigspaeth: :+1: solid

scripts/scheduled_posts.coffee
```diff
@@ -1,9 +1,5 @@
-Q = require 'bluebird-q'
```
craigspaeth: Nice to see this script cleaned up from the model-level work!

craigspaeth:
Solid! Look at all those features in a mere net 2 lines of code! `+151 −149` I don't have any real blockers on my end, but I'll let you sit on my comments about `where` and `async.map` and merge yourself when ready.

:clap: nice work as always!

kanaabe:
Thanks! Yeah I was contemplating using `where` vs `db.find` but it's true that we don't need to expose the query for any reason since the scheduler runs internally. Good call! I'll take a look at using `async.map` too.

kanaabe:
Margin'!

PR #689, opened 2016-03-23
@craigspaeth Adds sections to input in scheduler

kanaabe:
`input` has some dependencies when it comes to the contents in it (author_id, id) -- but I seemed to have left out `sections`. This was causing scheduled articles to save with no sections! :(

### DIFF
--- a/api/apps/articles/model/save.coffee
+++ b/api/apps/articles/model/save.coffee
@@ -107,6 +107,7 @@ setEmailFields = (article, author) =>
   intersection = _.intersection(before, after)
   if input.sections and input.sections.length > 0
     article.sections = input.sections
+  return cb null, article unless input.sections
   return cb(null, article) if intersection.length is before.length and intersection.length is after.length
   # Try to fetch and denormalize the artworks from Gravity asynchonously
   artworkIds = _.pluck (_.where input.sections, type: 'artworks' ), 'ids'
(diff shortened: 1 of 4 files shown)


craigspaeth:
Ah :disappointed: that's no good. I think ideally we'd remove the `_.omit(input, 'sections')` [here](https://github.com/artsy/positron/blob/master/api/apps/articles/model/save.coffee#L179)—it's a bad API behavior to remove sections if one doesn't send them in the body. Sorry for not catching that in code review.

craigspaeth:
Just to recap we'll put an `if input.sections` guard [here](https://github.com/artsy/positron/blob/master/api/apps/articles/model/save.coffee#L108) and a test to ensure `save`ing without `sections` will leave the data alone.

kanaabe:
Looks like it won't _quite_ work with just that. This line still depends on input.sections too: https://github.com/artsy/positron/blob/master/api/apps/articles/model/save.coffee#L180

I switched to onPublish on the scheduler which works since we're not dealing with input, only thing is, it uses the article's author instead of a fetched one.

api/apps/articles/model/save.coffee
```diff
@@ -177,7 +177,7 @@ getPartnerLink = (artwork) ->
     return cb err if err
     publishing = (input.published and not article.published) or (input.scheduled_publish_at and not article.published)
     article = _.extend article, _.omit(input, 'sections'), updated_at: new Date
-    article.sections = [] unless input.sections?.length
+    article.sections = [] unless input.sections.length > 0
```
craigspaeth: I think you still want the `?` guard here in case a client doesn't send `sections`.

craigspaeth:
Even better! Would still like a test to ensure `save`ing without `sections doesn't empty them out :grin:

kanaabe:
:+1: definitely!

kanaabe:
Ok check it out! @craigspaeth

craigspaeth:
Solid, well done.

kanaabe:
Weird, I think this deploy brought staging's db down. Wonder why it's not recovering. I'm going to hold on on deploying to prod until tomorrow.

Issue #695, opened 2016-03-29
Investigate artwork selection component not displaying correct image size


kanaabe:
Example artworks: 

https://www.artsy.net/artwork/paul-winstanley-art-school-iii
https://www.artsy.net/artwork/paul-winstanley-art-school-v

These artworks are pulling the square image instead of the large or larger sizes in original asp ratio

This might affect denormalizing so check before backfilling.

PR #699, opened 2016-03-30
@craigspaeth Adds AdditionalImages and fixes Artwork denormalization issue

kanaabe:
Closes https://github.com/artsy/positron/issues/695, https://github.com/artsy/positron/issues/698

This PR adds a model/collection similar to Force's `AdditionalImages`. The issue is that `ArtworkHelper`'s `imageUrl` function doesn't take an array of images, only a single size or 'larger' default which some artworks don't have. If an artwork doesn't have 'larger', it defaults to the first image, like 'square' or 'four_thirds'. Anyway, we want to access `artsy-backbone-mixin`'s `Image` helper to be able to pass in an array of options instead.

I'm open to other implementations, but this one seemed to make sense to me. The downside is that the API can't use this model/collection so there is some kinda funny extension stuff happening on [save](https://github.com/artsy/positron/compare/master...kanaabe:denorm-issue?expand=1#diff-f0960fb95e1353f30292b72ab800dd1eR139). 

The other issue this PR fixes is the case that an artwork doesn't have an artist.

### DIFF
--- a/api/apps/articles/model/save.coffee
+++ b/api/apps/articles/model/save.coffee
@@ -13,7 +13,7 @@ Q = require 'bluebird-q'
 request = require 'superagent'
 requestBluebird = require 'superagent-bluebird-promise'
 Backbone = require 'backbone'
-{ ArtworkHelpers } = require 'artsy-backbone-mixins'
+{ Image } = require 'artsy-backbone-mixins'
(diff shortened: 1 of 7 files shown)


craigspaeth:
:+1: works for me! Nice pragmatic solution.

Issue #697, opened 2016-03-29
Artist link/follow


owendodd:
Highlighting text and clicking the artist icon should pop a search tool (prefilled text should be "Search for artist by name"). Selecting an artist from the dropdown adds a link to the highlighted text (normal styling) and appends a follow button.

<img width="722" alt="screen shot 2016-03-29 at 7 29 16 pm" src="https://cloud.githubusercontent.com/assets/3196536/14127651/8e391914-f5e7-11e5-96cf-0da589268836.png">
<img width="711" alt="screen shot 2016-03-29 at 7 34 46 pm" src="https://cloud.githubusercontent.com/assets/3196536/14127652/8e48f406-f5e7-11e5-9a80-2526ea740e95.png">

kanaabe:
Is this Editorial Team only?

owendodd:
I'd like to launch this ed only just so we can test it, then we will roll out to others.

kanaabe:
Ok!

PR #705, opened 2016-04-01
@craigspaeth Follow artist plugin

kanaabe:
Closes https://github.com/artsy/positron/issues/697

followbutton


craigspaeth:
Nice! Back at it again with the Scribe plugins :100:

PR #712, opened 2016-04-12
@craigspaeth Remove intercom

kanaabe:
I removed the "Report"  button on error pages too -- not sure if there's another place we want those to go, but I don't think anyone was monitoring that anyway? Was Segment only added to support Intercom here? From my understanding of this, it was, and if so, we should probably remove the Segment integration too!


craigspaeth:
Thanks!

Segment wasn't _just_ for Intercom. In the earlier days where it was up-in-the-air how much external use Writer would get, it made sense to provide Intercom + other monitoring/analytics via Segment to provide tracking/tech support. Now-a-days it seems to be more and more an internal tool. If the Publishing team isn't using analytics in Writer, then I say go ahead and remove Segment too.

Issue #719, opened 2016-04-18
When you delete an image in a set, all the captions shift to the right


halleyjohnson:
if you delete an image in the middle of a set, its caption shifts to the image to its right, and the last caption disappears

PR #720, opened 2016-04-19
@craigspaeth Simplify autosave, remove image set button, fix captions, new layout widths

kanaabe:
Closes https://github.com/artsy/positron/issues/719 https://github.com/artsy/positron/issues/718

I'd like to remove autosave from add/remove/reset on sections. Anytime we create a new section, it temporarily saves an empty section to the DB. Then when we click off, it removes it again and saves. The number of writes to an article during editing ends up being a lot more than I expected. Instead, just warn the user to save if a new section is added and they try to X out. Keyup will still trigger an autosave of the entire article.


craigspaeth:
👍  solid, works for me!

Issue #716, opened 2016-04-14
Article markup for Instant Articles


owendodd:
There should be a toggle on the admin panel that dictates whether an article should have an IA version.

![screen shot 2016-04-14 at 3 47 55 pm](https://cloud.githubusercontent.com/assets/3196536/14541506/4914c46e-0258-11e6-8e28-6fb3fb3b14c8.png)

Once the markup is complete I should be able to set up a Style for it in their GUI.

https://developers.facebook.com/docs/instant-articles/reference

PR #721, opened 2016-04-20
@craigspaeth Adds FB :zap: IA flag

kanaabe:
Closes https://github.com/artsy/positron/issues/716

The xapp stuff is to handle tests better, there was this occasional test that randomly couldn't access artsyXapp.token so just set it to '' if it doesn't have it.

### DIFF
--- a/api/apps/articles/model/save.coffee
+++ b/api/apps/articles/model/save.coffee
@@ -23,7 +23,7 @@ cloneDeep = require 'lodash.clonedeep'
 EMBEDLY_KEY, FORCE_URL, ARTSY_EDITORIAL_ID, SECURE_IMAGES_URL } = process.env
 sailthru = require('sailthru-client').createSailthruClient(SAILTHRU_KEY,SAILTHRU_SECRET)
 { crop } = require('embedly-view-helpers')(EMBEDLY_KEY)
-artsyXapp = require('artsy-xapp')
+artsyXapp = require('artsy-xapp').token or ''
 
 @validate = (input, callback) ->
   whitelisted = _.pick input, _.keys schema.inputSchema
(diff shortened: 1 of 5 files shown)


kanaabe:
I wish GH would acknowledge emojis in headers 😿

api/apps/articles/model/save.coffee
```diff
@@ -23,7 +23,7 @@ cloneDeep = require 'lodash.clonedeep'
 EMBEDLY_KEY, FORCE_URL, ARTSY_EDITORIAL_ID, SECURE_IMAGES_URL } = process.env
 sailthru = require('sailthru-client').createSailthruClient(SAILTHRU_KEY,SAILTHRU_SECRET)
 { crop } = require('embedly-view-helpers')(EMBEDLY_KEY)
-artsyXapp = require('artsy-xapp')
+artsyXapp = require('artsy-xapp').token or ''
```
craigspaeth: I think ideally we'd [rewire](https://github.com/jhnns/rewire) `artsyXapp` instead here, but I don't see any big immediate issues with this.

craigspaeth:
Easy peasy! I too wish emojis worked in the headers 😭  Minor comment, feel free to follow up PR about it or not.

api/apps/articles/model/save.coffee
kanaabe: Yeah, do you mean like [this?](https://github.com/artsy/positron/blob/master/api/apps/articles/test/model/save.coffee#L25). I'm probably rewiring it incorrectly, but this works intermittently

craigspaeth: Shoot, yeah 😛  I'll have to investigate more to see why that's failing intermittently ¯\_(ツ)_/¯

Issue #723, opened 2016-04-21
Remove Embedly cropping


dzucconi:
https://github.com/artsy/positron/blob/6c5b38b4d9c8b597c52efc835c95ec0c0785ea55/api/apps/articles/model/save.coffee#L266

`crop` shows up in a few places:

Places where it's used for displaying images within Positron's UI can move over to using Gemini: https://github.com/artsy/force/blob/master/components/resizer/services/gemini.coffee

It looks like we're returning pre-sized URLs through the API for Article#thumbnail_url: ![](http://static.damonzucconi.com/_capture/tY9g1HQ3Z7.png)

I'm not sure the reasoning here but these should be bare URLs that resolve to some un-proxied object on S3.

kanaabe:
thanks @dzucconi, i'm working on this now. the weird thumbnail save is for Sailthru, not on regular thumbnails in the article. They need different size versions of the images they upload -- is there a better way we can handle this?

dzucconi:
Pre-process a version you need in Gemini or resize it on the client, using Gemini's proxy. (Though I don't really know what Sailthru is)

PR #725, opened 2016-04-27
@craigspaeth Remove Embedly

kanaabe:
Closes https://github.com/artsy/positron/issues/723


craigspaeth:
Solid! Thank you for tackling this!!

client/components/resizer/index.coffee
```diff
@@ -0,0 +1,80 @@
...
+    options =
+      resize_to: resize_to
+      src: url
+      width: width if width?
+      height: height if height?
+      quality: quality or defaults.quality
+
+    [endpoint , qs.stringify options].join '?'
+
+  crop: (url, options = {}) =>
```
craigspaeth: I mentioned this on Slack, but at one point—if this logic, that is sorta copypasta-ed between Force and Positron API + Client, is truly valuable enough to warrant a component on top of the Gemini url API—then we should probably extract it into it's own node module. For the time being, I don't have a problem with the copy pasting—especially since we're not 100% off embedly yet.

PR #726, opened 2016-04-28
@craigspaeth Default non-italic captions + bug fixes

kanaabe:
Makes non-italic the default for captions
fixes a bug @halleyjohnson noticed with clicking into an input
pulls the latest `scribe-plugin-enhanced-link-tooltip` which fixes a bug with input link toggle
fixes a bug where italics and links weren't being applied to the caption preview in an image section


craigspaeth:
🐛 🔨  👍

PR #727, opened 2016-05-03
@craigspaeth Curations

kanaabe:
Just waiting on a final SVG but code-wise this is good to go! Sorry for the lengthy PR but this is very similar to Sections.  

Allow Editorial to manage featured links in Writer. Two cases where we already need this is for Homepage links, and rotating email signup images

### Demo
link_set

### API endpoint

```
/api/links?slug=link-six
```

```
/api/links/57291e24fbb972cac51165d8
```

```json
{
   "id": "57291e24fbb972cac51165d8",
   "featured_links": [
   {
      "thumbnail_url": "https://artsy-media-uploads.s3.amazonaws.com/kQ3PBAFFVqv9-2ilcPeqAA%2Ficon2.png",
      "title": "Link 1 test",
      "url": "https://artsy.net/some-link"
   },
   {
      "thumbnail_url": "https://artsy-media-uploads.s3.amazonaws.com/aT17mVXxmhqgbLkXGVgCCw%2Fmedium.jpg",
      "title": "title",
      "url": ""
   }
],
   "title": "Link Set 6",
   "slug": "link-six"
}
```


kanaabe:
cc @owendodd, LMK if you have any thoughts on the design!

craigspaeth:
Code looks great, and I'm excited to see more editorial data moving into Positron!

That said, I'd highly caution against introducing a generic, yet rigid, data model like this. I'd instead encourage creating data models in shapes catered to the use cases. By "data model" I don't mean you have to write new sub-apps, schemas, mongo collections, and REST endpoints for each and every form of curated data—you could for instance store arbitrary blobs (thank you Mongo) of curated data that's modeled specifically to the use case.

For instance imagine a data model called "curations" that requires only two fields `id: @objectid()` and `name: @string()`, then you `allowUnknown: true, stripUnknown: false` [when validating](https://github.com/hapijs/joi/blob/v8.0.5/API.md#validatevalue-schema-options-callback) to accept anything else. Then you could have data models that are much more appropriate for the use case, without having to morph/duplicate/dumb-down data into the "featured link" model.

e.g.

Homepage articles could be stored as a "curation" that looks like this:
````
{
  id: ObjectID(123456),
  name: 'Homepage articles',
  article_ids: [ObjectID(123), ObjectID(123)]
}
````
And when querying /api/curations/123456 the API joins the article_ids into real article data.

Email signup images could be stored like this:
````
{
  id: ObjectID(123456),
  name: 'Email images',
  images: [{ src: 'kitten.jpg', href: 'foo.com' }]
}
````
Which might look a lot like a "featured links" in this case—but the advantage is it's  

I can see a number of advantages to doing it this way including...

* Less duplicating data (e.g. when someone changes the thumbnail of the article, it updates on the homepage as well.
* Relations to real entities and unique `id`s instead of slug  (e.g. we've had to do horrible hacks involving parsing out slugs from featured link urls to do analysis on `ordered_sets`, an array of article_ids 
* Flexible enough to say omit the "title" part of featured links, or expand to add "title" and "subtitle", etc. freeing up the designer to not have to conform the "featured links" model 
* No weird "joining" of featured link sets. e.g. the fairs page is a huge hack of "ordered sets of featured links", "ordered sets of genes", "banner images", etc. If the "fairs page" was one big curated blob of data it would be much saner to piece together. Ditto for auction pages and more.
* We already have a poorly functioning "featured link" model in Gravity—less confusion there
* Also Metaphysics can "join" featured links on the Gravity side to accomplish, more or less, the same thing here
* And more, you can read [here](https://github.com/artsy/force/issues/2811) for some discussion on it with Gravity.

All that said, this implementation does have the advantage of being denormalized and not expecting clients to crawl and join endpoints themselves. I'd be willing to be convinced we should merge this—but given we have the opportunity to think of something better, I'd urge we attempt it.

😅  sorry for the rant. I have an unusual passion against anything called ordered sets or featured links.

**Comment by craigspaeth** [comment]
Also to note from a UI perspective a more flexible "curations" data model could make it easier to build more appropriate "one-off" admin UIs. For instance in the case of homepage articles you could have 7 autocomplete components labeled "top left", "bottom right", etc. to make it obvious which articles go where. Or if a generic editor is more desirable, we've been [using hulk.js](https://github.com/kevinburke/hulk) for things like the about page.

Finally, if there's any time sensitivity behind this then I'm cool with merging it. Just wanted to get my caution out there having been down a similar road that turned out not so great.

**Comment by kanaabe** [comment]
Thanks for this feedback, Craig! I definitely love the idea of having a curation with looser validation instead of a rigid data model. I'm going to revise this and I'll get back to you.

File: api/apps/curations/index.coffee [hunk 1]
```diff
@@ -0,0 +1,11 @@
+express = require 'express'
+routes = require './routes'
+{ setUser, authenticated, adminOnly } = require '../users/routes'
+
+app = module.exports = express()
+
+app.get '/curations', routes.index
+app.get '/curations/:id', routes.find, routes.show
+app.post '/curations', setUser, authenticated, adminOnly, routes.save
```
craigspaeth: 👍  for `adminOnly`

kanaabe:
image

api/apps/curations/model.coffee
```diff
@@ -0,0 +1,80 @@
...
+
+#
+# Retrieval
+#
+@find = (id, callback) ->
+  query = if ObjectId.isValid(id) then { _id: ObjectId(id) }
+  db.curations.findOne query, callback
+
+@where = (input, callback) ->
+  Joi.validate input, querySchema, OPTIONS, (err, input) =>
```
craigspaeth: I wonder how useful querying across curations will be. Just want to note that `OPTIONS` here will allow any query to Mongo's `find` through. Which is fine b/c this is admin only, but maybe preferable to leave off this `validate` call out of the gate.

api/apps/curations/index.coffee
kanaabe: actually this specific route doesn't get used and this can be removed, the UI doesn't allow creation of new curations, just editing existing ones.

client/apps/settings/index.coffee
```diff
@@ -0,0 +1,14 @@
...
+express = require 'express'
+routes = require './routes'
+
+app = module.exports = express()
+app.set 'views', __dirname
+app.set 'view engine', 'jade'
+
+app.get '/settings', routes.index
+app.get '/settings/:id/edit', routes.edit
+app.post '/settings/:id', routes.save
```
kanaabe: this part felt a little wrong -- maybe it should be posting to curations? the route takes care of that but the naming is a little off.

api/apps/curations/model.coffee
```diff
@@ -0,0 +1,80 @@
...
+      callback err, {
+        total: total
+        count: count
+        results: curations.map(@present)
+      }
+
+#
+# Persistence
+#
+@save = (input, callback) ->
```
craigspaeth: Look at how thin this method becomes without a rigid schema! 😛

client/test/models/link_set.coffee
```diff
@@ -0,0 +1,66 @@
...
+      benv.expose $: benv.require 'jquery'
+      sinon.stub Backbone, 'sync'
+      @section = new Section fixtures().articles.sections[0]
+
+  afterEach ->
+    Backbone.sync.restore()
+
+  describe '@getIframeUrl', ->
+
+    it 'converts youtube urls to an iframe frriendly one', ->
```
craigspaeth: _frriendly_

client/test/collections/link_sets.coffee
```diff
@@ -0,0 +1,51 @@
+_ = require 'underscore'
```
craigspaeth: I think these tests can be deleted now?

craigspaeth:
Looks great! I have one comment about the `where` validation, and some unnecessary tests. Otherwise 👍 to merge when 💚

client/test/collections/link_sets.coffee
kanaabe: yeaaaaa :)

client/apps/settings/index.coffee
craigspaeth: Nah, given the UI I think this makes sense. Typically for 1:1 "CRUD" views it's good to keep that "RESTful" naming convention, but not a hard fast rule.

PR #728, opened 2016-05-05
@kanaabe changed date format for report

nicholassewitz:
I don't know if the "T" in the first half of the date format change will work. Currently the format isn't up to ISO standards which is what redshift reads. Trying to get the date field into a usable date format.


kanaabe:
Cool, works for me. Nice to see a new face around here! 👍

Issue #732, opened 2016-05-11
Revisit tag section


owendodd:
- [x] Add contributing others as tags

- [x] Manual tags should be top in heirarchy

(anything else to add? cc @kanaabe @halleyjohnson)

PR #734, opened 2016-05-13
@craigspaeth Revist tagging

kanaabe:
Closes https://github.com/artsy/positron/issues/732

- Adds contributing authors to the end of the keywords array
- Reverse order of tags -- manually added ones should appear in front

### DIFF
--- a/api/apps/articles/test/model/save.coffee
+++ b/api/apps/articles/test/model/save.coffee
@@ -50,6 +50,21 @@ describe 'Save', ->
         @sailthru.apiPost.args[0][1].tags.should.containEql 'magazine'
         done()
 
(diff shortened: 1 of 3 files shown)


api/apps/articles/test/model/save.coffee
```diff
@@ -50,6 +50,21 @@ describe 'Save', ->
         @sailthru.apiPost.args[0][1].tags.should.containEql 'magazine'
         done()
 
+    it 'concats the keywords at the end', (done) ->
```
craigspaeth: ❤️

craigspaeth:
Solid solid solid.

PR #737, opened 2016-05-18
@craigspath has_video query

kanaabe:
To make https://github.com/artsy/force/issues/4947 easier, this addition to the query schema lets someone search for articles that contain video.


craigspaeth:
Solid!

PR #739, opened 2016-05-19
@kanaabe: Configure the Gemini app used for uploading

joeyAghion:
This app uses Gemup's [default of `force`](https://github.com/artsy/gemup/blob/b858caa93ebac601065663c4d03a3287b45c483c/gemup.js#L7) for uploading to Gemini. But really, it should manage its images independently. E.g., staging and production environments shouldn't bundle their uploads together, and if particular image versions need to be generated or callbacks added in the future, Force and Positron should be able to do that separately.

I wasn't able to fully test uploads locally because of a `null` reference in `isEditorialTeam` (but tests succeed). If you want to help me figure out the local settings, I can test more.

I've already created and set `writer` and `writer-staging` Gemini apps and set the ENV vars on Positron apps.


kanaabe:
Looks great, thanks for doing this! I'll review the tests myself later. Merging!

PR #745, opened 2016-06-01
@craigspaeth Handle reordering and deleteing on denormalized artworks

kanaabe:
Soo I thought that looking for the empty artworks is enough but I forgot that reordering/deleting artworks will also mean we need to reorder/delete the denormalized version according to the ids array.


craigspaeth:
Good catch and nice test! Clean and clear!

PR #754, opened 2016-06-06
@craigspaeth Adds channels endpoint

kanaabe:
Just the API side, includes `name` and `user_ids`.

Not quite sure where to go with "tiering" yet, but i imagine two sets of tiering: 1. for the UI features like super articles + image sets, and 2. for publishing settings (can make any associations, can make some associations, can make no associations(partners)  ).


craigspaeth:
👍 👍

PR #757, opened 2016-06-07
@craigspaeth Adds many-to-many associations on fairs and auctions

kanaabe:
Fairs + Auctions are the only two entities that have singular associations. I'll delete the singular versions after backfilling + getting client apps to use plural forms


craigspaeth:
Solid code and migration plan!

PR #758, opened 2016-06-08
@craigspaeth Association cleanup

kanaabe:
Backfilled the fair_ids so we're good to switch over. Last step is deleting those fields from the DB


craigspaeth:
w00t w00t!

PR #759, opened 2016-06-09
@craigspaeth Channel mgmt UI

kanaabe:
This just gets a UI going for channels in the /settings tab and restructures settings a bit to use nested routing. Adds an autocomplete list with users.

cc @owendodd 

channelmgmt


craigspaeth:
Nice! Looks good to me :)

PR #760, opened 2016-06-10
@craigspaeth Adds channel_id to articles

kanaabe:
Instead of having both `channel_ids` and `partner_channel_ids`, I'm thinking it can be consolidated to one `channel_id` since we've established that articles can only be in one channel, whether it be a partner or our own. 

The desired workflow is that we'd copy the article if we want in two channels (benefits analytics, tweak content for channel, extended versions, etc).

The one weird thing is mixing id types, where `channel_id` can be a Channel or Partner. Is that too confusing? I don't mind separating it to `channel_id` and `partner_channel_id` either.

### DIFF
--- a/api/apps/articles/model/retrieve.coffee
+++ b/api/apps/articles/model/retrieve.coffee
@@ -11,7 +11,7 @@ moment = require 'moment'
     # Separate "find" query from sort/offest/limit
     { limit, offset, sort } = input
     query = _.omit input, 'limit', 'offset', 'sort', 'artist_id', 'artwork_id', 'super_article_for',
-      'fair_ids', 'fair_programming_id', 'fair_artsy_id', 'fair_about_id', 'partner_id', 'auction_id', 'show_id', 'q', 'all_by_author', 'section_id', 'tags', 'has_video', 'fair_id'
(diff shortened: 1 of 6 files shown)


craigspaeth:
Adding `channel_id` is necessary regardless if you decide to add `partner_channel_id` later, so I'll merge.

My initial reaction is that it's better to keep these concepts separate and not try to reuse the same field for it, e.g. Positron "channels" are kinda their own entity that has their own access control concept and Gravity "partners" and those access control rules !== channels and maybe should stay living in Gravity (even though they might both be referred to as channels in irl). One way I could see this getting sticky is by considering them the same thing one will have to find a way to keep Gravity partners and Positron channels in sync (e.g. when Partner's are renamed, or users accts are added to partners).

On the other hand if you don't add complexity in the syncing Gravity partner <> Positron channels, that complexity will get moved elsewhere in the form of querying Gravity for it's access control rules (e.g. hit api.artsy.net/api/v1/me/partners and join it with writer.artsy.net/api/channels?has_access=me to determine what Writer channels you can see).

So I'd be happy with either approach, just laying the pros/cons that I can foresee.

Issue #752, opened 2016-06-06
Articles should go to Sailthru if they're scheduled


halleyjohnson:
Articles should load in Sailthru if they're scheduled or published

PR #762, opened 2016-06-13
@craigspaeth Send Scheduled Articles to Sailthru

kanaabe:
Allows for better prep time for ST campaigns, especially with the upcoming Digital Daily. Moves the `setEmailFields` method to the `sanitizeAndSave` callback and lets scheduled articles get sent to ST on save. 

Closes https://github.com/artsy/positron/issues/752


craigspaeth:
👍 good work on code + tests!

PR #765, opened 2016-06-16
@craigspaeth Channel query, biz logic, save partner + channel on user

kanaabe:
This is the final PR before actually switching over the Writer UI to using Channels-- trying to do as much backend work as possible. Will make notes below! 

- Saves Partner and Channel ids to the user
- Adds some biz logic for channel types
- Query for channels by `user_id`

### DIFF
--- a/api/apps/users/model.coffee
+++ b/api/apps/users/model.coffee
@@ -12,7 +12,7 @@ debug = require('debug') 'api'
 async = require 'async'
 bcrypt = require 'bcrypt'
 { ObjectId } = require 'mongojs'
-{ ARTSY_URL, SALT } = process.env
+{ ARTSY_URL, SALT, API_URL } = process.env
 
 #
 # Retrieval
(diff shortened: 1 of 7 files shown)


api/apps/users/model.coffee
```diff
@@ -45,38 +45,35 @@ bcrypt = require 'bcrypt'
...
+      db.channels.find {user_ids: ObjectId(user.id)}, cb
+    (cb) ->
+      bcrypt.hash accessToken, SALT, cb
+  ], (err, results) ->
     return callback err if err
-    encryptedAccessToken = results[0]
+    user.partner_ids = _.map results[0].body, (partner) ->
+      partner.property._id
+    user.channel_ids = _.pluck results[1], '_id'
+    encryptedAccessToken = results[2]
```
kanaabe: The diff is a little hard to read here, but I'm adding the channel + partner fetches to the `save`.

The basic motivation is that eventually I'm going to get rid of `findOrInsert` which is used primarily for impersonating users (`writer.artsy.net/api/v1/user/:id`). We also use it on article saving, but I don't think we need that -- should just be enough to use `fromAccessToken` since you'll always publish as `me` (`api.artsy.net/api/v1/me`). 

Todo: ensure that a fresh login will save the latest user data

client/models/channel.coffee
```diff
@@ -4,3 +4,26 @@ sd = require('sharify').data
...
+      when 'support' then {
+        features: []
+        sections: ['text', 'artworks', 'images', 'video', 'callout']
+        associations: ['artworks', 'artists', 'shows', 'fairs', 'partners','auctions']
+      }
+      when 'partner' then {
+        features: []
+        sections: ['text','artworks', 'images','video']
+        associations: []
+      }
```
kanaabe: Maybe this is a bad idea, but I'm thinking the `Channel` model in the client can be both channels and partners in order to keep things clean in templates + client-side code... All it needs is the `id` and `type` which I can set at the session level.  thoughts?

craigspaeth: I think this is a great idea! Keeping them separate is useful at the data-level to avoid redundant data and "syncing" from services—but at the client-level I think it makes sense to join these for the UI to be easier to reason about.

craigspaeth:
Solid work! Looking forward to seeing all these backend PRs build up to the frontend!!

PR #769, opened 2016-06-21
@craigspaeth Fixes forbidden error

kanaabe:
New users are tripping on the save because `user/id:/access_controls` is only available for Admins. Use the `/me/partners` endpoint instead. 

Test will fail until this gets merged: https://github.com/artsy/antigravity/pull/69


kanaabe:
hey @craigspaeth i'm going to merge this for now since new users can't access Writers, but please let me know if you think this should be changed!

craigspaeth:
Nope looks good, thanks for taking the initiative to merge and push the fix!

PR #781, opened 2016-07-20
@xtina-starr Gives associations to partner channels

kanaabe:
The `hasAssociation` method gets used in templates to determine if you should be able to edit associations depending on what Channel its in. Since the Admin panel is only available to Artsy Admins and based [this convo](https://artsy.slack.com/archives/publishing-support/p1469040288000092), we should allow Admins to edit all associations.

### DIFF
--- a/client/models/channel.coffee
+++ b/client/models/channel.coffee
@@ -77,7 +77,14 @@ module.exports = class Channel extends Backbone.Model
         'auctions'
       ], association
     else if type is 'partner'
-       false
+      _.contains [
+        'artworks'
+        'artists'
+        'shows'
+        'fairs'
+        'partners'
+        'auctions'
+      ], association
 
   fetchChannelOrPartner: (options) ->
     async.parallel [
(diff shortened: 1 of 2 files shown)


client/models/channel.coffee
```diff
@@ -77,7 +77,14 @@ module.exports = class Channel extends Backbone.Model
         'auctions'
       ], association
     else if type is 'partner'
-       false
+      _.contains [
```
xtina-starr: super small, but it looks like the elements in the array passed to `_.contains` are the same elements. maybe defining an associations array after line 58 could remove some duplication. nothing to hold up the PR tho. something to consider on the next one...

xtina-starr:
🌟

client/models/channel.coffee
kanaabe: yeah good point --  i have a feeling these associations will be tweaked often so i'll leave it in for now but definitely true about DRYing up code

xtina-starr: ah that makes sense!

PR #800, opened 2016-08-31
@craigspaeth => Adds ability to query for multiple articles

kanaabe:
What do you think? It'd help a lot for getting stuff like sub-super articles + pinned articles.


craigspaeth:
💯  I'm all for this!

Issue #805, opened 2016-09-05
Yoast focus keyword in first paragraph


kanaabe:
@mdole I think this metric isn't working? When I include a keyword that exists in the first paragraph, it still says "keyword not found in first paragraph".

PR #806, opened 2016-09-06
@kanaabe Fixes issue with Yoast and keywords in the first paragraph (#805)

mdole:
The issue was that Yoast was only checking the Lead Paragraph of the article, not the first text that appeared (if the author chose not to write a Lead Paragraph). This fix causes Yoast to first check if there's a LP and then check the keyword against the first paragraph of the text if there isn't.

### DIFF
--- a/client/apps/edit/components/layout/index.coffee
+++ b/client/apps/edit/components/layout/index.coffee
@@ -158,14 +158,26 @@ module.exports = class EditLayout extends Backbone.View
     imageCount = 0
     textIndex = 0
     @fullText = []
-    if $(@article.get('lead_paragraph')).text() isnt ''
+    if $(@article.get('lead_paragraph')).text()
       @fullText.push(' <p> ' + $(@article.get('lead_paragraph')).text() + ' </p> ')
       textIndex += 1
 
     for section in $(@article.get('sections'))
       if section.type is 'text' && textIndex is 0
-        @fullText.push(' <p> ' + $(section.body).find('p').andSelf().filter('p:first').text() + ' </p> ')
-        @fullText.push($(section.body).find('p').andSelf().filter('p:not(:first)').text())
+        @fullText.push(
+          ' <p> ' + 
+          $(section.body)
+          .find('p')
+          .andSelf()
+          .filter('p:first')
+          .text()
+           + ' </p> ')
+        @fullText.push(
+          $(section.body)
+          .find('p')
+          .andSelf()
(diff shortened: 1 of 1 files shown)


client/apps/edit/components/layout/index.coffee
```diff
@@ -156,18 +156,26 @@ module.exports = class EditLayout extends Backbone.View
 
   checkSeo: =>
     imageCount = 0
+    textIndex = 0
     @fullText = []
-    @fullText.push(" <p> " + $(@article.get("lead_paragraph")).text() + " </p> ")
-    for section in $(@article.get("sections"))
-      if section.type is "text"
+    if ($(@article.get('lead_paragraph')).text() != '')
```
kanaabe: you can use the `not` keyword here, this is coffeescript-land! 🏰

kanaabe: actually what's even better is `if @article.get('lead_paragraph').text()` because an empty text will be a falsey value

client/apps/edit/components/layout/index.coffee
```diff
@@ -156,18 +156,26 @@ module.exports = class EditLayout extends Backbone.View
...
+      @fullText.push(' <p> ' + $(@article.get('lead_paragraph')).text() + ' </p> ')
+      textIndex += 1
+
+    for section in $(@article.get('sections'))
+      if section.type is 'text' && textIndex == 0
```
kanaabe: same, we can use coffescript here `if section.type is 'text' and textIndex is 0`

client/apps/edit/components/layout/index.coffee
```diff
@@ -156,18 +156,26 @@ module.exports = class EditLayout extends Backbone.View
...
+        @fullText.push(' <p> ' + $(section.body).find('p').andSelf().filter('p:first').text() + ' </p> ')
+        @fullText.push($(section.body).find('p').andSelf().filter('p:not(:first)').text())
```
kanaabe: wish there was an easier way to do this! if anything, we should break out these lines so we can read them better.
```
@fullText.push(
   $(section.body)
      .find('p')
      .andSelf()
      .filter('p:not(:first)')
      .text()
)
```
syntax above might be slightly off so test to make sure it's works!

mdole:
@kanaabe thank you for the comments! the most recent commit should address them; let me know if you have any other thoughts

kanaabe:
Woo! LGTM. Awesome work!

PR #807, opened 2016-09-07
@kanaabe => Yoast snippet properly reflects the title Artsy uses for Google

mdole:
Had initially set the title for yoast's snippet editor (which imitates the appearance of the page on a google SERP) to be the article title; Artsy uses the thumbnail title instead.

### DIFF
--- a/client/apps/edit/components/layout/index.coffee
+++ b/client/apps/edit/components/layout/index.coffee
@@ -190,7 +190,7 @@ module.exports = class EditLayout extends Backbone.View
     
     yoastView = new YoastView
       contentField: @fullText
-      title: @article.get('title')
+      title: @article.get('thumbnail_title')
       slug: @article.getSlug()
 
   onKeyup: =>
--- a/client/apps/edit/components/layout/index.coffee
+++ b/client/apps/edit/components/layout/index.coffee
@@ -188,9 +188,15 @@ module.exports = class EditLayout extends Backbone.View
     @fullText.push('<img></img>') for num in [imageCount..1]
     @fullText = @fullText.join(' ')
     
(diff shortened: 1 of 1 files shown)


mdole:
cc @kanaabe

client/apps/edit/components/layout/index.coffee
```diff
@@ -188,9 +188,15 @@ module.exports = class EditLayout extends Backbone.View
...
+    @yoastTitle = ''
+    if @article.get('thumbnail_title')
+      @yoastTitle = @article.get('thumbnail_title')
+    else
+      @yoastTitle = @article.get('title')
+
     yoastView = new YoastView
       contentField: @fullText
-      title: @article.get('title')
+      title: @yoastTitle
```
kanaabe: ooh we can make this a one-line PR, ready? 
```
title: @article.get('thumbnail_title') or @article.get('title') or ''
```
and then if possible, we should add a test! you should be able to see what gets passed into YoastView by doing something after [this line](https://github.com/artsy/positron/blob/master/client/apps/edit/components/layout/test/index.coffee#L152)
```
console.log @EditLayout.YoastView.args
```

mdole:
@kanaabe the one-liner is so much more elegant, thanks. I wasn't able to figure out a test that worked--I think we had some problems with this when we were designing the original tests, as we ended up checking fullText instead of the yoast args. Thoughts?

kanaabe:
Ah yeah, you're right. Merging for now and I'll try to iron out the YoastView args for tests sometime later. Probably will become important to test YoastView args eventually as this module get more feature-packed and aWeSoMe

PR #812, opened 2016-09-26
@kanaabe + @eessex : *Don't merge* Quick GraphQL expirement

craigspaeth:
**DO NOT MERGE** Just a proof of concept.

This is a quick branch to experiment with what it would take to get an MVP GraphQL endpoint on Positron. Turns out it's not that much work at all to tack on my [joiql-mongo](https://github.com/craigspaeth/joiql-mongo) project for CRUDL operations based on our current Joi schemas. That said, there's quite a bit more to go to integrate the rest of the persistence "middleware" currently in Positron, as well as cleaning up the Joi schema to map validations better to GraphQL vs. REST.

Quickly discovered there's a couple of larger items to make this happen...

1. More [robust support](https://github.com/craigspaeth/joiql/issues/1) for Joi types in JoiQL
2. Refactor the Joi schema to add better conditional validation e.g. we broke up our schemas into "input" and "query" schemas which are basically WRITE and READ schemas—we'd want to map that better to GraphQL query/mutation validation. For instance, in this implementation `author_id` is required to _find_ an article when it should only be required when `create` an article.
3. Integrate Article#save logic into `.on('create', ...)` [JoiQL middleware](https://github.com/craigspaeth/joiql-mongo#modelonmethod-middlewarefunction)

All said and done, it looks like there's not a ton of work here.

cap


kanaabe:
😱 ❤️

PR #815, opened 2016-10-11
@kanaabe : index articles in elasticsearch

cavvia:
- Adds elasticsearch indexing to articles so we can display them in the artsy.net auto-suggest and subsequently in our planned (elasticsearch backed) search results page.
- Articles are sent to Elasticsearch in an asynchronous fashion on save and deleted on destroy. 
- Adds a reindex script for performing full reindexes manually.
- Adds a new dev dependency in the form of Elasticsearch, which has to be running in order for the test suite to pass. I've added a note to the `README` and it should be very simple to run in the background, let me know if you have or foresee any issues with that. 

**Migration**
- I've added a new `ELASTICSEARCH_URL` config var to Heroku.
- I've created an empty index called `articles_production` on our production Elasticsearch cluster.
- Add Elasticsearch dependency to Semaphore CI.
- Reindexing depends on an upcoming Gravity PR (which will add our article mappings to elasticsearch) so I won't run it immediately, but this is the reindexing script:

```
NODE_ENV=${ENV} MONGOHQ_URL=${URL} coffee scripts/reindex_articles.coffee > reindex.log
```


api/apps/articles/model/save.coffee
```diff
@@ -290,6 +330,12 @@ typecastIds = (article) ->
     debug err if err
     cb()
 
+stripHtmlTags = (str) ->
```
cavvia: i imagine there is an implementation of this tucked into one of our dependencies but i couldn't find it.

api/apps/articles/test/model/index.coffee
```diff
@@ -682,6 +689,22 @@ describe 'Article', ->
...
         done()
 
+    it 'indexes the article in elasticsearch on save', (done) ->
+      Article.save {
+        author_id: '5086df098523e60002000018'
+        title: 'foo article'
+        published: true
+      }, 'foo', (err, article) ->
+        return done err if err
+        sleep.sleep 1
```
cavvia: i don't know a way around this given that the elasticsearch indexing refresh rate can't go lower than 1 second and receiving an http response is no guarantee that the indexing operation has completed. we have to do this in gravity too for end-to-end integration tests involving elasticsearch.

craigspaeth: 😮  interesting! Makes sense for the sake of integration testing. However, maybe we could instead stub `search.client` and avoid the need to request to ES in tests? Happy to save that for another PR, but quickly to note I think what you want is to use `setTimeout` instead here. That sleep module seems like a kinda strange debugging module that purposely blocks, whereas Node is designed to intentionally use non-blocking callback based APIs.

So this would look like...

````coffeescript
Article.save {
 ...
}, 'foo', (err, article) ->
  return done err if err
  setTimeout(=>
    search.client.search(...)
  , 1000)
````

File: api/lib/elasticsearch.coffee [hunk 3]
```diff
@@ -0,0 +1,8 @@
+elasticsearch = require('elasticsearch')
+{ NODE_ENV, ELASTICSEARCH_URL } = process.env
+
+client = new elasticsearch.Client(host: ELASTICSEARCH_URL)
+
+module.exports =
+    index: 'articles_' + NODE_ENV
```
craigspaeth: Minor style comment: looks like there's an extra 2 spaces in this indentation.

scripts/reindex_articles.coffee
```diff
@@ -0,0 +1,22 @@
...
+env = require 'node-env-file'
+switch process.env.NODE_ENV
+  when 'test' then env path.resolve __dirname, '../.env.test'
+  when 'production', 'staging' then ''
+  else env path.resolve __dirname, '../.env'
+
+# Connect to database
+db = mongojs(process.env.MONGOHQ_URL, ['articles'])
+
+db.articles.find({ }).toArray (err, articles) ->
```
craigspaeth: Looks like the main use case is to sync to ES upon save/delete, so probably not a blocker—but doing this will load all of the articles in memory and iterate through them. To work around this it's often [streams](https://github.com/mafintosh/mongojs#streaming-cursors) are helpful, but more likely this would involve some kind of `limit` and iterating in batches.

craigspaeth:
Kana's OOO right now, so I'll take this.

Looking good! I left some comments about Node best practices to be aware of. The failing build looks like it's related to the addition of elastic search as a dependency, so it'll probably be more convenient to make it an optional dependency of this project by stubbing it in the tests and putting in `return cb() unless ELASTICSEARCH_URL` guards in the additional model functions.

After you take a stab at a revision of this PR I'd be happy to pair through getting the build 💚  or simply wrapping it up for you if you need to move on to bigger and better things.

Thank you for this work!! Excited to make this step in replacing site search with ES!

api/apps/articles/model/save.coffee
```diff
@@ -150,6 +151,43 @@ getDescription = (article) =>
...
+      scheduled_publish_at: article.scheduled_publish_at
+      author: article.author and article.author.name or ''
+      featured: article.featured
+      tags: article.tags
+      body: sections and stripHtmlTags(sections.join(' ')) or ''
+    , (error, response) ->
+      console.log(error) if error
+  )
+
+  cb(article)
```
cavvia: That's right I'd like the indexing to happen in the background so I'll remove the callback.

api/apps/articles/test/model/index.coffee
cavvia: Sure, have made this switch to `setTimeout`.

api/apps/articles/model/index.coffee
```diff
@@ -80,7 +80,8 @@ Q = require 'bluebird-q'
 # Destroy
 #
 @destroy = (id, callback) ->
-  db.articles.remove { _id: ObjectId(id) }, callback
+  db.articles.remove { _id: ObjectId(id) }, =>
+    removeFromSearch id, callback
```
cavvia: i've removed the callback here as i'm comfortable just logging errors for this.

cavvia:
Thanks for the in-depth review! The build failure relates to Elasticsearch js library internals rather than Elasticsearch itself, looking into it. Semaphore comes with elasticsearch installed by default, I just added the following line to the setup to run the server in the background:

```
sudo /etc/init.d/elasticsearch start
```

I'd rather preserve end-to-end integration tests if possible as they are more rigorous than mocks - I've switched the `sleep` statements to `setTimeout` declarations.

cavvia:
The build was fixed by updating Semaphore to use the correct version of node (5.1.1) - it was previously running on 5.0.0 which was not compatible with the `elasticsearch` library i'm using. We stipulate 5.1.1 in our `package.json` so this should be fine in production.

scripts/reindex_articles.coffee
cavvia: That's right this is more of a one-off job - we will run it each time we update the mappings in the elasticsearch index but that will be an infrequent occurrence. I have been able to run this fine for all our articles, though it is a slowish operation which I guess could block the db for a short period (2-5 mins).

craigspaeth:
Thanks for addressing the comments and getting the build 💚  👍

PR #816, opened 2016-10-11
@craigspaeth => Fill width artworks

eessex:
Closes https://github.com/artsy/positron/issues/795
Removed manual padding from fillwidth artwork sections.


craigspaeth:
Nice! Always appreciate a solution that results in less code.

PR #821, opened 2016-10-20
@craigspaeth => GraphQL read-only endpoint

kanaabe:
Do not merge yet...there are some things I'd like to discuss! 

- Updates to Node 6
- Decorates meta properties on schema
- Adds GraphQL endpoint (api/graphql)

**Needs tests


api/apps/graphql/index.coffee
```diff
@@ -0,0 +1,34 @@
...
+
+api = joiql
+  query:
+    articles: array().items(object(
+      Article.inputSchema
+    )).meta(
+      args: Article.querySchema
+    )
+
+api.use (ctx, next) ->
```
craigspaeth: A suggestion for testing these parts—it would be good to extract these functions into testable libraries. A "resolvers" layer is a popular abstraction in GraphQL, so you could start organizing things that way e.g. this woudl be `api.use resolvers.articles` and this middleware function would be moved out to resolvers/articles.coffee, and exported e.g.

````coffeescript
module.exports.articles = (ctx, next) ->
````

You could then write require them into tests and writing tests similar to how we test routing middleware functions—by stubbing the `ctx` and `next` objects/functions passed in.

**Comment by craigspaeth** [comment]
Looking good! Happy to answer any inline questions.

File: api/apps/graphql/index.coffee [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+      next()
+      resolve()
+
+app.use '/graphql', graphqlHTTP(
+  schema: api.schema
+  graphiql: true
+  formatError: (error) => ({
+    message: error.message,
+    locations: error.locations,
+    stack: error.stack
```
kanaabe: Do we want to omit the stack trace for production? Like, this should only apply for `['development', 'test'].includes NODE_ENV`, correct?

craigspaeth: I actually this it's fine to leave these verbose errors/stacks in production. It'll be easier to debug any issues in production and our code is open source anyways.

api/apps/articles/test/integration.coffee
```diff
@@ -63,6 +64,7 @@ describe 'articles endpoints', ->
         .send(title: 'Hi', partner_channel_id: ObjectId('5086df098523e60002000012')).end (err, res) ->
           res.body.title.should.equal 'Hi'
           done()
+      return
```
kanaabe: [Mocha 3.0](https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#300--2016-07-31) no longer lets you return a promise and inject a callback in a single test

api/test/helpers/db.coffee
```diff
@@ -22,5 +22,7 @@ fixturize = (collection, data) ->
   data
 
 @empty = (callback) =>
-  cb = _.after @db.collections.length, callback
-  @db[col].drop(cb) for col in @db.collections
+  @db.getCollectionNames (err, names) =>
+    return callback() if names.length is 0
+    cb = _.after names.length, callback
+    @db.collection(col).drop(cb) for col in names
```
kanaabe: [Mongojs 2.4](https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#300--2016-07-31) only exposes `_getConnection` and `getCollection` -- no longer an array of collections we can iterate over. We can get the collections via the `getCollectionNames` method instead!

kanaabe:
Quick branch created as a POC rendering articles from this PR running locally! :  https://github.com/artsy/microgravity/compare/master...kanaabe:graphql-articles?expand=1

craigspaeth:
👏  well done on the refactors. I'll have to take note as I look into upgrading the other apps. Excited to see this in action!!

PR #824, opened 2016-10-26
@craigspaeth => Small ES fix and cursor readability

kanaabe:
The ES thing is the only thing I saw 500 errors from, but it's very minor.


kanaabe:
Gonna go ahead and merge this guy and test!

PR #830, opened 2016-11-04
@kanaabe : fix elasticsearch reindex script

cavvia:
The elasticsearch reindex script was overloading our elasticsearch servers with parallel indexing requests, which was eventually leading to timeouts and connection failures. 

I've adjusted the config to provide a `maxSockets` which the client can bind to in parallel and a 50ms interval between requests. This fixes the issue and allows us to index all our articles in about 15 minutes.


kanaabe:
Excellent, thanks! If there's anything I can help with on the Force/MG side for getting the articles in search, please let me know/let's discuss!

PR #831, opened 2016-11-04
@cavvia => Editorial only search

kanaabe:
Will update config vars on Heroku, but this should do it for searchable items being Editorial-only.


cavvia:
Looks good! Let me know when it's in production and I'll run the migration, which basically just needs production env:

```
NODE_ENV=production MONGOHQ_URL=$X ELASTICSEARCH_URL=$Y coffee scripts/reindex_articles.coffee
```

PR #833, opened 2016-11-07
@kanaabe : fix visible_to_public criteria

cavvia:
- Fixes a bug in `visible_to_public` flag indexed in elasticsearch, which was affecting visibility of results. The `channel_id` equality check was failing on an `ObjectId` type issue when models were loaded from the db.
- Adds thumbnail image urls to elasticsearch.


cavvia:
Bumping this!

kanaabe:
Whoops thank you!

Issue #834, opened 2016-11-08
Slight Article Tab Restructure


owendodd:
Content and Display are linked while Admin is treated as a separate button. 

![screen shot 2016-11-08 at 4 52 21 pm](https://cloud.githubusercontent.com/assets/3196536/20118767/d5fb0536-a5d3-11e6-93b5-5e4ae84f44b5.png)

PR #839, opened 2016-11-09
@eessex => Header styling update

kanaabe:
Closes https://github.com/artsy/positron/issues/834

image


eessex:
looks good, merging

PR #840, opened 2016-11-10
@craigspaeth => EOY Curation

eessex:
"Curation" for the EOY feature article
Adds user interface for editing curations with type 'editorial-feature'

curation-form

### DIFF
--- a/client/apps/settings/client/curations.coffee
+++ b/client/apps/settings/client/curations.coffee
@@ -8,7 +8,7 @@ _ = require('underscore')
 module.exports.CurationEditView = class CurationEditView extends Backbone.View
 
   events:
-    'click .edit-feature__section-title': 'revealSection'
+    'click .settings-edit-feature__section-title': 'revealSection'
 
   initialize: ->
     @curation = new Curation sd.CURATION
@@ -16,12 +16,11 @@ module.exports.CurationEditView = class CurationEditView extends Backbone.View
       model: @curation
       el: $('body')
       onDeleteUrl: '/settings/curations'
-    if this.curation.get('type') == "editorial-feature"
-      @initMenuState()
+    @initMenuState() if @curation.get('type') is 'editorial-feature'
 
   initMenuState: =>
     $('.page-header').addClass 'sticky'
-    $('.edit-feature').fadeIn()
+    $('.settings-edit-feature').fadeIn()
 
   revealSection: (e) ->
     $(e.target).next().toggleClass('active').slideToggle()
--- a/client/apps/settings/client/curations.coffee
+++ b/client/apps/settings/client/curations.coffee
@@ -2,9 +2,22 @@ AdminEditView = require '../../../components/admin_form/index.coffee'
 Curation = require '../../../models/curation.coffee'
 sd = require('sharify').data
 _s = require 'underscore.string'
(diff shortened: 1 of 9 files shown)


client/components/admin_form/index.coffee
```diff
@@ -30,3 +30,7 @@ module.exports = class AdminEditView extends Backbone.View
 
   ignoreUnsaved: ->
     window.onbeforeunload = ->
+
+module.exports.init = ->
+  new AdminEditView
+    el: $('body')
```
kanaabe: Just curious what the reasoning behind wrapping this in an init was? I would check to make sure other pages that use `AdminEditView` aren't affected! ie, articles/:id/edit.

eessex: @kanaabe - I was experiencing the same problem of an empty el on the AdminEditView -- seems to be working on the other curation, which is the only place I saw the class used.

eessex: Ah - now i see it called in the articles/edit -- double checking this now

client/apps/settings/client/curations.coffee
```diff
@@ -1,10 +1,31 @@
...
+  events:
+    'click .edit-feature__section-title': 'revealSection'
+
+  initialize: ->
+    @curation = new Curation sd.CURATION
+    new AdminEditView
+      model: @curation
+      el: $('body')
+      onDeleteUrl: '/settings/curations'
+    if this.curation.get('type') == "editorial-feature"
```
craigspaeth: Minor coffeescript code ⛳️  tip: `@initMenuState() if @curation.get('type') is 'editorial-feature'`

client/apps/settings/stylesheets/index.styl
```diff
@@ -28,3 +28,49 @@
...
+
+.page-header.sticky
+  position fixed
+  top 0
+  left 110px
+  right 0
+  background-color white
+  z-index 100
+
+.edit-feature
```
craigspaeth: Minor but ideally this is namespaced by the containing app, e.g. `.settings-edit-feature`.

client/components/admin_form/index.coffee
eessex: @kanaabe @craigspaeth  - I just updated the routes so that edit will return back to the edit form.  Previously it was rendering the response to the curations_edit route, but the route is broken and doesn't give the correct URL.  I was having trouble figuring out how to change the route, but replaced it with string interpolation and it goes to the correct place.

client/apps/settings/stylesheets/index.styl
```diff
@@ -28,3 +28,49 @@
     color gray-darker-color
   &:last-of-type
     border-bottom none
+
+.page-header.sticky
```
craigspaeth: This should probably live [here](https://github.com/artsy/positron/blob/74c5611d004687347efa0109e1582b9db94d0fb5/client/components/layout/stylesheets/index.styl#L28) to co-locate the styling with the original `.page-header` definition. It's minor, but for example—leaving in local to the app could make it easy to for someone to forget to port over this generic style when faced with rebuilding and app (something not often in Positron, but we do a lot in Force).

craigspaeth:
👍  nice! The UI looks slick! I have some minor comments, but nothing blocking. Feel free to address and merge yourself when ready.

PR #844, opened 2016-11-16
@kanaabe => EOY secondary image field

eessex:
Adds support for a second image and caption to "editorial-feature" edit form


kanaabe:
Cool, looks good. One thing I forgot to mention in your last PR is tests! We should definitely be adding some now that we have some more things happening in the client-side. Please follow up with some tests for `curations` when you get a chance!

PR #845, opened 2016-11-17
@kanaabe => Edit Editorial Feature mods

eessex:
Added field-name distinctions for second image and caption
Added tests covering editing curations


kanaabe:
👏  thanks for writing those tests!

Issue #837, opened 2016-11-08
Queue Panel


owendodd:
Adds the ability to schedule the daily and weekly editorial emails in Writer. The post queue is a nice-to-have at the this point. Let's add if it doesn't expand the scope by much. Otherwise let's just do the daily and weekly components. 

![screen shot 2016-11-08 at 5 14 06 pm](https://cloud.githubusercontent.com/assets/3196536/20119557/f356bee2-a5d6-11e6-83fa-a03fc456d7f2.png)

Sketch file here: https://cl.ly/220X1m3A2X1M

PR #847, opened 2016-11-22
@eessex => Queue panel

kanaabe:
cc: @craigspaeth 
Closes https://github.com/artsy/positron/issues/837

### Mock
queue

### New dependency
[`Lokka`](https://github.com/kadirahq/lokka) is a simple GraphQL client that's easy to use and supports caching (not making use of this part).

### Main functionality
Selecting items from a filterable search saves the article with a `daily_email` or `weekly_email` flag and sends the article data to Sailthru.

### How the React stuff works
I got a little React-happy with this so let me know if anything is unclear. There is a parent React component, called `QueueView` which has child components called `QueueArticles` and `FilterSearch` which are both basically a list of articles (the latter has filter capability). The `QueueView` keeps the two lists of articles in its `state`, which makes it easy to move a selected article from one list to the other. 

### Splitting out the components
The reason I moved `FilterSearch` to the components folder is so we can use it on the Published and Drafts pages later on.

I think we can go even further with this and make `QueueArticles` a generic thing with a list of articles. I'll think about breaking that out. 

Will comment on the other stuff below!


client/apps/queue/routes.coffee
```diff
@@ -0,0 +1,28 @@
...
+  latestQuery = query "channel_id: \"#{channel_id}\", published: true, sort: \"-published_at\", daily_email: false"
+
+  client = new Lokka
+    transport: new Transport(API_URL + '/graphql')
+
+  client.query(queuedQuery).then (result) =>
+    if result.error
+      return res.backboneError
+    res.locals.sd.QUEUED_ARTICLES = result.articles
+    client.query(latestQuery).then (result) =>
```
kanaabe: Nesting the queries like this is really not preferred and soo counterintuitive with GraphQL, but I couldn't get around using aliases with JoiQL simply. See: https://github.com/muraljs/joiql/issues/11

api/apps/articles/model/index.coffee
```diff
@@ -104,3 +104,30 @@ Q = require 'bluebird-q'
...
+    auction_ids: article.auction_ids.map(String) if article.auction_ids
+    partner_ids: article.partner_ids.map(String) if article.partner_ids
+    show_ids: article.show_ids.map(String) if article.show_ids
+    primary_featured_artist_ids: article.primary_featured_artist_ids.map(String) if article.primary_featured_artist_ids
+    featured_artist_ids: article.featured_artist_ids.map(String) if article.featured_artist_ids
+    featured_artwork_ids: article.featured_artwork_ids.map(String) if article.featured_artwork_ids
+    biography_for_artist_id: article.biography_for_artist_id.toString() if article.biography_for_artist_id
+    super_article: if article.super_article?.related_articles then _.extend article.super_article, related_articles: article.super_article.related_articles.map(String) else {}
+    channel_id: article.channel_id.toString() if article.channel_id
+    partner_channel_id: article.partner_channel_id.toString() if article.partner_channel_id
```
kanaabe: @craigspaeth/@eessex: This is kind of the workaround for getting article updates working properly. 

On a regular article save, we serialize all the article on the client and send that over, but for an update to a single field, like this, I really didn't want to fetch the entire article twice. Instead, we pass the id and have our middleware `find` the article. Only problem with that part is, Joi validations can only compare primitive stuff, and not `ObjectId`.

If you have any ideas, please let me know!

kanaabe:
Gonna go ahead and merge this and address the component refactor later!

api/apps/articles/model/index.coffee
craigspaeth: Joi finally has the ability to create custom extensions, and with that comes the ability to implement your own typecasting. I did this in JoiQL Mongo [here](https://github.com/muraljs/joiql-mongo/blob/master/index.js#L26) to convert string input to objectids after running Joi validation. It looks like you might be going the other way around here however—accepting ObjectIDs and typecasting them to strings. Either way, I'm sure there's something more elegant to be explored in [Joi's extensions](https://github.com/hapijs/joi/blob/v10.0.1/API.md#extendextension). Happy to pair on this too.

client/apps/queue/routes.coffee
craigspaeth: Still need to look at that 😅

client/apps/queue/routes.coffee
```diff
@@ -0,0 +1,27 @@
...
+  channel_id = req.user?.get('current_channel').id
+  queuedQuery = query "daily_email: true"
+  latestQuery = query "channel_id: \"#{channel_id}\", published: true, sort: \"-published_at\", daily_email: false"
+
+  client = new Lokka
+    transport: new Transport(API_URL + '/graphql')
+
+  client.query(queuedQuery).then (result) =>
+    if result.error
+      return res.backboneError
```
craigspaeth: I don't think you want to use `res.backboneError` here. Since this is now in GraphQL land we don't need that helper. You could use next here, e.g. `return next result.error if result.error` or more likely since Lokka uses promises you would just catch the errors all at once like...

````coffeescript
client.query(...)
  .then (result) =>
    # ...
  .catch(next)
````

craigspaeth:
❤️  the liberal use of React here. Well architected and tested. I have some small comments, but overall well done! 👏

PR #857, opened 2016-12-09
@eessex => Remove description from Admin and fix saving

kanaabe:
The reason we don't want to keep those fields in `serialize` anymore is that those inputs don't exist on the front-end. Soo they were defaulting to `false` which was actually removing things in the queue. 🙈


eessex:
oh my, good catch

Issue #859, opened 2016-12-16
Can't add section after callout


halleyjohnson:
The plus button doesn't pop up on hover at the bottom of a callout section @kanaabe

eessex:
@halleyjohnson, it looks like the edit tool below the callout is explicitly disabled.  

I'm noticing that there are also some limitations on where callouts can be placed (only between two text blocks), guessing the edit tools are disabled to help preserve this rule.

kanaabe:
@eessex yep, this was placed when we had callouts that were floated to the right. Now that they're stand-alone blocks we can remove those rules.

PR #861, opened 2016-12-19
@kanaabe => Remove callout edit restrictions

eessex:
Closes https://github.com/artsy/positron/issues/859

Remove restrictions on callouts
- Callouts can be added adjacent to any block
- Callouts have edit tools below them


kanaabe:
Woot! Thanks.

Issue #863, opened 2016-12-19
Queue Panel issues with articles remaining after flush


kanaabe:

kanaabe:
Potential solution: on Unpublish, make a delete call to ST

Issue #860, opened 2016-12-19
Artworks being lost after saving in a draft


kanaabe:

halleyjohnson:
https://writer.artsy.net/articles/585838f7c7bfd80011339847/edit
this artwork was in there: https://www.artsy.net/artwork/faith-ringgold-change-3-faith-ringgolds-over-100-pound-weight-loss-performance-storyquilt

halleyjohnson:
and I need to publish this one today, but there are links to the artworks that were in there multiple times and went missing multiple times: https://writer.artsy.net/articles/5851832e802d6d0011f0d726/edit

halleyjohnson:
@kanaabe

Issue #823, opened 2016-10-25
Autocomplete select issues


kanaabe:
Seems like some autocomplete select fields are not clearing out sometimes. An example came up with https://writer.artsy.net/articles/580fb0efd9207e21000429ed/edit where we saw it listed twice, and neither of them would delete.

![image](https://cloud.githubusercontent.com/assets/2236794/19702401/79ff3ace-9acd-11e6-9d31-13fd69397187.png)

kanaabe:
This came up again today

PR #864, opened 2016-12-20
@eessex => Writer bugs

kanaabe:
Closes https://github.com/artsy/positron/issues/863, https://github.com/artsy/positron/issues/860, https://github.com/artsy/positron/issues/823

- Be more specific about the `onKeyUp` event handler (exclude the triggers from the Display Panel). This is what was breaking the artwork saves that Halley suddenly started seeing.
- Handles an `unpublishing` state, which is when an article goes from `published: true` to `published: false`. This is necessary for updating ST by deleting the article when something goes unpublished
- Major 😅  typos with fair autocomplete


eessex:
So great you were able to hit all these bugs in 1 PR! Merging

Issue #820, opened 2016-10-17
Strip stop words from slugs


owendodd:
cc @halleyjohnson

eessex:
@kanaabe is there a collection of stop words that I can use for reference?

PR #866, opened 2016-12-20
@kanaabe => Remove stop words from slugs

eessex:
Closes https://github.com/artsy/positron/issues/820

- Removes stop words from slugs


api/apps/articles/model/save.coffee
```diff
@@ -33,7 +34,7 @@ artsyXapp = require('artsy-xapp').token or ''
 
 @onPublish = (article, cb) =>
   unless article.published_at
-    article.published_at = new Date()
+    article.published_at = new Date().toISOString()
```
kanaabe: Did you test this locally on both GraphQL and REST-based responses?

kanaabe:
On the right track so far! I would take a look at `Joi`, our validation library, for the ISO/Date issue.  There are some fancier things we can do regarding validating inputs for ISO. Please take a look! We want to actually validate that both saving and retrieving articles use ISO. 

Regardless of what format we use inside the API functions, Mongo will ultimately save dates in their own Date type. See https://docs.mongodb.com/manual/reference/method/Date/. Make a note of that. What does that mean for saving? 

Finally, we should add many more tests for stop words and dates. I think it's fine to tack on to other tests, but since stop words are a new feature, we should create tests on the method, as well as handling as many edge cases as possible.

eessex:
Cool, I'm going to pull the date line out of here for the moment -- sounds like I have a lot more research to do on this problem!  Will add some new tests for the stop words as well

kanaabe:
🙌  Good call

eessex:
Added a few new tests, let me know if there is another angle you'd like me to address 🙃

api/apps/articles/test/model/save.coffee
```diff
@@ -116,20 +117,42 @@ describe 'Save', ->
...
         done()
 
+  describe '#removeStopWords', ->
+
+    it 'removes stop words from a string', (done) ->
+      @removeStopWords("I'll be there").should.containEql ''
+      @removeStopWords('Why the Internet Is Obsessed with These Videos of People Making Things').should.containEql 'Why Internet Is Obsessed These Videos People Making Things'
+      @removeStopWords('Heirs of Major Jewish Art Dealer Sue Bavaria over $20 Million of Nazi-Looted Art').should.containEql 'Heirs Major Jewish Art Dealer Sue Bavaria  20 Million Nazi-Looted Art'
+      @removeStopWords('Helen Marten Wins UK’s Biggest Art Prize—and the 9 Other Biggest News Stories This Week').should.containEql 'Helen Marten Wins UK Biggest Art Prize 9 Other Biggest News Stories This Week'
+      done()
```
kanaabe: 👍

kanaabe:
LGTM!

PR #867, opened 2016-12-21
@craigspaeth => [Queue] Scheduled Articles List + GraphQL Permissions

kanaabe:
## What's New

- **View drafts with GraphQL**. View a draft by passing in an `X-Access-Token` in a header + providing a `channel_id`. This is exactly the same as our REST-based API.
- **Find articles that are scheduled**. Pass in `scheduled: true` to view a collection of articles with a `scheduled_publish_at` field
- **Update JoiQL** The latest version of JoiQL strips out the Koa-style middleware and only uses the schema conversion (wonderful idea by @craigspaeth)
- **Scheduled Articles in the Queue panel**: 
image

## TODO
- Tests
- Get Lokka working on Client


api/apps/graphql/resolvers.coffee
```diff
@@ -1,11 +1,17 @@
 _ = require 'underscore'
 { where, presentCollection } = Article = require '../articles/model'
+User = require '../users/model.coffee'
+
+module.exports.articles = (root, args, req, ast) ->
+  if (not args.published or args.scheduled) and not args.channel_id
```
craigspaeth: 👍  nice. I wonder if it'd make sense to abstract these permissions to work based on GraphQL args and query params (probably a leaky abstraction, but food for thought).

api/apps/graphql/resolvers.coffee
```diff
@@ -1,11 +1,17 @@
...
 
-module.exports.articles = (ctx, next) ->
-  return next() unless ctx.req.query.articles
-  return next() if ctx.req.query.articles.args?.published is false
   return new Promise (resolve, reject) ->
-    where _.extend(ctx.req.query.articles.args, published: true), (err, results) ->
-      ctx.res.articles = presentCollection(results).results
-      next()
-      resolve()
+    where args, (err, results) ->
```
craigspaeth: 👌  glad that worked out cleanly.

api/apps/graphql/resolvers.coffee
```diff
@@ -1,11 +1,17 @@
 _ = require 'underscore'
 { where, presentCollection } = Article = require '../articles/model'
+User = require '../users/model.coffee'
+
+module.exports.articles = (root, args, req, ast) ->
+  if (not args.published or args.scheduled) and not args.channel_id
+    throw new Error 'Must pass channel_id to view unpublished articles. Or pass ' +
+      'published: true to only view published articles.'
+
+  access = User.hasChannelAccess req.user, args.channel_id
```
craigspaeth: Ah nice that `req` gets passed into through the GraphQL tree so simply.

api/apps/graphql/resolvers.coffee
kanaabe: Ah good thought! I think we could do this on the [Joi side](https://github.com/hapijs/joi/blob/v10.0.6/API.md#alternativeswhenref-options), but I'm not sure if there's a way to bubble up a clear error message like: `Must pass channel_id to view unpublished articles. Or pass published: true to only view published articles.` should someone forget this rule.

kanaabe:
@craigspaeth I think this should be good now! Latest things I've added since you commented last was adding moving client fetching to superagent and tests!

craigspaeth:
👏  nice!

PR #868, opened 2017-01-03
@craigspaeth => Remove duplicate articles in Sailthru

kanaabe:
We're having an issue where sometimes multiple articles appear in Sailthru because Sailthru uses the slug as the key. When we send data to Sailthru, we should ensure that old slugs are also removed from Sailthru.

### DIFF
--- a/api/apps/articles/model/save.coffee
+++ b/api/apps/articles/model/save.coffee
@@ -323,7 +323,7 @@ typecastIds = (article) ->
     full: url: crop(imageSrc, { width: 1200, height: 706 } )
     thumb: url: crop(imageSrc, { width: 900, height: 530 } )
   html = if article.send_body then getTextSections(article) else ''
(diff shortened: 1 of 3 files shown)


api/apps/articles/model/save.coffee
```diff
@@ -342,9 +343,15 @@ typecastIds = (article) ->
     debug err if err
     cb()
 
-@deleteArticleFromSailthru = (article, cb) =>
+cleanArticlesInSailthru = (slugs = []) =>
+  if slugs.length > 1
+    _.map slugs, (slug, i) =>
```
craigspaeth: Minor: Not sure if `_.map` is needed here because you could just use a loop or `slugs.forEach`.

craigspaeth:
Solid! I only have a super minor stylistic comment, so I'll go ahead and merge.

craigspaeth:
Solid work!

Issue #873, opened 2017-01-05
Steps to reproduce hang from #861


craigspaeth:
TBD
Poll localhost

````
while true
do
	curl http://localhost:3005/api/articles; sleep 1
done
````

kanaabe:
I was able to get these three spikes to happen:
![image](https://cloud.githubusercontent.com/assets/2236794/21701755/8a280cf6-d375-11e6-95af-9143858f8dc0.png)

After a fresh restart, hit the API a bunch of times. Nothing happens and there are no memory spikes or slow requests. Then, hit the client. In all three spikes, requests failed after playing around with the client. In the first two, it happened on the first request to stagingwriter.artsy.net. Then a `MongoError` occurred which forced a restart. The third spike didn't have an Mongo Errors (perhaps connection to our API was already down so we didn't even get to Mongo tries) and I had to force a restart to clear again.

PR #876, opened 2017-01-06
@craigspaeth => Add better logging and update superagent

kanaabe:
The repro steps are [here](https://github.com/artsy/positron/issues/873). I can only reproduce it on staging which means we'll need to bubble up better error messages. I can determine it's caused from the client, and farthest traceable error in papertrail is: `superagent timeout 10000ms GET https://stagingwriter.artsy.net/api/articles`

Also updates `superagent` to 3.x.x from 1.x.x. I did some spot checking to make nothing breaks locally.


api/apps/articles/routes.coffee
```diff
@@ -16,6 +16,7 @@ User = require '../users/model.coffee'
       'Pass published=true to only view published articles.')
 
   Article.where req.query, (err, results) ->
+    debug err if err
```
craigspaeth: I would think [this `debug` in the error handler middleware](https://github.com/artsy/positron/blob/master/api/lib/middleware.coffee#L25) would take care of that no?

craigspaeth:
👍  to updating superagent, but see my comment above about the use of debug. Otherwise looks good.

api/apps/articles/routes.coffee
kanaabe: Hmm yeah you're right that it should get handed in the error handler. So assuming then that any error from the API fetch/mongo stuff is handled, what else could cause the hanging `superagent` request? `superagent timeout 10000ms GET https://stagingwriter.artsy.net/api/articles`. It's always after this error (called from the client) that memory consumption gets piled up.

kanaabe:
Ok I'll merge this without the debug. Basically just another shrinkwrap and update to `superagent`

PR #880, opened 2017-01-10
@kanaabe => Stopwords improvements

eessex:
- Eliminates bug that adds 'undefined' to slug on drafts
- Don't return empty if all words are stopwords
- Adjust stopwords list
- Remove digits


api/apps/articles/model/save.coffee
```diff
@@ -62,11 +62,13 @@ getDescription = (article) =>
   text
 
 removeStopWords = (title) ->
-  title = title.replace(/[.,\/#!$%\^&\*;:{}=\—_`’~()]/g," ")
-  newTitle = _.difference(title.split(' '), stopWords.stopWords).join(' ')
-  return newTitle
+  title = title.replace(/[0-9.\/#!$%\^\*;{}=_`’~()]/g,"")
```
kanaabe: Hmm why do we want to move numbers? Is this something that was discussed separately? I think this would affect drafts that don't have titles. In those cases we use dashed dates.

kanaabe: I see, thanks! If it doesn't affect drafts that works for me but in general I don't know if numbers should be a stop word. @owendodd can you comment?

eessex: @owendodd also would be good for you to check out the full list of stopwords -- https://github.com/eessex/positron/blob/2a2f784f70540b5b465e45437152d608d6246380/api/lib/stopwords.coffee

owendodd: I think we should keep numbers as non stop words. I think for lists, sometimes the number can be memorable and I could imagine users potentially using it in a search (thinking '30 under 30' and similar). I think that we could prob keep units as none stop words as well. Everything else looks good to me!

kanaabe:
Woot! 👏

PR #881, opened 2017-01-11
@kanaabe => ISO dates

eessex:
- Formats article date fields in ISO before GraphQL

Ended up pulling all the pre-save conversion out of this, as we do indeed need to send a Date object to the database, rather than ISO string, for sorting to work.

### DIFF
--- a/api/apps/articles/model/index.coffee
+++ b/api/apps/articles/model/index.coffee
@@ -118,12 +118,18 @@ Q = require 'bluebird-q'
   }
 
 @present = (article) =>
+  if article?.scheduled_publish_at
+    scheduled = moment(article.scheduled_publish_at).toISOString()
+  else
+    scheduled = null
+
   _.extend article,
     id: article?._id?.toString()
     _id: undefined
     slug: _.last article.slugs
     slugs: undefined
     published_at: moment(article?.published_at).toISOString()
+    scheduled_publish_at: scheduled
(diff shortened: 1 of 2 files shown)


kanaabe:
Excellent! Should we also include `scheduled_publish_at`? Glad we were able to come up with a simple solution. 😸

kanaabe:
Also, this doesn't have to be in this PR but we should add an assertion in https://github.com/artsy/positron/blob/master/api/apps/articles/test/model/index.coffee#L1069 eventually.

eessex:
@kanaabe, just added your suggestions

kanaabe:
⭐️

api/apps/articles/model/index.coffee
```diff
@@ -118,11 +118,19 @@ Q = require 'bluebird-q'
   }
 
 @present = (article) =>
+  if article?.scheduled_publish_at
+    scheduled = moment(article.scheduled_publish_at).toISOString()
+  else
+    scheduled = null
```
kanaabe: Ah ok, so we should also do this for `published_at` too, right? Since drafts don't have `published_at`

Could be a one-liner:
```
scheduled = if (date = article?.scheduled_publish_at) then moment(date).toISOString() else null
```

kanaabe:
Wahoo! Looks great.

PR #885, opened 2017-01-12
@kanaabe => published_at bugs

eessex:
- Joi doesn't allow null published_at,  If there is no publish date, give undefined rather than null.


kanaabe:
👍

PR #887, opened 2017-01-13
@craigspaeth => Support Channels and Curations in api/graphql

kanaabe:
Closes https://github.com/artsy/positron/issues/855

image

### DIFF
--- a/api/apps/curations/model.coffee
+++ b/api/apps/curations/model.coffee
@@ -20,12 +20,12 @@ OPTIONS = { allowUnknown: true, stripUnknown: false }
 #
 # Schemas
 #
-schema = (->
+@schema = (->
   id: @objectId()
   name: @string().allow('', null)
 ).call Joi
 
-querySchema = (->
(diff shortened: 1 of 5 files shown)


api/apps/curations/model.coffee
```diff
@@ -20,12 +20,13 @@ OPTIONS = { allowUnknown: true, stripUnknown: false }
 #
 # Schemas
 #
-schema = (->
+@schema = (->
   id: @objectId()
   name: @string().allow('', null)
+  content: @any()
```
kanaabe: This could be a solution for getting other fields in Curations, since it's a flexible data structure. We'd have to support `any` on the JoiQL side I believe. Happy to pull this part out until I sort things out with JoiQL!

kanaabe: Removed bc the app won't run otherwise 🙃

craigspaeth:
👌  easy peasy! Exciting!

PR #889, opened 2017-01-18
@kanaabe => Articles list QA

eessex:
- Can account for users who have drafts but no published articles
- Fix broken href on new articles link

### DIFF
--- a/client/apps/articles_list/routes.coffee
+++ b/client/apps/articles_list/routes.coffee
@@ -18,8 +18,8 @@ Transport = require('lokka-transport-http').Transport
       if result.articles.length > 0
         renderArticles res, req, result, true
       else
-        publishedQuery = query "published: false, channel_id: \"#{channel_id}\""
-        client.query(publishedQuery)
+        unpublishedQuery = query "published: false, channel_id: \"#{channel_id}\""
+        client.query(unpublishedQuery)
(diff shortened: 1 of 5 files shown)


client/apps/articles_list/routes.coffee
```diff
@@ -15,10 +15,20 @@ Transport = require('lokka-transport-http').Transport
...
-      res.render 'index',
-        articles: result.articles || []
-        published: true
-        current_channel: req.user?.get('current_channel')
-    .catch -> next()
\ No newline at end of file
+      if result.articles.length > 0
+        renderArticles res, req, result, true
+      else
+        publishedQuery = query "published: false, channel_id: \"#{channel_id}\""
```
kanaabe: It's a little confusing to have `publishedQuery` be a query for unpublished articles -- correct me if i'm wrong but this would automatically display the drafts view instead of published right? what if they click the published tab after the page initially loads?

eessex: Changed the naming for this query.  If they click the published tab it says 'no results found'

kanaabe: Ok, thanks 👍

kanaabe:
Ok looks good to me! Merging

PR #890, opened 2017-01-18
@kanaabe => Articles list can accept url query

eessex:
re: @halleyjohnson's request, new articles page is aware of URL queries 

'/articles?published=false' will query unpublished and go to drafts page

### DIFF
--- a/client/apps/articles_list/routes.coffee
+++ b/client/apps/articles_list/routes.coffee
@@ -6,17 +6,20 @@ Transport = require('lokka-transport-http').Transport
 
 @articles_list = (req, res, next) ->
   channel_id = req.user?.get('current_channel').id
-  publishedQuery = query "published: true, channel_id: \"#{channel_id}\""
+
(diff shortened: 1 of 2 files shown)


halleyjohnson:
Amazing! Thanks, Eve.

client/apps/articles_list/routes.coffee
```diff
@@ -6,17 +6,20 @@ Transport = require('lokka-transport-http').Transport
 
 @articles_list = (req, res, next) ->
   channel_id = req.user?.get('current_channel').id
-  publishedQuery = query "published: true, channel_id: \"#{channel_id}\""
+
+  if req._parsedUrl.query?.includes 'published=false' then published = false else published = true
```
kanaabe: i think you can use `req.query` here for the express-y way. i believe it's the safer option after doing a quick search for `_parsedUrl`. It seems like it's added as a caching optimization, so _might_ not be as reliable? i'm not sure..

Issue #891, opened 2017-01-18
Drafts showing incorrect last edited time


halleyjohnson:
Looks like something's up with the "last edited" timestamp, all are saying "a few seconds ago":

![screen shot 2017-01-18 at 5 23 17 pm](https://cloud.githubusercontent.com/assets/10982707/22085649/f787b0b2-dda2-11e6-97f8-f59aa9462a8b.png)

PR #894, opened 2017-01-19
@kanaabe => Updated at bug in articles list

eessex:
Closes https://github.com/artsy/positron/issues/891

Add updated_at to article schema and query

### DIFF
--- a/api/apps/articles/model/schema.coffee
+++ b/api/apps/articles/model/schema.coffee
@@ -74,6 +74,7 @@ denormalizedArtwork = (->
   tags: @array().items(@string()).default([])
   title: @string().allow('', null)
   layout: @string().allow('left', 'center', null).default('center')
+  updated_at: @date()
   published: @boolean().default(false)
   published_at: @date()
   scheduled_publish_at: @date().allow(null)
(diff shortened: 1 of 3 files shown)


kanaabe:
This means people _could_ manipulate this field which I'm ok with but a new field in the schema means we should have a test!

eessex:
Is there a way to include this in the query without adding it to the schema?  The application seems to crash without it.  Will add a test now.

kanaabe:
Not that I know of. We're using the same schema in two ways, originally to control what can be saved, and now for presentation/queryable fields in GraphQL. So probably the right thing to do is eventually separate these out. I think for now it's fine to add `updated_at` to the schema.

eessex:
The saving of this field is already tested against date here -- https://github.com/artsy/positron/blob/master/api/apps/articles/test/model/index.coffee#L435 -- and presentation here - https://github.com/artsy/positron/blob/master/api/apps/articles/test/model/index.coffee#L435

Is there another way we should be testing it?

kanaabe:
Riight so the new functionality is that you can save `updated_at` as an input. As opposed to it being auto-generated. So we should have tests for both! Lmk if that makes sense.

eessex:
Cool! That does make sense, just added more tests.

kanaabe:
Great thanks!

Issue #856, opened 2016-12-07
Smaller SVGs on sidebar


kanaabe:
On small screens, the profile icon gets hidden if you have access to queue. We should change to the same format maybe as CMS? 

On a 1250px screen:
![image](https://cloud.githubusercontent.com/assets/2236794/20990134/6d6e0660-bca5-11e6-8a05-48a7d6a834b1.png)

cc @owendodd

owendodd:
Wonder if we could consolidate these more, could see us moving drafts/published to a tab system similar to what exists in the new Queue view.

kanaabe:
Ah sure I'm for that. Also `Sections` should just be removed entirely.

craigspaeth:
👍  to consolidating drafts/published, I had originally considered that in the beginning of Writer, but there wasn't enough functionality to warrant it.

PR #902, opened 2017-01-24
@eessex => Remove client-side sections

kanaabe:
Closes https://github.com/artsy/positron/issues/856

We should still support it in the API since we support the 2015 Venice Biennale section still. However, we should remove the ability to CRUD them.


eessex:
goodbye sections! 
I have been noticing some weird action with the settings icon going black, is this what the new svg fixes or another issue?

kanaabe:
Yep! It was actually kinda weird -- it was inheriting a fill style from a different SVG that had the same class name. So that new SVG simply changes the name of the class that's used.

eessex:
:1st_place_medal:

kanaabe:
Mergin this!

PR #908, opened 2017-01-30
@eessex => Image size support

kanaabe:
All `ImageCollection` sections will now also provide a width/height. Same for `ImageSet`.

The WIP part is for the conversion process. We'll have to fetch images/artworks when converting Image and Artwork sections to ImageCollection so we can provide the width/height info. I'm playing around with moving this logic to the `Sections` collection where it seems fit a little better...sorry it's taking a while to get this out! Really hate halting other deploys for this kinda thing but 🤷‍♀️

### DIFF
--- a/api/apps/articles/model/schema.coffee
+++ b/api/apps/articles/model/schema.coffee
@@ -58,6 +58,9 @@ denormalizedArtwork = (->
         name: @string().allow('', null)
         slug: @string().allow('', null)
     ).allow(null).default([])
+    artist: @object().keys
+      name: @string().allow('', null)
+      slug: @string().allow('', null)
     width: @number().allow(null)
     height: @number().allow(null)
 ).call Joi
--- a/api/apps/articles/model/schema.coffee
+++ b/api/apps/articles/model/schema.coffee
@@ -57,9 +57,9 @@ denormalizedArtwork = (->
       @object().keys
         name: @string().allow('', null)
         slug: @string().allow('', null)
+    ).allow(null).default([])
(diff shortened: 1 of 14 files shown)


client/apps/edit/client.coffee
```diff
@@ -13,20 +13,94 @@ EditAdmin = require './components/admin/index.coffee'
...
+
+convertImageSet = (section, callback) ->
+  async.parallel(
+    section.get('images').map (image) -> (cb) ->
+      img = new Image()
+      img.src = image.image or image.url # image or artwork
+      img.onload = ->
+        image.width = img.width
+        image.height = img.height
+        image.artists = [image.artist] if image.type is 'artwork'
```
eessex: do all the image set artworks just have the singular artist field?

kanaabe: yeah i was going to do the same conversion process for image sets but i think we said that the frontend doesn't support it yet, right?

api/apps/articles/model/schema.coffee
```diff
@@ -56,6 +58,11 @@ denormalizedArtwork = (->
         name: @string().allow('', null)
         slug: @string().allow('', null)
     ).allow(null).default([])
+    artist: @object().keys
+      name: @string().allow('', null)
+      slug: @string().allow('', null)
```
kanaabe: Adding back in singular artist for now because it looks like ImageSet is broken on staging

kanaabe:
Merging this!

PR #910, opened 2017-01-31
@kanaabe => Queue displays email headlines and images

eessex:
Closes https://github.com/artsy/positron/issues/909

Queue/schedule panel displays email headline and image in list-views
Adds a 'display' prop to the articles list so email and potentially other metadata can be accessible in list views


kanaabe:
💯

PR #915, opened 2017-02-03
@craigspaeth => Be able to save zero sections

kanaabe:
### Problem
If you go from having sections in an article to deleting all the sections and try to save an empty article, it will fall back to its previous attrs. 

I think this predates me, but i wonder what the reason we wanted to fallback on attrs was. It seems like the reason was that on the [edit router](https://github.com/kanaabe/positron/blob/master/client/apps/edit/routes.coffee), we need to do `article.toJSON()` before rendering but we didn't have `article.sections` [defined](https://github.com/artsy/positron/blob/master/client/models/article.coffee#L17) when first initializing.

I'm sure there was a good reason at the time but if we listen for changes to the article's `attrs`, we can set `article.sections` after the sync is triggered.

### Solution
Listen to changes on the article, and set `article.sections` after the fetch. 
It's actually just the same thing as HeroSection [here](https://github.com/artsy/positron/blob/master/client/models/article.coffee#L24).

### DIFF
--- a/client/models/article.coffee
+++ b/client/models/article.coffee
@@ -15,6 +15,7 @@ module.exports = class Article extends Backbone.Model
 
   initialize: ->
     @sections = new Sections @get 'sections'
+    @on 'change:sections', => @sections.set @get 'sections'
     @featuredPrimaryArtists = new Artists
     @featuredArtists = new Artists
     @mentionedArtists = new Artists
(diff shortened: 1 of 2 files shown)


craigspaeth:
🤔  I can't remember why that is either, so 👍  thanks for the fix!

Issue #913, opened 2017-02-01
HeroSection image component broken


kanaabe:
HeroSection component likely needs to update its image saving requirements. I think this can be bundled with providing image heights/widths as well as the conversion process.

PR #917, opened 2017-02-07
@eessex => ImageCollection cleanup

kanaabe:
Closes https://github.com/artsy/positron/issues/913

- Remove `generateArtworks` because we're now generating artwork metadata on the client
- Sanitize `ImageCollection` and `ImageSet` captions
- Fix HeroSection
- Remove varying layouts from the Image component. Now only the HeroSection uses it, which has one layout.


eessex:
Looks good to me

kanaabe:
If a PR looks good and it's not [WIP], feel free to merge it! Makes the process a little smoother.

PR #918, opened 2017-02-07
@kanaabe => Update React to 15

eessex:
- Update all render functions to use ReactDOM.render
- Remove all instances of getDOMNode
- Add display names (for easy debugging)
- Add keys when erroring
- Update all tests to use reactTestUtils, reactDOMServer for rendering

There are still two tests in here that aren't passing, will wrap those up soon


kanaabe:
Woo very exciting! 

It looks like the shrinkwrap was removed. I'm hesitant to re-shrink everything because of those memory leak issues we had a while back so I think the best bet is to copy over the current shrinkwrap and `npm install <pkg> --save` each package individually. Probably could even be --save-dev for the test utils.

client/apps/edit/components/section_container/index.coffee
```diff
@@ -61,6 +63,7 @@ module.exports = React.createClass
           when 'fullscreen' then SectionFullscreen
           when 'callout' then SectionCallout
           when 'toc' then SectionToc
+          when 'image' then SectionImage
```
kanaabe: Ah see line 18 and 69, I've added these already from [this PR](https://github.com/artsy/positron/pull/917/commits/917cf32ef04dafd4313cd445b5092a4643ed81a1).

client/apps/edit/components/section_embed/index.coffee
```diff
@@ -8,6 +8,7 @@ sd = require('sharify').data
 { div, section, label, nav, input, a, h1, p, strong, span, form, button, iframe } = React.DOM
 
 module.exports = React.createClass
+  displayName: 'SectionEmbed'
```
kanaabe: Thanks for adding these!

client/apps/edit/components/section_embed/index.coffee
```diff
@@ -26,9 +27,9 @@ module.exports = React.createClass
 
   submitUrl: (e) ->
     e.preventDefault()
-    url = @refs.url.getDOMNode().value
-    height = @refs.height.getDOMNode().value
-    mobileHeight = @refs.mobileHeight.getDOMNode().value
+    url = @refs.url.value
+    height = @refs.height.value
+    mobileHeight = @refs.mobileHeight.value
```
kanaabe: So much cleaner ✨

kanaabe:
Very awesome work here! After resolving that one dupe comment, this is ready to be merged.

eessex:
Cool! Thanks for noticing the duplicates -- didn't see them when I pulled changes in -- gonna go ahead and merge when this turns green :)

PR #922, opened 2017-02-08
@eessex => Fix packages

kanaabe:
Some deps seemed to be missing from the shrinkwrap, not sure how that happened but i think this should fix it.


kanaabe:
Integration tests kept on failing on this PR and it was because we had NODE_ENV=production in Semaphore. I've removed it for now since it looked like that wasn't the fix for React stuff anyway.

PR #931, opened 2017-02-14
@craigspaeth => Switch packaging to Yarn

eessex:
- Remove shrinkwrap and switch to Yarn for package management

Will keep a close eye on this in staging for a day or so before moving to production.


craigspaeth:
W00t! Another project on Yarn! Mind updating the [setup](https://github.com/artsy/positron#set-up) to add a "Install Yarn"

```
- Install Yarn
npm install -g yarn

- Install node modules
yarn install
```

We should also make sure to get this deployed first https://github.com/artsy/positron#set-up before we deploy these changes to minimize the changeset in case of any issues with the updates set of modules.

eessex:
@craigspaeth  good call!  Just added this to the mix

PR #932, opened 2017-02-14
@craigspaeth => Artwork by URL bug

eessex:
State was being set on an object that was not a react component, which breaks with the new React version. 
Loading indicator and error message are now set using jQuery.

### DIFF
--- a/client/apps/edit/components/section_image_set/index.coffee
+++ b/client/apps/edit/components/section_image_set/index.coffee
@@ -94,16 +94,18 @@ module.exports = React.createClass
     e.preventDefault()
     val = @refs.byUrl.value
     slug = _.last(val.split '/')
-    @refs.byUrl.setState loading: true
+    $(@refs.byUrl).siblings('button').addClass('is-loading')
     new Artwork(id: slug).fetch
       error: (m, res) =>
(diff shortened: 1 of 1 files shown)


client/apps/edit/components/section_image_set/index.coffee
```diff
@@ -94,16 +94,18 @@ module.exports = React.createClass
     e.preventDefault()
     val = @refs.byUrl.value
     slug = _.last(val.split '/')
-    @refs.byUrl.setState loading: true
+    $(@refs.byUrl).siblings('button').addClass('is-loading')
```
craigspaeth: Using jQuery to manipulate the DOM like this is typically an anti-pattern in React-land and usually you'd want to reflect UI updates by branching logic within the `render` function. This style of jQuery code is referred to as "imperative" programming whereas React encourages "declarative" programming—that is instead of saying "find this element, change this class" React would encourage saying "the UI looks like this when the state is updated". That said, I know this is a quick time-sensitive fix so I'm not going to hold it up—but we should dig in more when we pair tomorrow.

craigspaeth:
Thanks for the quick fix! This works, but I noted what might be a better approach to explore which we can refactor later.

PR #933, opened 2017-02-15
@craigspaeth => Refactor add-url

eessex:
Refactor of addArtworkFromUrl in imageset and image collection to use React refs rather than jquery for dom manipulation. 

Also updates the captions on imageset to match Force styles.

### DIFF
--- a/client/apps/edit/components/section_image_collection/index.coffee
+++ b/client/apps/edit/components/section_image_collection/index.coffee
@@ -115,18 +115,18 @@ module.exports = React.createClass
     e.preventDefault()
     val = @refs.byUrl.value
     slug = _.last(val.split '/')
-    $(@refs.byUrl).siblings('button').addClass('is-loading')
+    @refs.urlAdd.classList.toggle 'is-loading'
+    @refs.byUrl.value = ''
     new Artwork(id: slug).fetch
       error: (m, res) =>
         if res.status is 404
-          $(@refs.byUrl).val('').attr('placeholder', 'Artwork not found')
+          @refs.byUrl.placeholder = 'Artwork not found'
           setTimeout( =>
-            $(@refs.byUrl).siblings('button').removeClass('is-loading')
-            $(@refs.byUrl).attr('placeholder', 'Add artwork url')
+            @refs.urlAdd.classList.toggle 'is-loading'
+            @refs.byUrl.placeholder = 'Add artwork url'
           , 3000)
       success: (artwork) =>
-        $(@refs.byUrl).removeClass('is-loading').val ''
-        $(@refs.byUrl).siblings('button').removeClass('is-loading')
+        @refs.urlAdd.classList.toggle 'is-loading'
         newImages = @state.images.concat [artwork.denormalized()]
         @setState images: newImages
         @props.section.set images: newImages
@@ -208,6 +208,7 @@ module.exports = React.createClass
               placeholder: 'Add artwork url'
             }
               button {
+                ref: 'urlAdd'
                 className: 'esic-byurl-button avant-garde-button'
                 onClick: @addArtworkFromUrl
               }, 'Add'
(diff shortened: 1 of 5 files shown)


client/apps/edit/components/section_image_collection/index.coffee
```diff
@@ -208,6 +208,7 @@ module.exports = React.createClass
               placeholder: 'Add artwork url'
             }
               button {
+                ref: 'urlAdd'
                 className: 'esic-byurl-button avant-garde-button'
```
craigspaeth: Here's where you would ideally change the class by checking the component's state. For instance, instead of doing `@refs.urlAdd.classList.toggle 'is-loading'` you would do `@setState loading: true` then here you could write `className: 'esic-byurl-button avant-garde-button' + if @state.loading then ' is-loading' else ''`

craigspaeth:
Left a comment about using state + render over refs. Refs work okay too but not much better than using jQuery, you see how Facebook talks about this in [their docs](https://facebook.github.io/react/docs/refs-and-the-dom.html)

> There are a few good use cases for refs:

> - Managing focus, text selection, or media playback.
> - Triggering imperative animations.
> - Integrating with third-party DOM libraries.
> Avoid using refs for anything that can be done declaratively.

client/apps/edit/components/section_image_collection/index.coffee
```diff
@@ -208,7 +212,7 @@ module.exports = React.createClass
               placeholder: 'Add artwork url'
             }
               button {
-                className: 'esic-byurl-button avant-garde-button'
+                className: 'esic-byurl-button avant-garde-button' + isLoading
```
craigspaeth: 👌

client/apps/edit/components/section_image_collection/index.coffee
```diff
@@ -115,20 +116,19 @@ module.exports = React.createClass
     e.preventDefault()
     val = @refs.byUrl.value
     slug = _.last(val.split '/')
-    $(@refs.byUrl).siblings('button').addClass('is-loading')
+    @setState isLoading: !@state.isLoading
+    @refs.byUrl.value = ''
```
craigspaeth: The same pattern would apply here. Instead of setting `@refs.byUrl.value = ''` or `@refs.byUrl.placeholder = ` you would use `@setState` and reflect the updated UI in the `render function.

e.g.

````coffeescript
componentDidMount: ->
  @setState isLoading: !@state.isLoading
  new Artwork().fetch
    error: ->
      @setState placeholder: 'Artwork not found' if res.status if 404
      setTimeout => @setState placeholder: 'Add artwork url', isLoading: false

render: ->
  input {
    value: if @state.isLoading then '' else null
    placeholder: @state.placeholder
  },
````

_The React Way_ is that basically _everything_ that can change over time should go into state and if it's "config" that doesn't change over time it passes through props. Using refs or manipulating elements directly is more of an anti-pattern.

I hope this is helpful and not annoyingly pedantic 😶

**Comment by eessex** [comment]
@craigspaeth, tried to make this a bit drier by moving the input into its own component that can be shared between image_set and image_collection

File: client/apps/edit/components/section_image_set/index.coffee [hunk 4]
```diff
@@ -10,6 +10,7 @@ icons = -> require('./icons.jade') arguments...
 Autocomplete = require '../../../../components/autocomplete/index.coffee'
 Artwork = require '../../../../models/artwork.coffee'
 Input = React.createFactory require './input.coffee'
+UrlArtworkInput = React.createFactory require './url_artwork_input.coffee'
```
craigspaeth: 👍 `React.createFactory` didn't know about that one!

client/apps/edit/components/section_image_set/url_artwork_input.coffee
```diff
@@ -0,0 +1,55 @@
...
+  addArtworkFromUrl: (e) ->
+    e.preventDefault()
+    slug = _.last(@state.url.split '/')
+    @setState isLoading: !@state.isLoading, url: ''
+    new Artwork(id: slug).fetch
+      error: (m, res) =>
+        if res.status is 404
+          @setState error: !@state.error
+          setTimeout( =>
+            @setState
```
craigspaeth: 👌

craigspaeth:
👏  refactors look great! Thanks for working through those!

PR #936, opened 2017-02-17
@craigspaeth => Yarn / bcrypt require

eessex:
An experimental second attempt at Yarn -- Yarn replied to my issue noting a change to bcrypt's require statement should resolve our issue.


craigspaeth:
`Cannot find module 'bcryptjs'` from Semaphore might need to do a `yarn upgrade`.

eessex:
Looks like we actually need a different module, a vanilla javascript version of bcrypt -- https://www.npmjs.com/package/bcryptjs

Sounds like it is the same except 30% slower. What do you think?

craigspaeth:
I don't suspect that'll be a bottleneck for us since we only use it to encrypt access token as we cache session data, so 👍

PR #937, opened 2017-02-17
@craigspaeth => move uglify to production dependencies

eessex:
One step closer on yarn -- for some reason this error doesn't even appear in staging, but in production heroku wants access to uglify


craigspaeth:
Werid, not sure where that's used in production but happy to move forward. Thanks for slogging through these changes!

eessex:
@craigspaeth, yeah seems weird that would hold things up in production, but very excited to see that it seems to be running in production now 🎉

PR #940, opened 2017-02-24
@craigspaeth => Hero section tests running

eessex:
Repairs some tests that were broken after updating React


craigspaeth:
Nice work!

PR #943, opened 2017-03-01
@craigspaeth => Moves particle up to dependencies from devDependencies

kanaabe:
Had a silly issue last night when trying to deploy where the newly added `particle` dependency was added as a devDependency instead of a regular one. App crashed for a few minutes because it couldn't find the `particle` module. Since it had the NODE_ENV=production, this failed in production but not in staging. I'm going to add a `yarn install --production` to the staging build steps in Semaphore to avoid this kind of error in the future!


craigspaeth:
Good catch, this has bit us in other apps too where you don't notice it until a production deploy because everything else uses full installs. I think `yarn install --production` in Semaphore is a solid fix with Positron because that should surface issues in its [integration tests](https://github.com/artsy/positron/blob/master/test/integration.coffee), but I wonder if we can also have staging install production only dependencies as well.

kanaabe:
Oh yeah that's what i meant, to add `--production` to the build _and_ staging deploy. 👍 👍

PR #948, opened 2017-03-02
@eessex => Mark article as published when sending to FB

kanaabe:
We should indicate that the article is published if we're in NODE_ENV='production'. Currently staging articles are being published to FB's IA manager, but production isn't, and I think it's because of this flag. It works for me now if I do this locally with the flag set. 

https://developers.facebook.com/docs/instant-articles/api/


eessex:
👍

PR #949, opened 2017-03-03
Handle missing thumbnails in search results

joeyAghion:
I noticed client-side errors when hitting the v2 search API from this app:

    Uncaught TypeError: Cannot read property 'href' of undefined
      at Object.filter (main-e6c1d45e.js.jgz:3)

I reproduced it locally and it corresponds with article search results that don't have images. The new code tolerates that better. (This could use a test, but I couldn't find a close enough example and wasn't so bold as to create a new one.)

A reasonable question is why the new ES-backed search results lack some images while the Google-backed results didn't. I'm not sure, but [this](https://github.com/artsy/positron/blob/694736206d1998f5e956afa59ffe15f158f846ec/api/apps/articles/model/distribute.coffee#L125) is where we specify the corresponding image for ES. Maybe there's a better or more commonly available image instead?

### DIFF
--- a/client/apps/edit/components/section_artworks/index.coffee
+++ b/client/apps/edit/components/section_artworks/index.coffee
@@ -51,7 +51,7 @@ module.exports = React.createClass
             vals.push
               id: id
               value: r.title
-              thumbnail: r._links.thumbnail.href
+              thumbnail: r._links.thumbnail?.href
         return vals
       templates:
         suggestion: (data) ->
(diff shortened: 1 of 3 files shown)


eessex:
Hi Joey, thanks for tackling this.  The changes make sense for dealing with images missing from artworks.  Regarding the second question, do you mean that you're getting articles in search that don't have images?  This is a required field so they should be there, maybe there are older entries that do not.

Issue #944, opened 2017-03-01
Queue items intermittently not saving


kanaabe:
Sometimes articles added to the queue don't 'stick'. They have to be added multiple times to save.

cc @halleyjohnson

PR #950, opened 2017-03-06
@craigspaeth => Queue bug fix and loading/error flash

kanaabe:
Closes https://github.com/artsy/positron/issues/944

It was reported that sometimes items to the queue take a few clicks to save properly. Attempting two things in this PR. First one is a real bug that happens when you try to save an article with an old 'image' section. It's solved by the first commit. Second is a preventative measure that displays a loading spinner until the save callback is finished, and also displays an error message if anything went wrong!

queue-saving
queue-error-msg

### DIFF
--- a/client/apps/queue/client/client.coffee
+++ b/client/apps/queue/client/client.coffee
@@ -19,13 +19,23 @@ module.exports = QueueView = React.createClass
     queuedArticles: @props.queuedArticles or []
     scheduledArticles: @props.scheduledArticles or []
     feed: @props.feed or 'scheduled'
+    loading: false
+    errorMessage: ''
 
   saveSelected: (data, isQueued) ->
-    article = new Article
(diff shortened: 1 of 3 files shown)


client/apps/queue/client/client.coffee
```diff
@@ -19,13 +19,23 @@ module.exports = QueueView = React.createClass
...
-    article.save()
+    @setState loading: true
+    new Article().save article,
+      success: =>
+        @setState loading: false
+      error: =>
+        @setState
+          loading: false
+          errorMessage: 'There has been an error. Please contact support.'
+        setTimeout ( => @setState(errorMessage: '')), 2000
```
craigspaeth: 👍  solid state managing here.

craigspaeth:
LGTM!

Issue #838, opened 2016-11-08
Consider moving count out of every API request


kanaabe:
The performance of `count()` in mongodb isn't great, and we don't really need it anywhere. We should make this an optional field.

PR #951, opened 2017-03-06
@craigspaeth => Make count/total optional in API response

kanaabe:
Closes https://github.com/artsy/positron/issues/838

There are a few places in Force/MG/maybe-Metaphysics that use `count`, so once this gets merged, I'll be sure to add `count: true` flags to those fetches before deploying this one to production. I considered making `count: true` the default, and make `count: false` optimizations manually on the client, but it probably doesn't make sense for a client to have to think about how it affects the database 👀 .

### DIFF
--- a/api/apps/articles/model/index.coffee
+++ b/api/apps/articles/model/index.coffee
@@ -18,7 +18,7 @@ Q = require 'bluebird-q'
 # Retrieval
 #
 @where = (input, callback) ->
-  retrieve.toQuery input, (err, query, limit, offset, sort) ->
+  retrieve.toQuery input, (err, query, limit, offset, sort, count) ->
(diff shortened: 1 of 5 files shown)


joeyAghion:
I don't have much code input, but seems like Force and Metaphysics (I _think_ we can ignore Microgravity at this point, right?) should be updated to pass the new parameter in first. I'm assuming the new parameter will be ignore by the current code.

kanaabe:
Yep! Already done here: https://github.com/artsy/force/pull/974

craigspaeth:
Solid! Thanks! This also points out a nice part of how GraphQL can avoid expensive queries by not having to make assumptions about what the client needs.

PR #952, opened 2017-03-08
@craigspaeth => Adds _id to presentation

kanaabe:
At 10:07AM, this error started causing problems -- https://papertrailapp.com/systems/positron-production/events?focus=776149217075957783&selected=776149217075957783

I'm still trying to determine why, but those kinds of requests have started to pile up and even brought down some requests at one point. Positron recovered on its own but currently memory usage is high. 

[Here](https://github.com/artsy/force/blob/master/desktop/models/article.coffee#L69) is where we use the `super_article_for` field, and somehow, based on the logs, it's getting passed a slug instead of the id. Maybe this has to do with caching? We do set `id` [here](https://github.com/artsy/force/blob/master/desktop/apps/article/routes.coffee#L14). Anyway my idea to mitigate this is to add in the `_id` field for guaranteed mongoid.

### DIFF
--- a/api/apps/articles/model/index.coffee
+++ b/api/apps/articles/model/index.coffee
@@ -124,7 +124,7 @@ Q = require 'bluebird-q'
   published = if (date = article?.published_at) then moment(date).toISOString() else undefined
   _.extend article,
     id: article?._id?.toString()
-    _id: undefined
+    _id: article?._id?.toString()
     slug: _.last article.slugs
     slugs: undefined
     published_at: published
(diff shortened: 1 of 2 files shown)


craigspaeth:
That is a really strange sudden spike in memory usage. I'm very skeptical that a validation error like that would lead to that kind of impact on the Positron server—but happy to merge this regardless. I wonder if Force's logs/NewRelic around the same time would give more insight. Also FWIW we could probably bump up Positron's dynos to 2X and add the `--max_old_space_size=960` to avoid crashes (Node's V8 defaults to 1.4gbs and Heroku doesn't accommodate for this despite being on ~1/2 gb servers).

Issue #953, opened 2017-03-08
Search bar broken


eessex:
Filter search is erroring on articles and queue pages
![screen shot 2017-03-08 at 2 31 46 pm](https://cloud.githubusercontent.com/assets/1497424/23720188/0831dda4-040c-11e7-9f29-a04c3370296d.png)

PR #954, opened 2017-03-08
@kanaabe => Filter search bug

eessex:
Closes https://github.com/artsy/positron/issues/953
Removes total and count from filter search.


kanaabe:
Whoops sorry I didn't see this one.

PR #957, opened 2017-03-09
@eveessex => [WIP] Airtable <> Parsely integration

kanaabe:
This might not be the best place to put this script, but since we have the Heroku Scheduler running on this app anyway, it's not _terrible_. 

There are a couple places we could be more efficient about this script, let me know if you notice anything!

[WIP] because 1. rethink speed, 2. need final base information from Halley after we meet tomorrow.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "test": "make test"
   },
   "dependencies": {
+    "airtable": "^0.4.5",
     "artsy-backbone-mixins": "git://github.com/artsy/artsy-backbone-mixins.git",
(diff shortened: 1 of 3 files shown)


eessex:
Regarding the placement comment, is it because you aren't touching things positron specific besides variables?

kanaabe:
Yep! The ideal future combines all flows into Writer, so maybe this is an ok direction.

kanaabe:
Closing this one, this version doesn't pull data that's very useful for editorial. Might opt for a csv download instead

Issue #878, opened 2017-01-09
Find solution for Scribe/AMD


kanaabe:
Scribe uses AMD and doesn't add deamdify to their package.json locally. Browserify doesn't have an easy way of forcing a transform on a sub module. We've had issues with packages like Lokka and Yoast because of this.

Options:
1. Send PRs to all the Scribe repos to add `"browserify": { "transform": [ "deamdify" ] }`
2. Use [Draft.js](https://github.com/facebook/draft-js). It's well-maintained and since we're hopping on the FB bandwagon, it seems to fit. This might be a lot of work up front (ie creating [custom components](https://facebook.github.io/draft-js/docs/advanced-topics-block-components.html#content) for the custom scribe plugins we currently have) but probably worth it in the end.
3. Make Yoast and Lokka happy with deamdify

Related conversations:
https://github.com/guardian/scribe/pull/175#issuecomment-240230366
https://github.com/substack/node-browserify/issues/1127

eessex:
[ ] get draft running in positron
[ ] caption input
[ ] textarea input

CAPTION 
[ ] Doesn't allow line breaks
[ ] Strips only not-allowed html on paste

TEXT EDITOR
[ ] Can accept custom link-type for Artist
[ ] Can accept custom link-type for TOC
[ ] Custom types are converted correctly

Issue #664, opened 2016-01-29
Disable italics for H3's


owendodd:
<img width="437" alt="screen shot 2016-01-27 at 5 47 04 pm" src="https://cloud.githubusercontent.com/assets/3196536/12690267/437ce236-c6b0-11e5-8a36-3400abe4575b.png">

PR #967, opened 2017-03-14
@kanaabe => Draft text component

eessex:
Closes https://github.com/artsy/positron/issues/878, https://github.com/artsy/positron/issues/664
- [x] Tests requiring text selection
- [x] Weird styling on OL and UL
- [x] Strikethrough converted to html correctly


kanaabe:
Would be nice to get a couple gifs or screenshots in here!

eessex:
Strange -- got the text selection running in tests locally, and they seem to be working fine. Semaphore is not happy, and aborting the test in a memory crash. 
screen shot 2017-03-14 at 6 53 06 pm

eessex:
draft-text-input

kanaabe:
I don't think this is contributing to the GC issue, but I'm seeing a lot of these warnings in the tests -- probably would be good to address these. 

```
WARNING: DraftEntity.create will be deprecated soon!
Please use "contentState.createEntity" instead.
```

As for the GC issue, have you tried rebuilding the suite? I think you already did this but make sure to teardown after benv setups too. Also look out for anywhere that we might be hanging on to some data -- it's not very clear yet if it's coming from the test or the code so try and determine that. The semaphore machine is probably much smaller than yours so it's probably choking on memory earlier.

eessex:
I think the test crash is now under control -- will try pushing again soon to double check.  The warning is caused by our converter plugin, and happens with both available libraries.  Looks like multiple people have PR'd a fix for this problem, but it only works on the recent version of draft and so hasn't been merged yet.

client/apps/edit/components/section_image_set/input.coffee
```diff
@@ -1,70 +0,0 @@
...
-          disabled: if @props.caption then false else true
-          onClick: @onEditableKeyup
-        }
-        button {
-          'data-command-name': 'linkPrompt'
-          dangerouslySetInnerHTML:
-            __html: "&nbsp;" + $(icons()).filter('.link').html()
-          disabled: if @props.caption then false else true
-          onClick: @onEditableKeyup
-        }
```
kanaabe: Makes me so happy to be getting rid of Scribe!

client/apps/edit/components/section_text/index.coffee
```diff
@@ -1,145 +1,390 @@
...
-keyboardShortcutsMap =
-  bold: (e) -> e.metaKey and e.keyCode is 66
-  italic: (e) -> e.metaKey and e.keyCode is 73
-  h2: (e) -> e.metaKey && e.keyCode is 50
-  h3: (e) -> e.metaKey && e.keyCode is 51
-  removeFormat: (e) -> e.altKey and e.shiftKey and e.keyCode is 65
-  linkPrompt: (e) -> e.metaKey and not e.shiftKey and e.keyCode is 75
-  unlink: (e) -> e.metaKey and e.shiftKey and e.keyCode is 75
-  insertOrderedList: (e) -> e.metaKey and e.shiftKey and e.keyCode is 56
-  insertUnorderedList: (e) -> e.metaKey and e.shiftKey and e.keyCode is 55
```
kanaabe: Do we lose this functionality or does Draft come with keyboard shortcuts?

eessex: Have been speaking with @halleyjohnson about this, need to do a little more research on overriding browser/os commands, as I haven't yet gotten their key command override functioning correctly for multiple keys at once. We currently have clipboard events, undo/redo, bold and italic.

client/apps/edit/components/section_text/index.coffee
```diff
@@ -1,145 +1,397 @@
...
+    return blocksFromHTML
+
+  convertToHtml: (editorState) ->
+    html = convertToHTML({
+      entityToHTML: (entity, originalText) ->
+        if entity.type is 'LINK'
+          sanitizeName = originalText.split(' ')[0].replace(/[.,\/#!$%\^&\*;:{}=\-_`’'~()]/g,"")
+          name = if entity.data.name then ' name="' + sanitizeName + '"' else ''
+          if entity.data.className.includes('is-follow-link')
+            return '<a href="' + entity.data.url + '" class="' + entity.data.className + '"' + name + '>' + originalText + '</a><a class="entity-follow artist-follow"></a>'
```
kanaabe: The sanitize could cause problems for artist follows. The way the follow link works is that you paste in a url that is an artsy artist page (artsy.net/artist/andy-warhol), we take the artist slug (andy-warhol), and use that as the name. We need that exact slug (which is likely something with dashes) to make the follow button work on Force. On the Force side, we send a follow request to Gravity which needs the artist slug. 

Definitely would be good to test this on Force too! You can point to staging and log in to make sure the follow button works as expected. Also would be worth doing the same for the other components.

eessex: Yeah was wondering about this -- writer doesn't currently have any input validation on that field to prove it is a link or an artsy artist, so maybe we could add a requirement here.

client/apps/edit/components/section_text/index.coffee
kanaabe: Cool, definitely low priority but good that it's on your radar. We don't advertise the more "specialty" commands anyway so I doubt people are using them but thanks for checking in with @halleyjohnson.

client/apps/edit/components/section_text/index.coffee
```diff
@@ -1,145 +1,397 @@
...
+            return a { name: sanitizeName, className: entity.data.className}
+          else
+            return a { href: entity.data.url}
+        return originalText
+      blockToHTML: (block) ->
+        if block.type is 'header-three'
+          return h3 {}, block.text
+      styleToHTML: (style) ->
+        if style is 'STRIKETHROUGH'
+          return span { style: {textDecoration: 'line-through'}}
```
kanaabe: Love when projects like this add little gifts!🥇 !!

client/apps/edit/components/section_text/index.coffee
```diff
@@ -1,145 +1,397 @@
...
+    for block, i in boldBlocks
+      if block.style.fontWeight is 'normal'
+        $(doc.getElementsByTagName('B')[i]).replaceWith(doc.getElementsByTagName('B')[i].innerHTML)
+    # replace bold spans with actual b tags
+    boldSpans = doc.getElementsByTagName('SPAN')
+    for span, i in boldSpans
+      if span?.style.fontWeight is '700'
+        newSpan = '<strong>' + span.innerHTML + '</strong>'
+        $(doc.getElementsByTagName('SPAN')[i]).replaceWith(newSpan)
+    return doc.innerHTML
```
kanaabe: Google Docs styles: so ugly :'( 

Nice that we can manage this in Positron now tho, I always did find it annoying to go into the Scribe plugin to make modifications.

client/apps/edit/components/section_text/index.coffee
eessex: Actually don't think the sanitize will effect follow links, it only applies to the name value on jump links and doesn't change the url.

client/apps/edit/components/section_text/index.coffee
```diff
@@ -1,145 +1,397 @@
...
+
+  getSelectedBlock: ->
+    { editorState } = @state
+    selection = editorState.getSelection()
+    contentState = editorState.getCurrentContent()
+    startKey = selection.getStartKey()
+    startOffset = selection.getStartOffset()
+    closestBlock = contentState.getBlockForKey(startKey)
+    blockKey = closestBlock.getEntityAt(startOffset)
+    return {key: blockKey, content: closestBlock}
```
kanaabe: Nice and clear

client/apps/edit/components/section_text/index.coffee
kanaabe: Sorry, what I mean is not `name` but `data-id` which should be present on the actually follow button, the second link. 

Here's a sample of how the follow links look like currently:
<a href=\"https://www.artsy.net/artist/ej-hill\" class=\"is-jump-link is-follow-link\" name=\"EJ_Hill\">EJ Hill</a><a data-id=\"ej-hill\" class=\"entity-follow artist-follow\"></a>

[and here's what I did in the Scribe plugin](https://github.com/artsy/scribe-plugin-follow-link-tooltip/blob/master/index.js#L193)

client/apps/edit/components/section_text/test/index.coffee
```diff
@@ -1,63 +1,221 @@
...
+      html = '<b style="font-weight:normal;" id="docs-internal-guid-ce2bb19a-cddb-9e53-cb18-18e71847df4e"><p><span style="font-size:11pt;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre-wrap;">Available at: Espacio Valverde • Galleries Sector, Booth 9F01</span></p>'
+      @component.stripGoogleStyles(html).should.eql '<p><span style="font-size:11pt;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre-wrap;">Available at: Espacio Valverde • Galleries Sector, Booth 9F01</span></p>'
+
+    it 'replaces bold spans with actual b tags', ->
+      html = '<p><span style="font-size:11pt;color:#000000;background-color:transparent;font-weight:700;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre-wrap;">B. 1981 in Madrid. Lives and works in Madrid.</span></p>'
+      @component.stripGoogleStyles(html).should.eql '<p><strong>B. 1981 in Madrid. Lives and works in Madrid.</strong></p>'
```
kanaabe: Great set of tests here! Glad you got this figured out

client/apps/edit/components/section_text/index.coffee
eessex: aha! got it, will add that in soon

kanaabe: ⚡️

kanaabe:
I'm so stoked for this to go out! The code is clear and broken out nicely. Thanks for moving around files to make sure components live in a sensible place. Feel free to make this a follow-up PR, but it'd be nice to get rid of the scribe packages as well as the global transform we do to support it. 

Just that follow update + making sure the component bits render nicely in Force and this should be good to go! Congrats! 💯 🎊

eessex:
Added support for 'data-id' on the artist follow buttons, have been using the staging database to test Force formatting and everything seems to be working ATM. 

I will address the scribe packages in a separate PR, it looks like the title and lead paragraph are still using them and will need to be converted as well before we remove scribe fully. What do you mean by global transform?

kanaabe:
We globally transform packages that are AMD modules to CommonJS via the `deamdify` package. But we only needed this for Scribe so now we can remove it: 
https://github.com/artsy/positron/blob/74c5611d004687347efa0109e1582b9db94d0fb5/client/lib/setup/env.coffee#L19

kanaabe:
🌟  Nice job!

PR #968, opened 2017-03-15
@eessex => Tag API

kanaabe:
Adds new fields for `vertical`, `sub_vertical`, and `internal_tags`.
Breaks out the api/article/test/model/index.coffee test file into `index_save.coffee` and `index_where.coffee` since that file was getting too long. 

All tags are arrays of strings. I'll need to confirm on Friday if this is definitely what we want (vertical and subvertical may just be a string) so please don't merge just yet :) Feel free to review though.


api/apps/articles/model/schema.coffee
```diff
@@ -75,7 +75,10 @@ denormalizedArtwork = (->
   thumbnail_title: @string().allow('', null)
   thumbnail_teaser: @string().allow('', null)
   thumbnail_image: @string().allow('', null)
+  vertical: @array().items(@string()).default([])
```
eessex: is there a chance these may turn into something more like a curation, with an object id?

api/apps/articles/model/retrieve.coffee
```diff
@@ -33,6 +33,7 @@ moment = require 'moment'
       query.biography_for_artist_id = ObjectId input.biography_for_artist_id
     query.featured_artwork_ids = ObjectId input.artwork_id if input.artwork_id
     query.tags = { $in: input.tags } if input.tags
+    query.internal_tags = { $in: input.internal_tags } if input.internal_tags
```
eessex: Noticed that vertical and subvertical are in the query schema, should they feature in here as well?

kanaabe: Since the queries on vertical and sub_vertical only allow a single string to be queried (and not an array), we don't need to do anything custom for it. Of course this is subject to change depending on our convo on Friday! This method is for queries that need a little more support than just passing them through to mongo directly.

api/apps/articles/model/schema.coffee
kanaabe: Yep, this is something to discuss. I was told that Vertical and Subverticals won't change often which gives us an opportunity to optimize for query speeds. If we were to validate an objectId instead, the article would also have to fetch a Curation each time the article got rendered. Maybe this won't be so bad with GraphQL though. 

My feeling is either it will be simple like
```
@array().items(@string()).allow('Art', 'Visual Culture', 'Market').default([])
```
where we allow certain vertical strings and make the front end admin match this, or
```
@array().items(
      @object().keys
          id: @objectId()
    ).allow(null).default([])
```
where we denormalize some of the data that makes it easier for display article info without aggregating fetches.

kanaabe:
Closing this in favor of upcoming PR.

PR #973, opened 2017-03-17
@kanaabe => Remove scribe from SectionImage

eessex:
Replaces SectionImage's caption with the new RichTextCaption. 
Updates SectionImage's tests. 
Adds an onChange prop to RichTextCaption to accommodate uses where props shouldn't be set directly. 
(minor, removes unused require statement from SectionImageSet)

image-section-caption


kanaabe:
👏

Issue #976, opened 2017-03-20
more minor link issue: can't preview links in Writer


halleyjohnson:
In the old editor, when you clicked on linked text, the link would pop up so you could see where it links to and then click the link if you want, and edit/remove the link as well.

I'm getting no similar functionality even though links work well on force

Issue #975, opened 2017-03-20
Having trouble adding links in text in the new editor


halleyjohnson:
The popup comes up but when I paste in a link and hit apply, nothing happens.

PR #977, opened 2017-03-20
@kanaabe => Check if link class exists

eessex:
Closes https://github.com/artsy/positron/issues/976
Closes https://github.com/artsy/positron/issues/975
Bugfix for text editor, links couldn't save with className undefined.


kanaabe:
Cool, can we get tests for these?

kanaabe:
👍

Issue #974, opened 2017-03-20
Italics stripped when pasting from Google docs into text component


halleyjohnson:
Found a new text editor bug!

PR #978, opened 2017-03-20
@kanaabe => Pasting from google doesn't strip em

eessex:
Closes https://github.com/artsy/positron/issues/974

Updates #stripCharacterStyles if statement (was not making it through or clause)
Adds explicit span replacement to #stripGoogleStyles
Adds support for span that is bold and italic
Additional tests

google-styles

### DIFF
--- a/client/apps/edit/components/section_text/index.coffee
+++ b/client/apps/edit/components/section_text/index.coffee
@@ -162,12 +162,17 @@ module.exports = React.createClass
     for block, i in boldBlocks
       if block.style.fontWeight is 'normal'
         $(doc.getElementsByTagName('B')[i]).replaceWith(doc.getElementsByTagName('B')[i].innerHTML)
-    # replace bold spans with actual b tags
-    boldSpans = doc.getElementsByTagName('SPAN')
-    for span, i in boldSpans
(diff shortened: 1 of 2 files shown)


client/apps/edit/components/section_text/index.coffee
```diff
@@ -185,7 +190,7 @@ module.exports = React.createClass
     characterList = contentBlock.getCharacterList().map (character) ->
       if keepAllowed
         unless character.hasStyle 'UNDERLINE'
-          return character if character.hasStyle 'BOLD' or character.hasStyle 'ITALIC' or character.hasStyle 'STRIKETHROUGH'
+          return character if character.hasStyle('BOLD') || character.hasStyle('ITALIC') || character.hasStyle('STRIKETHROUGH')
```
kanaabe: Curious do `or`'s not work here?

kanaabe:
Woot thanks for the test and screenshot!

client/apps/edit/components/section_text/index.coffee
eessex: Oh, they probably do -- it was the parens that were buggy

PR #979, opened 2017-03-20
@eessex => Less noisy morgan logs on papertrail

kanaabe:
Use morgan's `tiny` option to log less on papertrail

cc @izakp!

### DIFF
--- a/api/index.coffee
+++ b/api/index.coffee
@@ -14,7 +14,7 @@ app.use cors()
 app.use helpers
 app.use bodyParser.json limit:'5mb', extended: true
 app.use bodyParser.urlencoded limit:'5mb', extended: true
-app.use morgan 'dev'
+app.use morgan(if NODE_ENV is 'development' then 'dev' else 'tiny')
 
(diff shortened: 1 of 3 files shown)


api/index.coffee
```diff
@@ -14,7 +14,7 @@ app.use cors()
 app.use helpers
 app.use bodyParser.json limit:'5mb', extended: true
 app.use bodyParser.urlencoded limit:'5mb', extended: true
-app.use morgan 'dev'
+app.use morgan(if NODE_ENV is 'development' then 'dev' else 'tiny')
```
izakp: <3 thanks @kanaabe!  Outputting necessary logging keeps our syslog servers healthy and happy ^__^

kanaabe:
@izakp changed it to `combined` instead which gives us IP. Then, we can filter heroku instead!

eessex:
@kanaabe this looks good to me, will be sure to read up on morgan

izakp:
Great!

PR #980, opened 2017-03-22
@kanaabe => Lead paragraph converted to Draft

eessex:
Adds rich text paragraph component that accepts an array of allowed styles. 
Replaces lead paragraph with new component.


client/apps/edit/components/layout/index.coffee
```diff
@@ -68,7 +52,7 @@ module.exports = class EditLayout extends Backbone.View
       featured: @$('[name=featured]').is(':checked')
       exclude_google_news: @$('[name=exclude_google_news]').is(':checked')
       title: @$('#edit-title textarea').val()
-      lead_paragraph: @$('#edit-lead-paragraph').html()
+      lead_paragraph: @article.leadParagraph.get('text')
```
kanaabe: Curious, why do we still need to serialize if we're handling updates in the article model?

eessex: Good point, looks like this is extraneous

PR #981, opened 2017-03-22
@eessex => Colorize Morgan Logs

kanaabe:
Uses chalk to colorize morgan logs. We recently switched `morgan` to the `combined` format, but it's a bit difficult to read in Papertrail without colorization. This colorizes the logs using a custom format instead. It easily distinguishes between API/Client calls too.  Keeping `dev` mode for now since the user-agent gets a bit noisy when trying to develop. 

cc @izakp 

image


eessex:
Looks good from here 🌈

api/index.coffee
```diff
@@ -14,7 +15,22 @@ app.use cors()
 app.use helpers
 app.use bodyParser.json limit:'5mb', extended: true
 app.use bodyParser.urlencoded limit:'5mb', extended: true
-app.use morgan(if NODE_ENV is 'development' then 'dev' else 'combined')
+
+app.use morgan if NODE_ENV is 'production' then 'dev' else (tokens, req, res) ->
```
izakp: @kanaabe Nice pattern!  Can we extract this somehow to include across other Express apps using the Morgan logger?

Issue #987, opened 2017-03-23
Missing window.global


kanaabe:
I'd like to see a little more detail on this weird thing we have to do on React components. Is there something cleaner/clearer we can do to ensure that global is present on these components?

eessex:
Not sure why this would be affecting the list view, but the window.global must be declared in all files that require Draft. 

Currently that includes:
/client/apps/edit/components/section_text/index.coffee
/client/components/rich_text_caption/index.coffee
/client/components/rich_text/components/input_paragraph.coffee

As well as the corresponding tests to each.

kanaabe:
Cool, thanks! I'm definitely confused by this still but if it happens on `ArticleList` maybe it's a React (since 15?) thing. Is there a Draft issue/doc that says you need the global?

eessex:
Resolved in a newer version of draft

PR #988, opened 2017-03-23
@eessex => Make window.global available on ArticleList

kanaabe:
For now, adding the same window setter removes the `undefined global` issue, but there's def something about this that doesn't make sense. I made an issue [here](https://github.com/artsy/positron/issues/987) we can look at in our downtime.

### DIFF
--- a/client/apps/edit/components/section_text/index.coffee
+++ b/client/apps/edit/components/section_text/index.coffee
@@ -1,6 +1,5 @@
 React = require 'react'
 ReactDOM = require 'react-dom'
-window.global = window
 sd = require('sharify').data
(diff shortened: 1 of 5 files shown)


kanaabe:
@eessex changed this a bit and moved the global to main.coffee to ensure all client files have it. I still don't completely understand it so I'm keeping that issue open, but at least we won't have to think about requiring it on a per-file basis.

eessex:
Ok seems like a sane solution for the moment.  The only place I could see mention of requiring global was in this thread, where it has to do with creating/converting HTML Entities: https://github.com/facebook/draft-js/issues/55

PR #991, opened 2017-03-26
@eessex => Fix header saving issue and remove dead section code

### LINKED ISSUE #921: Full screen header disappears after publish
Unfortunately don't have a great handle on what exactly causes this but the last three articles we've published with full-screen headers (first one Jan. 14th) have had the header image module disappear sometime after publish. I always check them right when they publish, so it's not the "pushing publish" that does it, but sometime by the morning I notice that it's missing.

kanaabe:
Closes https://github.com/artsy/positron/issues/921

First, this PR removes an old restriction we had where fullscreen headers + sections/verticals were not allowed to exist together. Since Section/Vertical is no longer being used, we don't need this anymore. 

Also, this fixes #921. What's happening is that because we were setting `hero_section: null` whenever there was no hero content [here](https://github.com/artsy/positron/blob/master/client/models/article.coffee#L109), it was deleting the hero section on a Queue save which doesn't rely on pulling the entire article from the DB before saving. To solve, I added a `simple` mode to the article model which doesn't create any associations on a `new Article` which means unnecessary defaults will be ignored.


client/models/article.coffee
```diff
@@ -13,17 +20,18 @@ module.exports = class Article extends Backbone.Model
...
+      @sections = new Sections @get 'sections'
+      @featuredPrimaryArtists = new Artists
+      @featuredArtists = new Artists
+      @mentionedArtists = new Artists
+      @featuredArtworks = new Artworks
+      @mentionedArtworks = new Artworks
+      @leadParagraph = new Backbone.Model text: @get('lead_paragraph')
+      @heroSection = new Backbone.Model @get 'hero_section'
+      @on 'change:hero_section', => @heroSection.set @get 'hero_section'
+      @heroSection.destroy = @heroSection.clear
```
eessex: Would be glad to get a clearer idea of what 'simple' is -- which fields are the unnecessary defaults?

kanaabe: In short, all of them are unnecessary except those which you pass/set manually. By passing a `simple` flag, we don't make any associations on initialize. Happy to go through this in person if you have more questions!

eessex: Yeah would be glad for you to walk me through it

kanaabe:
Merging this for now, I'll explain more during pair session tomorrow!

Issue #989, opened 2017-03-24
Double spaces on Force/Email templates don't appear 


kanaabe:
on email: 
![image](https://cloud.githubusercontent.com/assets/2236794/24315274/86a06c98-10bc-11e7-99b4-9193618368fc.png)

on force: 
![image](https://cloud.githubusercontent.com/assets/2236794/24315287/9898ac26-10bc-11e7-943d-4662205e919c.png)

PR #996, opened 2017-03-29
@kanaabe => Rich text can accept multiple spaces

eessex:
Closes https://github.com/artsy/positron/issues/989
Updates all rich-text components to allow multiple consecutive spaces when exporting html. 
Adds multiple spaces to tests for each component.

### DIFF
--- a/client/apps/edit/components/section_text/index.coffee
+++ b/client/apps/edit/components/section_text/index.coffee
@@ -148,7 +148,10 @@ module.exports = React.createClass
           return span { style: {textDecoration: 'line-through'}}
     })(editorState.getCurrentContent())
     # put the line breaks back for correct client rendering
-    html = html.replace('<p></p>', '<p><br></p>').replace('<p> </p>', '<p><br></p>').replace('  ', ' &nbsp;')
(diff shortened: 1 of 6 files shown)


client/apps/edit/components/section_text/index.coffee
```diff
@@ -148,7 +148,7 @@ module.exports = React.createClass
           return span { style: {textDecoration: 'line-through'}}
     })(editorState.getCurrentContent())
     # put the line breaks back for correct client rendering
-    html = html.replace('<p></p>', '<p><br></p>').replace('<p> </p>', '<p><br></p>')
+    html = html.replace('<p></p>', '<p><br></p>').replace('<p> </p>', '<p><br></p>').replace('  ', ' &nbsp;')
```
kanaabe: would be nice to break this down: http://stackoverflow.com/questions/11195429/coffeescript-and-jquery-chaining

kanaabe:
👍 looks awesome! thanks for tackling

eessex:
@kanaabe yes wise idea, just added.

Issue #942, opened 2017-02-28
App trying to GET /null


kanaabe:
Happens when you load the articles list. Not urgent, but unnecessary.

kanaabe:
Reopening this because I'm still seeing /null fetches in my dev environment.

![image](https://cloud.githubusercontent.com/assets/2236794/26593363/c831c8e2-4531-11e7-95b8-37067530c184.png)

kanaabe:
Fwiw, it seems to be happening when I switch tabs from the Published list to Drafts.

eessex:
Thanks for catching- will check this out

PR #997, opened 2017-03-30
@eessex => Fixes filter zero state

kanaabe:
It's easier to see the actual issue when you have less records, but the empty input state should fetch the entire list of articles when the input is empty instead of ignoring it with [this line](https://github.com/artsy/positron/blob/master/client/components/filter_search/index.coffee#L24). 

filterzero

### DIFF
--- a/api/apps/articles/model/retrieve.coffee
+++ b/api/apps/articles/model/retrieve.coffee
@@ -62,7 +62,7 @@ moment = require 'moment'
     ) if input.artist_id
 
     # Allow regex searching through the q param
-    query.thumbnail_title = { $regex: new RegExp(input.q, 'i') } if input.q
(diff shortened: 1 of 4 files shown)


eessex:
Cool, this makes sense to me - - I had assumed that it was getting the empty articles from this line - https://github.com/artsy/positron/blob/master/client/apps/articles_list/client/client.coffee#L42

eessex:
Does this also solve https://github.com/artsy/positron/issues/942

kanaabe:
The `fetchFeed` function gets called when someone switches between tabs -- this issue is regarding the filter search which needs to still "fetch" (via Bloodhound) even when the input becomes empty. 

Unfortunately it doesn't fix the `null` issue. Will have to dig into that separately!

PR #998, opened 2017-03-31
@kanaabe => HTML space conversion is global

eessex:
Realized I was using replace in a way that only catches the first instance, this changes replace to `/ /g ` global style, which will hopefully fix places in which spacing was converting in unexpected ways.


kanaabe:
👍

PR #999, opened 2017-04-03
@eessex => Tag Admin

kanaabe:
tagadmin
tagadminerror


api/apps/tags/model.coffee
```diff
@@ -0,0 +1,82 @@
...
+@destroy = (id, callback) ->
+  db.tags.remove { _id: ObjectId(id) }, callback
+
+#
+# JSON views
+#
+@present = (tag) =>
+  _.extend
+    id: tag?._id?.toString()
+  , _.omit(tag, '_id')
```
kanaabe: Decided to go with making a proper API endpoint for tags. This uses a `public` field to determine if it's a topic tag or internal tag. The association on the article side is still a question, but for now this makes CRUD-ing tags easier.

client/apps/articles_list/client/client.coffee
```diff
@@ -10,7 +10,7 @@ sd = require('sharify').data
 request = require 'superagent'
 icons = -> require('../icons.jade') arguments...
 
-module.exports = ArticlesListView = React.createClass
+module.exports.ArticlesListView = ArticlesListView = React.createClass
```
kanaabe: Not quite sure why these were removed @eessex but I added them back into some components. This makes testing a bit easier -- 

```
{ ArticlesListView } = mod = require '... 
mod.__set__ 'SomeOtherComponent', sinon.stub()
React.createElement(ArticlesListView, props)
```
That way we can pass in ArticlesListView to the React rendering side AND modify the requires on the file. Lmk if that makes sense.

eessex: Ah cool -- I think they were removed when the displayName/new react naming was added to them, but didn't realize they added this utility.

kanaabe: 👍  nw, also sorry this PR is giant. Might help to look per commit

client/apps/articles_list/client/client.coffee
```diff
@@ -66,9 +66,10 @@ module.exports = ArticlesListView = React.createClass
         FilterSearch {
           url: sd.API_URL + "/articles?published=#{@state.published}&channel_id=#{sd.CURRENT_CHANNEL.id}&q=%QUERY"
           placeholder: 'Search Articles...'
-          articles: @state.articles
+          collection: @state.articles
           searchResults: @setResults
           selected: null
+          contentType: 'article'
         }
```
eessex: Nice that this is becoming cross-type!

client/apps/settings/client/tags.coffee
```diff
@@ -0,0 +1,107 @@
...
+          public: isPublic
+
+  getActiveState: (type) ->
+    if @state.public is type then 'is-active' else ''
+
+  render: ->
+    div {
+      className: 'tags-container'
+      'data-loading': @state.loading
+      ref: 'tagsContainer'
```
eessex: Looks like this is only used for testing, you can locate using the classname too.

client/components/tag_list/test/index.coffee
```diff
@@ -0,0 +1,76 @@
+benv = require 'benv'
+sinon = require 'sinon'
+{ resolve } = require 'path'
+fixtures = require '../../../../test/helpers/fixtures'
+React = require 'react'
+ReactDOM = require 'react-dom'
+ReactTestUtils = require 'react-addons-test-utils'
+ReactDOMServer = require 'react-dom/server'
+r =
+  findAll: ReactTestUtils.scryRenderedDOMComponentsWithClass
```
eessex: scryRenderedDOMComponentsWithClass is an exciting new one for me

client/apps/settings/client/add_tag.coffee
```diff
@@ -0,0 +1,22 @@
+React = require 'react'
+ReactDOM = require 'react-dom'
+{ div, input, button } = React.DOM
+
+module.exports = AddTag = React.createClass
+  displayName: 'TagsView'
+
+  addTag: ->
+    @props.addTag @refs.addTagInput.value
+    @refs.addTagInput.value = ''
```
eessex: I believe using state is the recommended method over refs - this is so concise that it doesn't seem pressing to give a state to this form.
https://facebook.github.io/react/docs/refs-and-the-dom.html#legacy-api-string-refs

eessex:
A few small comments but nothing that would hold up merging, excited to get this in action!

client/apps/settings/client/add_tag.coffee
kanaabe: Good catch! Changes made, will merge when 💚

PR #1002, opened 2017-04-05
A little help with .env.

dblock:
Okay this is not the prettiest, maybe someone can suggest better? This gives you the env variables from positron-staging that you need to replace in `.env`.


kanaabe:
Yes this is helpful to add, thank you!

PR #1003, opened 2017-04-05
@eessex => Refactor AutocompleteList & Select

kanaabe:
I still need to get tests in and work through one last `AutocompleteSelect` bug, but you can start to look at this for the approach. Instead of having a parent `Autocomplete`, this just moves the Bloodhound and Typeahead initialization to the React component and takes out the `Backbone` dependency. There wasn't enough code in my opinion to abstract this out, so it stays relatively silo-ed per component.

List has additional props: `idsToFetch`, `fetchUrl`, `resObject`
Select has: `idToFetch`, `fetchUrl`, `resObject`

Hopefully I'll have this wrapped up later tonight/tomorrow!


eessex:
So exciting, makes so much more sense seeing these self-contained. Do you think we should merge this before the admin panel, and I can refactor to include these changes?

kanaabe:
That's totally up to you since I mostly worked on this to unblock you from tests but if you got those tests working anyway, we can merge yours first and I can rework this to the new format later.

eessex:
Still haven't gotten to test the results of the Autocompletes, only that they were initialized -- so it's probably best to get this in first, and I can build out a few more tests on my side.

PR #1004, opened 2017-04-06
@kanaabe => Working developer setup with default partner ID.

dblock:
Added an unauthorized warning in the logs so it's clear what's going on.

Added `DEFAULT_PARTNER_ID` that is required for an empty local setup as that's the default partner to which to login. It's likely possible to setup a user with channel access, but it seems harder as it's a chicken-and-egg problem of adding IDs into records that don't exist until the first failed login.

It would be nice to document how the whole access thing works and how to add yourself to a channel, too.


README.md
```diff
@@ -22,43 +22,58 @@ Set-Up
...
 ```
 make s
 ```
-- Positron should now be running at [http://localhost:3005/](http://localhost:3005/)
+- Positron should now be running at [http://localhost:3005/](http://localhost:3005/), open a browser and navigate to it. That will redirect you to staging, login as an Artsy administrator and it will redirect you to `http://localhost:3005` logged into Writer with the default partner gallery channel (David Zwirner).
 
 Additional docs
 ---
 
-You can find additional documentation about Positron (deployments etc) in this repository's /doc directory.
```
dblock: There're no deployment docs, so ...

PR #1006, opened 2017-04-06
@kanaabe => Fix #1005: Surface error on save.

### LINKED ISSUE #1005: 403 errors fail silently
Seeing an auto-save article fail silently (and "hang", ie. the saving button is still saving). Server error was `{"status":403,"message":"child \"sections\" fails because [\"sections\" at position 0 does not match any of the allowed types]"}`. Needs to display some kind of error message.

dblock:
This is a fix for #1005. The error surfaces as a hover of an error state and the page doesn't navigate away.

errors

### DIFF
--- a/client/apps/edit/components/header/index.coffee
+++ b/client/apps/edit/components/header/index.coffee
@@ -8,18 +8,28 @@ module.exports = class EditHeader extends Backbone.View
 
   initialize: (options) ->
     { @article } = options
+    @finished = false
     @user = new User sd.USER
     @article.on 'change', @saving
     @article.on 'change', @toggleCheckmarks
-    @article.on 'sync', @doneSaving
+    @article.on 'error', @error
+    @article.on 'sync', @done
     @toggleCheckmarks()
 
-  saving: =>
-    @$('#edit-save').addClass 'is-saving'
-    @$('#edit-save').removeClass 'attention'
+  error: (model, e) =>
+    @finished = false
+    console.error "#{e.status}: #{e.responseText}."
+    @$('#edit-delete').text 'Delete'
+    @$('#edit-save').removeClass('is-saving').text('Save')
+    @$('#edit-error').text(e.responseText).show()
 
-  doneSaving: =>
-    @$('#edit-save').removeClass 'is-saving'
+  saving: =>
+    @$('#edit-error').hide()
+    @$('#edit-save').addClass('is-saving')
+
+  done: =>
+    @$('#edit-save').removeClass('is-saving')
+    @article.trigger('finished') if @finished
 
   toggleCheckmarks: =>
     @$('#edit-tabs a:eq(0)').attr 'data-complete', @article.finishedContent()
@@ -33,13 +43,14 @@ module.exports = class EditHeader extends Backbone.View
     'click #edit-save': 'save'
 
   togglePublished: (e) ->
+    @finished = true
     e.preventDefault()
     e.stopPropagation()
     if @article.finishedContent() and @article.finishedThumbnail()
       @$('#edit-publish').text(
         if @article.get('published') then 'Unpublishing...' else 'Publishing...'
       )
-      @article.trigger('finished').save published: not @article.get 'published'
+      @article.save(published: not @article.get 'published')
     else
       @article.trigger 'missing'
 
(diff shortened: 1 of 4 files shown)


client/apps/edit/components/header/test/index.coffee
```diff
@@ -30,7 +30,7 @@ describe 'EditHeader', ->
 
     it 'indicates saving on change', ->
       @view.article.trigger 'change'
-      @view.$('#edit-save').hasClass('is-saving').should.be.ok
```
dblock: This is incorrect and doesn't fail whether the class is there or not. The check needs to be `equal true/false`.

client/apps/edit/components/header/index.coffee
```diff
@@ -11,15 +11,28 @@ module.exports = class EditHeader extends Backbone.View
...
+    @article.on 'error', @errorSaving
     @article.on 'sync', @doneSaving
     @toggleCheckmarks()
 
+  errorSaving: (model, e) =>
+    message = JSON.parse(e.responseText).message ? e.responseText
+    console.error "Error saving #{model.get('id')}: #{e.status}: #{message}."
+    @$('#edit-save')
+      .text('Error')
+      .attr('title', message)
```
dblock: You can hover over the button and see the error.

eror

client/apps/edit/components/header/test/index.coffee
kanaabe: 👍  these tests are pretty old, thanks for catching.

client/apps/edit/components/header/index.coffee
kanaabe: 👀

client/apps/edit/components/header/test/index.coffee
```diff
@@ -77,3 +77,8 @@ describe 'EditHeader', ->
       @view.save preventDefault: ->
       spy.called.should.be.ok
       Backbone.sync.args[0][0].should.equal 'create'
+
+    xit 'reports the error', ->
+      # todo: simulate a server error
+      @view.$('#edit-save').hasClass('attention').should.be.true()
```
dblock: Right on, that works. Thx.

dblock:
I fixed the navigating away by moving `@article.trigger('finished')` into the successful save. This is ready to merge.

client/apps/edit/components/header/index.coffee
```diff
@@ -11,15 +11,28 @@ module.exports = class EditHeader extends Backbone.View
...
+    @$('#edit-save')
+      .addClass('is-saving')
+      .removeClass('attention')
+      .removeAttr('title')
 
   doneSaving: =>
-    @$('#edit-save').removeClass 'is-saving'
+    @$('#edit-save')
+      .removeClass('is-saving')
+    @article.trigger('finished')
```
kanaabe: Perfect, thanks!

dblock:
Actually hold on. Delete has the same problem. Let me add an error label.

dblock:
Is better now. Needs a careful review, I may have messed up some workflow like publishing an article.

I had to keep a `@finished` state because the article auto-saves and triggers `done` all the time. LMK if there's a cleaner way of dealing with that.

kanaabe:
Code wise this looks good, but this PR raises a problem with Writer's API which is that deletion incorrectly returns a 500 because [this callback](https://github.com/artsy/positron/blob/master/api/apps/articles/model/index.coffee#L110) mistakenly passes an id, whereas the [router thinks](https://github.com/artsy/positron/blob/master/api/apps/articles/routes.coffee#L56) it's an error. Funny I never noticed that because the of the non-blocking redirect so thanks for noticing!

I'll let you know when that's merged @dblock and you can test deletion on your end afterwards. I wonder what problem you're seeing with publishing an article? Requirements for publishing are submitting a thumbnail image on the "Display" panel and a title for the article. That _should_ be it.

dblock:
Publishing works correctly, not sure what that was about. Looking forward to your fix for that other problem @kanaabe. Glad we'll now actually see errors.

That 500 causes a test to spit it out, but there're no failing tests. So hopefully that happens too :)

kanaabe:
@dblock feel free to test it now with 200 deletions in place. If that looks good on your end, feel free to merge this! 👏  Thanks for the great work!

dblock:
This is all good!

PR #1007, opened 2017-04-06
@eessex => Added video caption

dblock:
Added captions to video.

### DIFF
--- a/client/apps/edit/components/section_video/test/index.coffee
+++ b/client/apps/edit/components/section_video/test/index.coffee
@@ -17,18 +17,20 @@ describe 'SectionVideo', ->
   beforeEach (done) ->
     benv.setup =>
       benv.expose $: benv.require('jquery'), resize: ->
+      global.HTMLElement = () => {}
       SectionVideo = benv.require resolve __dirname, '../index'
+      RichTextCaption = benv.requireWithJadeify(resolve(__dirname, '../../../../../components/rich_text_caption/index'), ['icons'])
+      SectionVideo.__set__ 'RichTextCaption', React.createFactory(RichTextCaption)
       SectionVideo.__set__ 'gemup', @gemup = sinon.stub()
       props = {
         section: new Backbone.Model
           url: 'https://www.youtube.com/watch?v=dQw4w9WgXcQ'
+          caption: '<p>Rick Astley - Never Gonna Give You Up</p>'
         editing: false
         setEditing: -> ->
       }
       @rendered = ReactDOMServer.renderToString React.createElement(SectionVideo, props)
       @component = ReactDOM.render React.createElement(SectionVideo, props), (@$el = $ "<div></div>")[0], => setTimeout =>
-        sinon.stub @component, 'setState'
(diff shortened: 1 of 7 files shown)


eessex:
The caption itself is tested in the rich_text_component, but we'd want to update the video section's tests as well to make sure it is rendering/previewing correctly: https://github.com/artsy/positron/blob/master/client/apps/edit/components/section_video/test/index.coffee

The image component does a similar thing with the same component: https://github.com/artsy/positron/blob/master/client/apps/edit/components/section_image/test/index.coffee

client/apps/edit/components/section_video/test/index.coffee
```diff
@@ -22,6 +22,7 @@ describe 'SectionVideo', ->
       props = {
```
eessex: You'll need to require the RichTextComponent with icons to get this test running - - see a working example here: https://github.com/artsy/positron/blob/master/client/apps/edit/components/section_image/test/index.coffee#L29

eessex: Additionally you will notice the [global window](https://github.com/artsy/positron/blob/master/client/apps/edit/components/section_image/test/index.coffee#L24) required just above on ln 24 and 25-- draft.js needs these declared in the test environment to run correctly

eessex:
@kanaabe brought up the concept that maybe this field should be required -- if it is not filled out on articles with more than one video, we will still have the same problem of duplicate sitemap results.

dblock:
Maybe we can make it required in a subsequent iteration? We should first go back and add descriptions to existing videos.

dblock:
I cannot get the component to re-render in this one test:

```js
  it 'previews captions on keyup', ->
    @component.onCaptionChange('<p>updated</p>')
    $(ReactDOM.findDOMNode(@component)).html().should.containEql '<p>updated</p>'
```

But it's not a blocker.

client/apps/edit/components/section_video/test/index.coffee
```diff
@@ -17,18 +17,20 @@ describe 'SectionVideo', ->
...
         section: new Backbone.Model
           url: 'https://www.youtube.com/watch?v=dQw4w9WgXcQ'
+          caption: '<p>Rick Astley - Never Gonna Give You Up</p>'
         editing: false
         setEditing: -> ->
       }
       @rendered = ReactDOMServer.renderToString React.createElement(SectionVideo, props)
       @component = ReactDOM.render React.createElement(SectionVideo, props), (@$el = $ "<div></div>")[0], => setTimeout =>
-        sinon.stub @component, 'setState'
```
dblock: This was 99% of my test problems, duh.

dblock:
This is all good to go!

client/apps/edit/components/section_video/test/index.coffee
eessex: Ah of course ¯\_(ツ)_/¯

eessex:
Looks good, thanks for adding this! I'll take a look at requiring this field today.

PR #1009, opened 2017-04-06
@eessex => Handle destroy error

kanaabe:
cc @dblock


eessex:
This looks good to me!

PR #1012, opened 2017-04-10
@kanaabe => Verticals api

eessex:


eessex:
@kanaabe, wanted to check if articles need to include verticals in their querySchema here - https://github.com/artsy/positron/pull/1012/files#diff-a762aa233b91e6790b001410d5769810L193

api/apps/articles/model/retrieve.coffee
```diff
@@ -54,6 +54,11 @@ moment = require 'moment'
       { contributing_authors: { $elemMatch: { id: ObjectId input.all_by_author} } }
     ) if input.all_by_author
 
+    # Convert query for articles by vertical
+    query.$or.push(
+      { vertical: { $elemMatch: { id: ObjectId input.vertical} } }
+    ) if input.vertical
+
```
kanaabe: This doesn't need the $or since we're not querying for more than one thing -- should suffice to just say 
```
query.vertical = { $elemMatch: { id: ObjectId input.vertical} }  if input.vertical
```

api/apps/articles/model/retrieve.coffee
```diff
@@ -40,7 +40,7 @@ moment = require 'moment'
 
     # Only add the $or array for queries that require it (blank $or array causes problems)
     query.$or ?= [] if input.artist_id or input.all_by_author or
-      input.has_video or input.channel_id
+      input.has_video or input.channel_id or input.vertical
```
kanaabe: That means this like could also use an update!

api/apps/articles/model/schema.coffee
```diff
@@ -77,6 +77,9 @@ denormalizedArtwork = (->
   thumbnail_teaser: @string().allow('', null)
   thumbnail_image: @string().allow('', null)
   tags: @array().items(@string()).default([])
+  vertical: @object().keys
+    name: @string().allow('').default('')
```
kanaabe: Sorry one last thing, do we need a default here? If we open up and resave an old article, it's going to have a `vertical: name: ''` in the article model and I think that'd be confusing without the id in there as well. Maybe it'd be more clear if we let the object be `null` which means no vertical field will be saved in the first place.

kanaabe:
This looks good! Thanks for adding. Once we address that last comment, feel free to merge.

PR #1014, opened 2017-04-11
[WIP] Apple news format

kanaabe:


kanaabe:
Closing this PR since it's pretty inactive, but we can continue to work on the branch.

Aside from that, I'm not totally convinced that we should be spending energy towards this after looking at their analytics docs (or lack thereof). We can't track pageviews from it and to make it work we need to add an additional field on `Article` to support unique Apple IDs on each article.

It seems like the way people get pageviews is by having people click "Read Full Article" which leads people to the entire site. From there you can distinguish between Apple News by checking the `user-agent` but it seems like they're [having trouble](https://ctrl.blog/entry/applenews-user-agent) getting a consistent user-agent that everyone can use. Also this user-agent is available on iOS10 only.

At the least, I think it's worth more research + data to see how it can benefit us. Maybe it's a brand thing. We can talk about it in the next #publishing meeting, @owendodd

eessex:
Have been spending time looking at this off and on but haven't PRd anything yet -- several components are now printing, but the styling has a long way to go.  

Layout/inline styles are pretty counterintuitive and have been a longer learning curve than I expected, mostly for the lack of examples covering our use cases.  It is moving along though, so I'm happy to return to this later if everyone agrees it is a good idea.

PR #1018, opened 2017-04-13
@kanaabe => tags bug

eessex:
Forgot to remove dead code for tags from the serialize function, which was overwriting them as empty when saving a published article.


eessex:
Wasn't sure how to add a test for this as the tag creation is already tested throughout, the only thing i could think of was testing that serialize doesn't overwrite fields it doesn't cover -- does this seem worth adding?

PR #1019, opened 2017-04-13
@izakp => Adds artsy-morgan

kanaabe:
✔️ 

Did want to add a Client/API distinction in the log output but I'm going to table that now and make an issue in artsy-morgan :)


izakp:
Oooo yea how does this work client-side?

kanaabe:
Ah well by client, i mean the backend of writer.artsy.net, so it's similar to Force @izakp

PR #1020, opened 2017-04-14
@kanaabe => Tier/featured bugs

eessex:
The article schema has default values for unset tier and featured fields, which were not being displayed in the UI. State is added for these fields, and a test to be sure the default values are displayed.

### DIFF
--- a/client/apps/edit/components/admin/article/index.coffee
+++ b/client/apps/edit/components/admin/article/index.coffee
@@ -13,8 +13,8 @@ module.exports = AdminArticle = React.createClass
     publish_date: moment().format('YYYY-MM-DD')
     publish_time: moment().format('HH:mm')
     focus_date: false
-    tier: @props.article.get('tier') || 2
-    featured: @props.article?.get('featured') || false
+    tier: @props.article.get('tier') or 2
+    featured: @props.article?.get('featured') or false
 
   componentWillMount: ->
     @setupPublishDate()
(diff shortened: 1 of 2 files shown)


client/apps/edit/components/admin/article/index.coffee
```diff
@@ -167,6 +171,8 @@ module.exports = AdminArticle = React.createClass
               input {
                 type: 'checkbox'
                 checked: @props.article.get 'exclude_google_news'
+                value: @props.article.get 'exclude_google_news'
+                readOnly: true
```
kanaabe: curious why is this readOnly?

client/apps/edit/components/admin/article/index.coffee
```diff
@@ -13,6 +13,8 @@ module.exports = AdminArticle = React.createClass
     publish_date: moment().format('YYYY-MM-DD')
     publish_time: moment().format('HH:mm')
     focus_date: false
+    tier: @props.article.get('tier') || 2
+    featured: @props.article?.get('featured') || false
```
kanaabe: i generally try to use coffeescript's `or` in places like this but not a blocker

client/apps/edit/components/admin/article/index.coffee
eessex: React was surfacing errors about values set on uncontrolled components (because the onChange event for our styled checkbox is in the parent div).

kanaabe: ok, i think that makes sense!

Issue #1011, opened 2017-04-07
Don't allow line breaks in titles


kanaabe:
https://github.com/artsy/force/issues/1195

PR #1021, opened 2017-04-14
@kanaabe => Don't allow linebreaks in titles

eessex:
Closes https://github.com/artsy/positron/issues/1011

- Strips linebreaks from titles on save
- Prevents use of enter key in title textarea


kanaabe:
👏  Thank you!!

PR #1022, opened 2017-04-17
@eessex => Fix autocomplete list

kanaabe:
Paired with @orta on this one!

Channel page's AutocompleteList was breaking because we were resetting the first `module.export` where we call `React.render`. 

This moves the `React.render` call to the Channel page since it's the only non-React place that uses this.


eessex:
Cool, in what way was it breaking?  Haven't had any trouble changing channels.

kanaabe:
It's the autocomplete on the channels settings page, not the switching

eessex:
Aha - good to know!

PR #1023, opened 2017-04-17
@kanaabe => Venice backend

eessex:
venice-backend


client/apps/settings/client/curations/venice_admin.coffee
```diff
@@ -0,0 +1,98 @@
...
+    @state.curation.set 'sections', newSections
+
+  onInputChange: (e) ->
+    @setState isChanged: true
+    @state.curation.set 'description', e.target.value
+
+  save: ->
+    @setState isSaving: true
+    @state.curation.save()
+    _.defer => @doneSaving()
```
eessex: It looks like curation.save() has callbacks for done and error, but I couldn't manage to hook into these correctly, and ended up with defer instead.

kanaabe: Hmm yeah this does seem a little odd since `save` doesn't really have anything to do with `defer` which just [waits for the current call stack](http://underscorejs.org/#defer) to clear. If you inspect @state.curation at this point, does it appear to be a Backbone model with attributes?

```
@state.curation.save
   success: ->
   error: ->
```
^ that should be the right way to hook into these callbacks.

client/apps/settings/client/curations/venice_admin.coffee
```diff
@@ -0,0 +1,98 @@
...
+
+  doneSaving: ->
+    setTimeout( =>
+      @setState isSaving: false, isChanged: false
+    , 750)
```
kanaabe: Hmm what's this timeout for?

client/apps/settings/client/curations/venice_section.coffee
```diff
@@ -0,0 +1,114 @@
...
+            }
+          div { className: 'field-group' },
+            label {},'Video URL'
+            input {
+              className: 'bordered-input'
+              placeholder: 'Enter a link'
+              defaultValue: @state.video_url
+              onChange: @onInputChange
+              name: 'video_url'
+            }
```
kanaabe: I'm sure there will be more updates to other sections, but I'll just note anyway that 
 there's a high chance this field will need to accept multiple files so we can optimize the video for device/bandwidth. Def not a blocker atm tho

kanaabe:
👏  So quick! Looks like you had some problems with the Curation save callbacks but it would be good to get that working. 

It looks like there are some real errors too but once those are passing and the save is in place, feel free to merge!

client/apps/settings/client/curations/venice_section.coffee
eessex: Sounds good, I can add this at a later date -- do we know how many versions or for what devices we will need them yet?

eessex: Also realized we will want to add the video length as well.

PR #1025, opened 2017-04-19
@eessex => Adds subtitle to venice curation

kanaabe:
Supports "Christian Marclay" as a subtitle.

image

image


eessex:
Just checked in w @owendodd  on this -- the 'Inside the Biennale' title appears on every page, so we only need one title field.

kanaabe:
Ah gotcha! Thanks for checking

PR #1028, opened 2017-04-24
@kanaabe => Related artists format

eessex:
To allow us to print a follow button without fetching the name, saves related artists as an object including verbose id and name, rather than mongo _id.


kanaabe:
👏  great idea!

PR #1030, opened 2017-04-28
@kanaabe => Venice artist mongo ids

eessex:
Analytics wants a mongo id for artist-follow, so adding this back into the mix


kanaabe:
fyi this build is failing for some reason: https://semaphoreci.com/artsy-it/positron/branches/master/builds/553

PR #1037, opened 2017-05-16
@craigspaeth => Airtable <> GA Data Copy Script

kanaabe:
This takes a bunch of Google Analytics data that gets reported in a [Google Sheet](https://docs.google.com/spreadsheets/d/1SDNxlERn9Z4zqRQ7RH50yVjiKe9hQqinm5D25AwwuSA/edit#gid=1936673791) and transfers it to Airtable in a sheet called ["Archive"](https://airtable.com/tblsbZtO1JK9bn6E2/viwDUqRX1P5HN01qa).

There are ~2000 records to be updated, and of course this number is expected to grow slowly, probably by 5 records a day. The entire thing takes about 7 minutes to complete.

The idea is to run this script on Heroku Scheduler once every 30 minutes. The ultimate goal is to provide editorial with real-ish time analytics on their articles in the comfort of their daily Airtable workflow, move away from using Google Sheets for reporting, and provide context to Writers on their past performance.

I know it's a pretty heavy script so I'm open to ideas about cleaning this up or if there are better paths.

### DIFF
--- a/scripts/ga_airtable_transfer.coffee
+++ b/scripts/ga_airtable_transfer.coffee
@@ -64,7 +64,6 @@ fetches.push (cb) ->
         airtableFetches = getAirtableFetches(rows)
         async.parallel airtableFetches, (err, res) ->
           return debug err if err
-          console.log "Completed creating/updating " + (offset + 5) + " articles."
           clb()
 
   console.log 'Starting fetches...'
@@ -82,7 +81,7 @@ getAirtableFetches = (rows) ->
   _.map rows, (row) ->
     (airtableCb) ->
       landingPage = row.galandingpagepath
-      return airtableCb() unless landingpage
+      return airtableCb() unless landingPage
       base('Archive').select
         filterByFormula: "({Name} = 'https://#{landingPage}')"
         view: 'Archive List'
@@ -96,7 +95,6 @@ getAirtableFetches = (rows) ->
             if err
               debug err if err
               return airtableCb()
-            console.log 'Created record: ' + record.fields["Name"]
             airtableCb()
         else
           return airtableCb() unless records[0]?.getId()
@@ -105,7 +103,6 @@ getAirtableFetches = (rows) ->
             if err
               debug err if err
               return airtableCb()
-            console.log 'Updated record: ' + record.fields["Name"]
             airtableCb()
 
 # Normalize data given a GS row
--- a/scripts/ga_airtable_transfer.coffee
+++ b/scripts/ga_airtable_transfer.coffee
@@ -81,7 +81,7 @@ getAirtableFetches = (rows) ->
   _.map rows, (row) ->
     (airtableCb) ->
       landingPage = row.galandingpagepath
-      return airtableCb() unless landingPage
+      return airtableCb() unless landingPage?.length
       base('Archive').select
         filterByFormula: "({Name} = 'https://#{landingPage}')"
         view: 'Archive List'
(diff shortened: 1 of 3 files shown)


scripts/ga_airtable_transfer.coffee
```diff
@@ -0,0 +1,114 @@
...
+# Authenticate with Google Sheets
+fetches.push (cb) ->
+  doc.useServiceAccountAuth creds, (err, res) ->
+    console.log 'Completed Authentication.'
+    cb()
+
+# Get some sheet info
+fetches.push (cb) ->
+  doc.getInfo (err, info) ->
+    @totalRows = info.worksheets[1].rowCount
```
craigspaeth: I'd be careful about binding to scope here as it's hard to keep track of `this` in anonymous functions like this one. It might be better to hoist `totalsRows` at the top along with `creds`, `fetches`, and friends and assign it here, e.g.

````coffeescript
fetches = []
totalRows = null
#...
fetches.push (cb) ->
  doc.getInfo (err, info) ->
    totalRows = info.worksheets[1].rowCount
````

Also be careful to handle `err` here.

File: scripts/ga_airtable_transfer.coffee [hunk 2]
```diff
@@ -0,0 +1,114 @@
...
+          clb()
+
+  console.log 'Starting fetches...'
+  console.log 'Created ' + dataCopyTasks.length + ' buckets'
+  async.series dataCopyTasks, (err, res) ->
+    console.log 'Completed all buckets'
+    cb()
+
+# Make fetches happen
+async.series fetches, (err, res) ->
```
craigspaeth: Nice building up of `fetches` and using `async` here.

scripts/ga_airtable_transfer.coffee
```diff
@@ -0,0 +1,114 @@
...
+getAirtableFetches = (rows) ->
+  _.map rows, (row) ->
+    (airtableCb) ->
+      landingPage = row.galandingpagepath
+      return airtableCb() unless landingPage?.length
+      base('Archive').select
+        filterByFormula: "({Name} = 'https://#{landingPage}')"
+        view: 'Archive List'
+      .firstPage (err, records) ->
+        if err
```
craigspaeth: It'd probably be good to extract this repeated block of `if err` code into a common function.

scripts/ga_airtable_transfer.coffee
```diff
@@ -0,0 +1,114 @@
...
+          return airtableCb() unless records[0]?.getId()
+          record = records[0].getId()
+          base('Archive').update record, rowValues(row, false), (err, record) ->
+            if err
+              debug err if err
+              return airtableCb()
+            airtableCb()
+
+# Normalize data given a GS row
+rowValues = (row) ->
```
craigspaeth: 👍  for extracting this.

craigspaeth:
I think this is a solid plan and the code is good to merge from my perspective.

Aside from the small comments that I left inline—if you wanted to explore ways to clean this up it might be a good opportunity to try writing this using plain old JS, and promises + async/await (which should be supported natively in the latest versions of Node 7 🎉 ). That could help clean up some of the `if err` error handling logic and avoid the need for the async library altogether. The callback code you've got going on here looks 👌  though, so this is just an opportunity to get comfortable with that approach if you want.

Feel free to address my comments if you want and merge at will. Good stuff!

scripts/ga_airtable_transfer.coffee
kanaabe: 👍

kanaabe:
Ooh that's a great idea. I'm definitely for refactoring using `async/await`

PR #1038, opened 2017-05-17
@kanaabe => Vertical/Tags admin

eessex:
- Finalizes component for new verticals/tags/tracking_tags
      * this component is not yet used, we still need to turn it on after the backfill is finished!
- Adds 'inline' option to AutocompleteList
- Adds `tracking_tags` to article model
- Adds come logic to AutocompleteList to avoid duplicate entries

tag-admin

### DIFF
--- a/api/apps/articles/model/schema.coffee
+++ b/api/apps/articles/model/schema.coffee
@@ -222,6 +222,7 @@ denormalizedArtwork = (->
   q: @string().allow('')
   all_by_author: @objectId()
   tags: @array().items(@string())
+  tracking_tags: @array().items(@string())
   is_super_article: @boolean()
   biography_for_artist_id: @objectId()
   layout: @string()
--- a/api/apps/articles/model/schema.coffee
+++ b/api/apps/articles/model/schema.coffee
@@ -77,6 +77,7 @@ denormalizedArtwork = (->
   thumbnail_teaser: @string().allow('', null)
   thumbnail_image: @string().allow('', null)
   tags: @array().items(@string()).default([])
+  tracking_tags: @array().items(@string()).default([])
   vertical: @object().keys(
     name: @string()
     id: @objectId()
--- a/api/apps/articles/test/model/index.coffee
+++ b/api/apps/articles/test/model/index.coffee
@@ -332,6 +332,39 @@ describe 'Article', ->
           results[2].title.should.equal 'Hello Wuuuurld - Food'
           done()
 
+    it 'can find articles by tracking_tags', (done) ->
+      fabricate 'articles', [
+        {
+          author_id: ObjectId '4d8cd73191a5c50ce220002a'
+          tracking_tags: [ 'video', 'evergreen' ]
+          title: 'The Shanghai Art Project That’s Working to Save Us from a Dystopian Future'
+        }
+        {
+          author_id: ObjectId '4d8cd73191a5c50ce220002b'
+          tracking_tags: [ 'video', 'evergreen' ]
+          title: '$448 Million Christie’s Post-War and Contemporary Sale Led by Bacon and Twombly'
+        }
+        {
+          author_id: ObjectId '4d8cd73191a5c50ce220002c'
+          tracking_tags: [ 'podcast', 'evergreen' ]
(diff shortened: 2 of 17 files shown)


client/components/autocomplete/index.styl
```diff
@@ -63,3 +63,45 @@
...
+      background-image url(/icons/remove_r.svg)
+  .twitter-typeahead
+    width initial
+    flex 1
+    position static !important
+    input
+      background none transparent
+      margin 0
+  .tt-dropdown-menu
+    top calc(100% \+ 2px) !important
```
eessex: Was trying to overwrite inline styles from typeahead here -- not sure if there's a way around important

client/apps/edit/components/admin/verticals_tags/index2.coffee
```diff
@@ -1,51 +1,97 @@
...
+AutocompleteList = React.createFactory require '../../../../../components/autocomplete_list/index.coffee'
 { div, section, label, button, input } = React.DOM
 
 module.exports = React.createClass
   displayName: 'AdminVerticalsTags'
 
   getInitialState: ->
-    vertical: null
-    verticals: ['Art', 'Art Market', 'Culture', 'Creativity']
+    vertical: null || @props.article.get('vertical')
```
kanaabe: `or`

client/components/autocomplete/index.styl
kanaabe: Usually `!important` can be avoided by more specificity in your selector. Check out how `top` is set in `.tt-dropdown-menu` on the typeahead side and see if you can use selectors that are more specific than that. Obviously if they're using `!important` this might be unavoidable.

client/components/autocomplete_list/index.coffee
```diff
@@ -58,33 +59,51 @@ module.exports = AutocompleteList = React.createClass
...
-        div { className: 'autocomplete-selected'},
-          for item in @state.items
-            div { className: 'autocomplete-select-selected', key: item.value }, item.value,
-              input { type: 'hidden', value: item.id, name: @props.name }
-              button { className: 'remove-button', onClick: @removeItem(item) }
+        if @props.inline
+          @printItems()
+        else
+          div { className: 'autocomplete-selected'},
+            @printItems()
```
kanaabe: Cool!

api/apps/articles/model/schema.coffee
```diff
@@ -77,6 +77,7 @@ denormalizedArtwork = (->
   thumbnail_teaser: @string().allow('', null)
   thumbnail_image: @string().allow('', null)
   tags: @array().items(@string()).default([])
+  tracking_tags: @array().items(@string()).default([])
```
kanaabe: Nice! Do you mind adding tests on the API side? We have places for both #save and #query which will be useful to have.

kanaabe:
Last thing, I think we also need to make sure the `tracking_tags` and `vertical` are being appended to the Sailthru distribution [here](https://github.com/artsy/positron/blob/master/api/apps/articles/model/distribute.coffee#L73).

eessex:
@kanaabe do we want to concat these into the tags field for sailtru, or send them as separate fields

kanaabe:
@eessex same field called `tags`.

client/components/autocomplete/index.styl
eessex: The styles are set inline with javascript, so I don't think they can be overwritten http://stackoverflow.com/questions/16234137/how-to-override-inline-css-without-using-important-or-javascript

api/apps/articles/test/model/index.coffee
```diff
@@ -991,6 +1026,15 @@ describe 'Article', ->
...
 
+    it 'saves tracking_tags', (done) ->
+      Article.save {
+        author_id: '5086df098523e60002000018'
+        tracking_tags: ['evergreen', 'video']
+      }, 'foo', (err, article) ->
+        return done err if err
+        article.tracking_tags.should.eql ['evergreen', 'video']
+        done()
+
```
kanaabe: Nice tests! Thanks for the add.

kanaabe:
Woohoo! 👏

PR #1044, opened 2017-05-18
@craigspaeth => Airtable <> GA Script (ES7 Refactor)

kanaabe:
This takes a bunch of Google Analytics data that gets reported in a [Google Sheet](https://docs.google.com/spreadsheets/d/1SDNxlERn9Z4zqRQ7RH50yVjiKe9hQqinm5D25AwwuSA/edit#gid=1936673791) and transfers it to Airtable in a sheet called [Archive](https://airtable.com/tblsbZtO1JK9bn6E2/viwDUqRX1P5HN01qa).

There are ~2200 records to be updated, and of course this number is expected to grow slowly, probably by 5 records a day. Unfortunately, this takes about 20 minutes to complete with the rate limitation. Looks like we can handle that though with a combo of Heroku Scheduler + [workers](https://devcenter.heroku.com/articles/scheduler#long-running-jobs)

The idea is to run this script once every hour. The ultimate goal is to provide editorial with real-ish time analytics on their articles in the comfort of their daily Airtable workflow, move away from using Google Sheets for reporting, and provide context to Writers on their past performance.


package.json
```diff
@@ -83,6 +85,10 @@
     "browserify-dev-middleware": "^0.1.2",
     "caching-coffeeify": "^0.5.1",
     "cssify": "^0.7.0",
+    "eslint": "^3.19.0",
+    "eslint-config-standard": "^10.2.1",
+    "eslint-plugin-promise": "^3.5.0",
+    "eslint-plugin-standard": "^3.0.1",
```
craigspaeth: 👍

scripts/ga_airtable_transfer.js
```diff
@@ -0,0 +1,149 @@
...
+    await authenticateGoogle()
+    console.log('Completed Authentication')
+
+    // Google Sheet Info
+    rowCount = await docInfo()
+    console.log('Total Rows: ' + rowCount)
+    var rowBuckets = Math.ceil(rowCount / 5)
+
+    // Copy Data Tasks
+    for (var i = 0; i <= rowBuckets; i++) {
```
craigspaeth: Might be nicer to use `forEach` here. e.g.

```javascript
rowBuckets.forEach(async (_, i) => await copyDataTask(i * 5))
```

Or, if reasonable, you could even do `copyDataTask` in parallel with `map` e.g.

```javascript
const copyTasks = rowBuckets.map((_, i) => copyDataTask(i * 5))
await Promise.all(copyTasks)
```

scripts/ga_airtable_transfer.js
```diff
@@ -0,0 +1,149 @@
...
+    let promises = rows.map((row) => airtableFetch(row))
+    await Promise.all(promises)
+  } catch (err) {
+    console.log(err)
+  }
+}
+
+async function airtableFetch (row) {
+  try {
+    const getRecord = () =>
```
craigspaeth: Might help clean things up to move these functions outside `airtableFetch` seeing as they don't depend on anything in the function scope like `row`.

craigspaeth:
Nice work refactoring to async/await! 👏  LGTM! I just left some minor style comments that you can feel free to address later or not. Excited to see some ES201X make its way into Positron!

PR #1045, opened 2017-05-18
@eessex: Updated bucket-assets module

craigspaeth:
This updates the bucket-assets module to avoid that getting stuck in S3 timeout issue we saw before.


eessex:
Excellent!

PR #1047, opened 2017-05-19
Revert "@eessex: Updated bucket-assets module"

kanaabe:
Reverts artsy/positron#1045

@craigspaeth sorry it looks like this is breaking the client. Probably something config-related/really simple, but I'm going to revert this for now since we have a time-sensitive backfill that needs to happen today. I'll take a look again after.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,3 @@ dump.rdb
 .env
 .env.ignore
 node_modules
-manifest.json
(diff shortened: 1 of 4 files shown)


craigspaeth:
No problem @kanaabe lmk if you want to pair on putting it back it—it's not time sensitive but just something that [could surprise us](https://github.com/artsy/post_mortems/issues/10) in the rare case when S3 is unavailable.

PR #1052, opened 2017-05-19
@craigspaeth => Send content to sailthru bug

eessex:
This checkbox was not doing anything when clicked, added a change event to the input.


craigspaeth:
Easy peasy, thanks!

PR #1055, opened 2017-05-22
@eessex => [Venice] Social fields

kanaabe:
image

### DIFF
--- a/client/apps/edit/components/admin/index.coffee
+++ b/client/apps/edit/components/admin/index.coffee
@@ -48,28 +48,77 @@ module.exports = React.createClass
   render: ->
     div { className: 'edit-admin' },
 
-      section { className: 'edit-admin--verticals-tags' + @getActiveSection 'verticals-tags' },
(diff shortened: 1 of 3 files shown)


eessex:
Looks good!

client/apps/edit/components/admin/index.coffee
```diff
@@ -48,28 +48,77 @@ module.exports = React.createClass
...
+          onClick: @setActiveSection
+        }
         if @isActiveSection 'additional-appearances'
-          Appearances {article: @props.article, channel: @props.channel, onChange: @onChange}
\ No newline at end of file
+          Appearances {
+            article: @props.article
+            channel: @props.channel
+            onChange: @onChange
+          }
```
kanaabe: This file is irrelevant to the PR, but I just noticed most of the lines here were longer than 100 chars and generally it's a bit easier to read when they are broken out to <100.

In Sublime, there's an option called Ruler which is nice to turn on as a guide (note the faint vertical line):
image

PR #1056, opened 2017-05-22
@eessex => Strip author fields

kanaabe:
[This PR](https://github.com/artsy/positron/pull/1051) doesn't handle the case when `facebook_handle` and `twitter_handle` only exist.

### DIFF
--- a/client/apps/edit/client.coffee
+++ b/client/apps/edit/client.coffee
@@ -121,9 +121,4 @@ convertArtworks = (section, callback) ->
   )
 
 convertAuthor = (article) ->
-  if article.get('author')?.profile_id or article.get('author')?.profile_handle
-    author = {
-      id: article.get('author').id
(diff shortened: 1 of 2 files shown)


client/apps/edit/client.coffee
```diff
@@ -121,9 +121,4 @@ convertArtworks = (section, callback) ->
   )
 
 convertAuthor = (article) ->
-  if article.get('author')?.profile_id or article.get('author')?.profile_handle
-    author = {
-      id: article.get('author').id
-      name: article.get('author').name
-    }
-    article.set 'author', author
+  article.set 'author', _.pick article.get('author'), 'id', 'name'
```
kanaabe: Note we can simplify this by using http://underscorejs.org/#pick

eessex: Ah thanks for catching this edge case.

Issue #1053, opened 2017-05-22
Tagging Admin QA


kanaabe:
- [x] Be able to unselect a vertical (verticals are not required)
- [x] Vertical should not appear in non-editorial channels
- [x] Tagging section for non-editorial should have old-style string inputs instead of autocomplete
- [x] https://writer.artsy.net/api/articles/591ee590137e36001740bec1 is picking up "Market Analysis" as a tag instead of treating them as separate tags

PR #1057, opened 2017-05-23
@kanaabe => Tagging admin QA

eessex:
Closes https://github.com/artsy/positron/issues/1053
- Adds tags-only component for non-editorial admin panel
- Allows verticals to be unset
- Adds safe mode to tag query api for exact name match
- Kills problem of tags surfacing with partial name match
- Does not remove tags that are not in DB

### DIFF
--- a/api/apps/tags/model.coffee
+++ b/api/apps/tags/model.coffee
@@ -26,7 +26,7 @@ Joi.objectId = require('joi-objectid') Joi
   offset: @number()
   count: @boolean().default(false)
   public: @boolean()
-  safe: @boolean().default(false)
+  strict: @boolean().default(false)
 ).call Joi
 
 #
@@ -39,8 +39,8 @@ Joi.objectId = require('joi-objectid') Joi
 @where = (input, callback) ->
   Joi.validate input, @querySchema, (err, input) =>
     return callback err if err
-    query = _.omit input, 'q', 'limit', 'offset', 'count', 'safe'
-    if input.safe
+    query = _.omit input, 'q', 'limit', 'offset', 'count', 'strict'
+    if input.strict
       query.name = { $eq: input.q } if input.q and input.q.length
     else
       query.name = { $regex: ///#{input.q}///i } if input.q and input.q.length
--- a/api/apps/tags/model.coffee
+++ b/api/apps/tags/model.coffee
@@ -26,6 +26,7 @@ Joi.objectId = require('joi-objectid') Joi
   offset: @number()
   count: @boolean().default(false)
   public: @boolean()
+  safe: @boolean().default(false)
 ).call Joi
 
 #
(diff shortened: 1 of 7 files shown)


client/apps/edit/components/admin/index.coffee
```diff
@@ -47,20 +48,24 @@ module.exports = React.createClass
 
   render: ->
     div { className: 'edit-admin' },
-
       section {
         className: 'edit-admin--verticals-tags' + @getActiveSection 'verticals-tags'
       },
         printTitle {
-          section: 'Verticals & Tagging'
+          section: if sd.CURRENT_CHANNEL.type is 'editorial' then 'Verticals & Tagging' else 'Tagging'
```
kanaabe: We should use [this model function](https://github.com/artsy/positron/blob/master/client/models/channel.coffee#L15) to check if it is editorial. If we ever need to make a global change related to this, it's easier to handle the business logic through the model instead of looking for instances of `is 'editorial'`

client/apps/edit/components/admin/test/verticals_tags.coffee
```diff
@@ -96,3 +96,65 @@ describe 'AdminVerticalsTags', ->
...
+        $: benv.require 'jquery'
+        _: benv.require 'underscore'
+        _s: benv.require 'underscore.string'
+      window.jQuery = $
+      @AdminTags = benv.require resolve __dirname, '../verticals_tags/index.coffee'
+      @AdminTags.__set__ 'sd', {}
+      @article = new Article
+      @article.attributes = fixtures().articles
+      @article.set 'tags', ['Artists']
+      @article.set 'tracking_tags', ['op-ed', 'profile']
```
kanaabe: Backbone lets us assign attributes at initialization. I think it makes a little more sense to do something along the lines of:
```
@article = new Article _.extend fixtures().articles,
   tags: ['Artists']
   tracking_tags: ['op-ed', 'profile']
```

client/apps/edit/components/admin/index.coffee
eessex: yes this looks better, thanks!

api/apps/tags/model.coffee
```diff
@@ -26,6 +26,7 @@ Joi.objectId = require('joi-objectid') Joi
   offset: @number()
   count: @boolean().default(false)
   public: @boolean()
+  safe: @boolean().default(false)
```
kanaabe: Curious what the naming of `safe` comes from? Is that a common term for this functionality? As a naive person, my guess of what `safe` might do isn't really related to string matching. I think when we discussed I mentioned maybe using `strict`. `exact` could be clearer too? Not a blocker, but I'd like to hear your thoughts. Hardest thing in CS is naming 😸

eessex: i think i just misremembered - will change it to strict

kanaabe:
🙌  Nice!

PR #1060, opened 2017-05-24
@orta => [Danger] Adds Danger and checks for Assignee

kanaabe:
This should be enough to get Danger working in Semaphore but I'll wait on https://github.com/danger/danger-js/issues/261 to make sure I can run locally.


orta:
Everything looks good - assuming CI which doesn't keep it's process in git 👍

PR #1066, opened 2017-05-26
@kanaabe => [WIP] Channels have public flag

eessex:
Was looking at ways to exclude by 'type' for sitemaps, but I think going with a public flag makes the most sense.  Seems like there may be reason to exclude other channels from the sitemaps. 

Alternatively, would it make sense to tell Cinder only to index channels with slugs, rather than making a new flag?

### DIFF
--- a/api/apps/channels/model.coffee
+++ b/api/apps/channels/model.coffee
@@ -30,6 +30,7 @@ request = require 'superagent'
     text: @string().allow('',null)
   ]).allow(null).default([])
   slug: @string().allow('',null)
+  public: @boolean().default(true)
   pinned_articles: @array().max(6).items(
     @object().keys
       index: @number()
@@ -41,6 +42,7 @@ request = require 'superagent'
   limit: @number().max(Number API_MAX).default(Number API_PAGE_SIZE)
   offset: @number()
   user_id: @objectId()
+  public: @boolean()
   q: @string()
 ).call Joi
 
--- a/api/apps/channels/test/model.coffee
+++ b/api/apps/channels/test/model.coffee
@@ -3,6 +3,7 @@ moment = require 'moment'
 { db, fabricate, empty, fixtures } = require '../../../test/helpers/db'
 Channel = require '../model'
 { ObjectId } = require 'mongojs'
+request = require 'superagent'
 
 describe 'Channel', ->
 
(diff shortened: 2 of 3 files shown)


kanaabe:
Gotcha, I think one reason that `public` can be confusing in this context is we don't know where `public` is used and what it means to be `public`. It'd be very hard to tell that `public: false` means we should exclude that channel's articles from sitemaps. From a n00b perspective, I'd guess that `public` means you shouldn't show that Channel's articles. 

On articles, we've used `exclude_google_news` which might be helpful for naming.

This is definitely the right solution, but it just feels like the naming is off. Also, we should have matching UI to be able to configure this!

eessex:
Was thinking that the test channel and artsy primer apply as non-public, meaning they aren't distributed to the public on artsy.net, or on google

joeyAghion:
I was just chatting with @kanaabe and trying to understand the current layers of public/private status around articles. (Since I don't have all the context, feel free to take this with a grain of salt.)

Gravity has suffered from lots of `public`-like fields that have overlapping and counterintuitive uses, so I have scars.

It seems like there are [at least] 4 potential exposure levels of article content today:
1. only visible to creator (`published: false`)
2. visible to anyone with the URL, but not linked from public forums: e.g., "primer" articles (excluded from sitemaps _and_ google results _and_ internal search)
3. visible to anyone with the URL and linked from public forums: e.g., jobs pages, partner articles, etc. (sitemapped and indexed by Google, but not surfaced in our internal search)
4. featured: e.g., editorial articles (surfaced in internal search)

Classes (1) and (4) have pretty clear flags that define them, but the difference between (2) and (3) is unclear. Since we're not actually sitemapping the channels but the articles, maybe there should be an additional field on articles to capture this? That would allow individual channels to have articles in the different classes at once. I'd consider calling it `indexable`. (It's odd that these still wouldn't be in our internal search, but maybe that can be corrected eventually.)

An alternative: if the "primer" article (2) is a hack and not a real-life business usage, I wouldn't add more modeling to support it (this will just get confusing). Maybe you could simply allow drafts to be visible on the public site, provided a certain authorization token is included in the querystring? Then your existing `published` and `featured` flags would be sufficient.

kanaabe:
Thanks for this insight @joeyAghion! 

I agree that that some of these exposure levels aren't as clear. If Primer becomes a thing, we'll definitely want to put it behind a login wall. We do have the ability to view drafts on the site if you're logged in as the creator or member of the channel (since we use Gravity users for Writer login). The creator should still have the ability to edit a draft without risking it being exposed to the collector which probably means more modeling.

A flag on the article that lets you decide if something is indexable makes the most sense to me. On the front-end we could name it something more friendly, but the idea to use this specifically for Google Search + sitemaps seems very clear. I'm sure there will be a time when Artsy Primer _wants_ to create an indexed article, so we probably shouldn't restrict it to the Channel. Right now we only use include Artsy Editorial articles in internal search which is more a branding decision I believe, but yea, it could work for internal eventually as well. 

If you want to set a field like this across a Channel, that could be a handled with defaults like we do on [create](https://github.com/artsy/positron/blob/master/client/apps/edit/routes.coffee#L8). Of course we'd require a backfill for past articles, but that's pretty trivial to do.

joeyAghion:
Yeah it's an interesting idea to add a `defaultIndexable` field on channels eventually, but if the only use case remains these odd primer ones... maybe it's sufficient for all articles to default to `indexable: true` (and for primer articles to opt-out).

Issue #1067, opened 2017-05-26
formatting on linked text is being stripped when pasting into Writer


halleyjohnson:
pasting from this doc
https://docs.google.com/a/artsymail.com/document/d/1FtBGAiomw85wDE-VetVQqssRhjyUMnyVAP3mkNL4wbo/edit?usp=sharing

in the doc the link is italicized
![screen shot 2017-05-26 at 1 01 31 pm](https://cloud.githubusercontent.com/assets/10982707/26504666/8a0f4ee8-4213-11e7-876e-f5c6daadc5ee.png)

in writer it's not
![screen shot 2017-05-26 at 1 01 27 pm](https://cloud.githubusercontent.com/assets/10982707/26504675/8f0c27d6-4213-11e7-84ff-77fa6faa375c.png)

and some links show up like this
![screen shot 2017-05-26 at 1 02 33 pm](https://cloud.githubusercontent.com/assets/10982707/26504692/a098aa60-4213-11e7-909f-a616c8b7f33f.png)

This has only happened in the news roundups so far, this doc is a great test case

PR #1068, opened 2017-05-26
@kanaabe => Bug pasting from google docs

eessex:
Closes https://github.com/artsy/positron/issues/1067

[This line](https://github.com/artsy/positron/compare/master...eessex:link-paste?expand=1#diff-8ac6d971617183b86f1d61823b3f2185R179) was replacing `span` with alternative elements, but it changed the length of the spans array while still iterating through it, losing some indexes before everything was replaced.  By wrapping the line in `<span>` tags, the array stays at its original length.  The tags are later removed when html is converted into an editor state.


kanaabe:
👍 gratzi

PR #1072, opened 2017-05-30
@eessex => Fix null fetch with background image check

kanaabe:
Closes https://github.com/artsy/positron/issues/942

Turns out it wasn't the SVG but a missing background image instead. The problem was that `background_image: url(undefined)` was doing a GET `/null` on root. For published, this was not a problem because all published articles require a thumbnail_image. Instead, don't try to fetch the background image if there isn't one.


eessex:
Aha!  Thanks for looking into this with me, the looks ready to merge when 🍏

PR #1073, opened 2017-05-30
@damassi/eessex => Use Sentry

kanaabe:
@damassi I'm making you a Reviewer on this PR but since this is your first one on Positron, I'm co-assigning @eessex as well. Positron is very similar to Force except that we split the repo into client and api. I added Sentry to both the client and api side. Let me know if you have any questions!


client/lib/setup/index.coffee
```diff
@@ -49,6 +54,10 @@ module.exports = (app) ->
   app.use bucketAssets()
   setupAuth app
   app.use locals
+
+  if SENTRY_PRIVATE_DSN
+    app.use RavenServer.errorHandler()
```
damassi: Just an FYI, Sentry error-handling must come before all other error handling on the server: https://docs.sentry.io/clients/node/integrations/express/

kanaabe: @damassi correct me if I'm wrong but I think this does happen before all other error handling. The main error handler is at the end of the file. Pretty sure it should be moved to after we mount the sub-apps tho. I'll do that now!

Issue #995, opened 2017-03-29
Create a way to exclude some channels from sitemaps


dblock:
We have channels in production like `Kanaz Test Channel` which ends up being included in sitemaps.

Maybe channels could have a `public` flag or something like that, although that's starting to smell like gravity ;)

eessex:
- [ ]  Writer: Add indexable field to articles api + user interface
- [ ]  Force: Add no-index meta tag to articles if indexable is false
- [ ]  Cinder: Exclude articles without indexable from sitemaps

PR #1076, opened 2017-05-31
@kanaabe => Add indexable fields to articles

eessex:
Part 1 of 4 for https://github.com/artsy/positron/issues/995

- Adds indexable field to articles api
- Adds user interface to change indexable on admin panel


kanaabe:
Woohoo! Looks great.

PR #1081, opened 2017-06-01
@kanaabe => Set indexable false on non-indexable channels

eessex:
If channel is not-indexable, sets article.indexable to false. 
Does not display UI for setting indexable if article is from non-indexable channel.


kanaabe:
Wouldn't we still want to give channels that have a default `indexable: false` the option to index if they want to? If Primer writes a public article, they wouldn't be able to uncheck the field otherwise. 

The power of a backfilling in this case is that we only have to worry about setting defaults on article creation, as opposed to also worrying about converting old articles. That means we could acheive the desired behavior by setting this:  https://github.com/artsy/positron/blob/master/client/apps/edit/routes.coffee#L8.

eessex:
@kanaabe this PR has been updated to set indexable on create

PR #1084, opened 2017-06-05
@craigspaeth => Refactor image collection 

eessex:
In anticipation to more functionality, a refactor moving elements of image collection into separate files:
ImageCollection is the parent container (index), with three children
- ImageDisplay
- ArtworkDisplay
- Controls

### DIFF
--- a/client/apps/edit/components/section_image_collection/index.coffee
+++ b/client/apps/edit/components/section_image_collection/index.coffee
@@ -3,151 +3,55 @@
 #
 
 _ = require 'underscore'
-gemup = require 'gemup'
 React = require 'react'
-sd = require('sharify').data
-icons = -> require('./icons.jade') arguments...
-Autocomplete = require '../../../../components/autocomplete/index.coffee'
-Artwork = require '../../../../models/artwork.coffee'
 imagesLoaded = require 'imagesloaded'
-RichTextCaption = React.createFactory require '../../../../components/rich_text_caption/index.coffee'
-UrlArtworkInput = React.createFactory require '../section_image_set/url_artwork_input.coffee'
-
-{ div, section, h1, h2, span, img, header, input, a, button, p, ul, li, strong, nav } = React.DOM
-{ resize } = require '../../../../components/resizer/index.coffee'
+DisplayArtwork = React.createFactory require './components/artwork.coffee'
+DisplayImage = React.createFactory require './components/image.coffee'
+Controls = React.createFactory require './components/controls.coffee'
+{ div, section, ul, li } = React.DOM
 
 module.exports = React.createClass
   displayName: 'SectionImageCollection'
 
   getInitialState: ->
-    images: @props.section.get('images') or []
-    layout: @props.section.get('layout') or 'overflow_fillwidth'
     progress: null
 
   componentDidMount: ->
     @$list = $(@refs.images)
-    @setupAutocomplete()
     imagesLoaded @$list, =>
-      if @state.images.length > 1 and @state.layout is 'overflow_fillwidth'
-        @toggleFillwidth()
-      else
-        @$list.animate({opacity: 1}, 500)
+      @onChange()
+      @$list.animate({opacity: 1}, 250) if @props.section.get('layout') != 'column_width'
 
-  componentWillUnmount: ->
-    @autocomplete.remove()
-
-  componentWillUpdate: ->
-    @props.section.set images: @state.images if @state.images.length > 0
-
-  setupAutocomplete: ->
-    $el = $(@refs.autocomplete)
-    @autocomplete = new Autocomplete
-      url: "#{sd.ARTSY_URL}/api/search?q=%QUERY"
-      el: $el
-      filter: (res) ->
-        vals = []
-        for r in res._embedded.results
-          if r.type?.toLowerCase() == 'artwork'
-            id = r._links.self.href.substr(r._links.self.href.lastIndexOf('/') + 1)
-            vals.push
-              id: id
-              value: r.title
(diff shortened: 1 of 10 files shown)


client/apps/edit/components/section_image_collection/index.coffee
```diff
@@ -3,151 +3,55 @@
 #
```
craigspaeth: 👍  nice breaking down of this component.

client/apps/edit/components/section_image_collection/index.coffee
```diff
@@ -3,151 +3,55 @@
...
-  onSelect: (e, selected) ->
-    new Artwork(id: selected.id).fetch
-      success: (artwork) =>
-        newImages = @state.images.concat [artwork.denormalized()]
-        @setState images: newImages
-        @props.section.set images: newImages
-        $(@refs.autocomplete).val('').focus()
-        @toggleFillwidth() if @state.images.length > 1
+      @onChange()
+      @$list.animate({opacity: 1}, 250) if @props.section.get('layout') != 'column_width'
```
craigspaeth: If possible it's often nicer to avoid using jQuery to manipulate the DOM with React—even in the case of animations. A common way to do this is to use CSS transitions that you toggle in `render`. e.g.

````coffeescript
render: ->
  # ...
  ul {
    className: 'esic-images-list',
    'data-layout': @props.section.get('layout'),
    ref: 'images'
  },
    # ...
````

````stylus
.esic-images-list
  transition opacity 2.5ms
  &:not([data-layout="column_width"])
    opacity 1
````

Not a blocker for this PR and there are legit reasons to do it this way too. Just food for thought.

File: client/apps/edit/components/section_image_collection/index.coffee [hunk 3]
```diff
@@ -3,151 +3,55 @@
...
-
-  changeLayout: (layout) -> =>
-    @setState layout: layout
-    @props.section.set layout: layout
-    if layout is 'overflow_fillwidth' and @state.images.length > 1
-      @toggleFillwidth()
+  onChange: ->
+    @forceUpdate() if @$list.children().length isnt @props.section.get('images').length
+    if @props.section.get('images').length > 1 and @props.section.get('layout') != 'column_width'
+      @fillwidth()
```
craigspaeth: I think @broskoski created a React version of fillwidth that could be nicer to integrate into this now.

client/apps/edit/components/section_image_collection/test/image.coffee
```diff
@@ -0,0 +1,60 @@
...
+{ resolve } = require 'path'
+React = require 'react'
+ReactDOM = require 'react-dom'
+ReactTestUtils = require 'react-addons-test-utils'
+ReactDOMServer = require 'react-dom/server'
+r =
+  find: ReactTestUtils.findRenderedDOMComponentWithClass
+  simulate: ReactTestUtils.Simulate
+
+describe 'ImageCollectionImage', ->
```
craigspaeth: Nice refactoring of tests too!

client/apps/edit/components/section_image_collection/index.coffee
```diff
@@ -3,151 +3,55 @@
...
-              @setState progress: null
-              @$list.animate({opacity: 1}, 500)
-
-  changeLayout: (layout) -> =>
-    @setState layout: layout
-    @props.section.set layout: layout
-    if layout is 'overflow_fillwidth' and @state.images.length > 1
-      @toggleFillwidth()
+  onChange: ->
+    @forceUpdate() if @$list.children().length isnt @props.section.get('images').length
```
craigspaeth: The way `forceUpdate` is used here is a bit of a code smell in React. Was React not re-rendering in certain cases?

craigspaeth:
Overall looks pretty good to me. The one thing that feels could be better is [the use of `onChange`](https://github.com/artsy/positron/pull/1084#discussion_r120412801). I wonder if there's a better way to achieve this through hooking into events on the Backbone collection `@props.section` or through `setState`. Typically you want to avoid having to manually tell React to do rendering things and would rely on automatic updates or hooking into lifecycle events.

client/apps/edit/components/section_image_collection/index.coffee
eessex: Sounds like a good idea, any idea where this is published? Didn't see it on @broskoski's GH profile

craigspaeth: Currently, it's [inside Reaction Force](https://github.com/artsy/reaction-force/blob/master/src/components/artwork/fillwidth.tsx), so not for this PR but down the line it might be worth pairing with him or asking in [#collector-gmv-dev](https://artsy.slack.com/messages/C0GNAB0US) if someone would pair on it.

client/apps/edit/components/section_image_collection/index.coffee
eessex: The artwork autocomplete (in controls) isn't changing the state of this component when something is selected, so I need to rerender manually.

client/apps/edit/components/section_image_collection/index.coffee
eessex: The jquery in here is all related to the imagesLoaded component -- going to keep this for now, but will definitely get in touch with someone from GMV on the topic of refactoring the react version for use here.

eessex:
@craigspaeth thanks for your comments on this -- i'm looking towards refactoring some of the higher level components soon, which should clear up some of the overlap of backbone and react, and less separation between the 'source of truth' -- will send some of these PRs your way for input as they come up.

Issue #1080, opened 2017-06-01
Author Admin


owendodd:
New way to admin authors, which will allow us to incorporate social handles / bios / photos. Should live as a section under 'Settings'. 

We will need to do a backfill once this is up and running to switch articles associated w/ users to being associated w/ an author instead.

![artboard](https://cloud.githubusercontent.com/assets/3196536/26698318/84153776-46e3-11e7-883d-b6e629108785.png)

Sketch: https://cl.ly/000j1M303c1s

PR #1089, opened 2017-06-07
@craigspaeth => Authors API

kanaabe:
API-side of https://github.com/artsy/positron/issues/1080

The first commit is pretty straightforward and similar to other API endpoints we've done. 

The second commit is an attempt to use extensions to validate + typecast mongo objectids from a string. I experimented a bunch with this, and found that adding a validation on `string` was the most straightforward way to achieve both validation + typecasting. I posted [this issue](https://github.com/hapijs/joi/issues/1212) that could be a cleaner solution -- ie being able to say `Joi.objectId()` instead of `Joi.string().objectId()` but it doesn't seem possible.

### DIFF
--- a/api/apps/authors/index.coffee
+++ b/api/apps/authors/index.coffee
@@ -1,0 +1,11 @@
+express = require 'express'
+routes = require './routes'
+{ setUser, authenticated, adminOnly } = require '../users/routes'
+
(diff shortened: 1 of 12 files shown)


craigspaeth:
Nice! Super straightforward and good test coverage. Thanks!

PR #1095, opened 2017-06-13
@eessex => Author Admin client

kanaabe:
Closes https://github.com/artsy/positron/issues/1080

Components are broken down as so: `AuthorsView` -> `AuthorModal` -> `AuthorImage`. I've been curious about React [HOC's](https://facebook.github.io/react/docs/higher-order-components.html) to refactor out some of the duplication regarding the image uploading. I'm sure also the modal could be reused at some point too. 

author-admin


client/apps/edit/components/admin/index.styl
```diff
@@ -39,20 +39,6 @@
     opacity 1
     transition all .65s
     overflow visible
-  .fields-left,
```
eessex: smart to move this

api/apps/graphql/resolvers.coffee
```diff
@@ -32,4 +33,9 @@ module.exports.channels = (root, args, req, ast) ->
 module.exports.tags = (root, args, req, ast) ->
   return new Promise (resolve, reject) ->
     Tag.where args, (err, results) ->
+      resolve results.results
+
+module.exports.authors = (root, args, req, ast) ->
+  return new Promise (resolve, reject) ->
```
eessex: what does ast refer to?  new to me

eessex:
Looks great!  Merging. Also re: image component, sounds like a good idea, we have uploads repeated all around --  I'm down to pair on building something for this.

api/apps/graphql/resolvers.coffee
kanaabe: ast stands for abstract syntax tree. you can think of it like an internal tree/json/schema that Graphql uses internally to process your request. your query gets converted to an ast - a format that GraphQL understands. try printing it out next time!

PR #1096, opened 2017-06-13
@kanaabe => Draggable Images

eessex:
Creates a drag_drop component, that accepts any components as children.  It requires an isDraggable boolean, a list of items, and an onDragEnd event as props. 

Currently this looks different from the section list, it is optimized for horizontal movement.  (Rather than rendering a placeholder above or below an item, it replaces the dragged-over content with a placeholder).

Next steps will be moving vertical placeholder logic into this component from the Section List, and then this component can be for the section list as well. 

drag-drop-images


kanaabe:
Cool! Nice drag_drop component. I'm looking forward to seeing this used for Section List

PR #1101, opened 2017-06-14
@kanaabe => Update captions, Video section mini-refactor

eessex:
- Converts rich-text-caption input to use same 'popup' menu as text-component
- Adds key command for link to rich-text-caption
- Video and Image section to use same caption style as image-collection (moved out of controls)

In video section- 
- Renders of caption field when cover image present
- Removes background color options (per conversation w @owendodd)
- Adds ability to remove a cover image (before had to delete section or replace)

### DIFF
--- a/client/apps/edit/components/section_video/icons.jade
+++ b/client/apps/edit/components/section_video/icons.jade
@@ -1,0 +1,2 @@
+.remove
+  include ../../public/icons/edit_section_remove.svg
--- a/client/apps/edit/components/hero_section/index.styl
+++ b/client/apps/edit/components/hero_section/index.styl
@@ -35,3 +35,5 @@
       line-height 500px
     iframe
       min-height 590px
+  .esic-img-container
+    padding 20px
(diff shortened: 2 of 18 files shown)


client/apps/edit/components/section_video/icons.jade
```diff
@@ -0,0 +1,2 @@
+.remove
+  include ../../public/icons/edit_section_remove.svg
```
kanaabe: not for this PR and more just a random thought, but if we're comfortable with continuing this pattern of including svg's, it would be nice to create a single `icons.jade` file that maps all icons so we don't have to keep creating these files.

eessex: for sure!  making an issue for it now

kanaabe:
Looks good -- I would just make a note of the one comment and try to figure out a better way to target the pieces of an SVG but it's not a blocker for me. Feel free to follow up with something later.

eessex:
Sounds good -- agreed the SVGs are a little funky, will keep this in mind!

PR #1105, opened 2017-06-15
@kanaabe : add freshness boost when indexing for search #migration

cavvia:
This uses [exponential decay](https://en.wikipedia.org/wiki/Exponential_decay) on the `published_at` date of the article to produce a freshness boost for articles on artsy.net site search. It should fix the issue of fair booths ranking above articles when searching fair names.

We cap the maximum score at 1000 for a new article and use a 3 year mean lifetime, meaning the exponential curve drops to a score of around 360 in that time period - meaning most articles, even old ones, will be boosted significantly higher than most booths, partners, shows, and artists in search.

This can be merged immediately, but will rely on the follow up 
 to https://github.com/artsy/gravity/pull/11108 in order to function in production.

**Migration**

I will do a full search reindex of articles once this is merged:

```shell
yarn coffee scripts/reindex_articles.coffee
```

### DIFF
--- a/scripts/count.coffee
+++ b/scripts/count.coffee
@@ -1,0 +1,12 @@
+require('node-env-file')(require('path').resolve __dirname, '../.env')
+mongojs = require 'mongojs'
+path = require 'path'
+Article = require '../api/apps/articles/model/index'
+env = require 'node-env-file'
+
+# Connect to database
+db = mongojs(process.env.MONGOHQ_URL, ['articles'])
+
+db.articles.find(indexable: true).toArray (err, articles) ->
+    console.log(articles.length)
+    #console.log(article.thumbnail_image) for article in articles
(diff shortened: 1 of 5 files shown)


scripts/count.coffee
```diff
@@ -0,0 +1,12 @@
...
+path = require 'path'
+Article = require '../api/apps/articles/model/index'
+env = require 'node-env-file'
+
+# Connect to database
+db = mongojs(process.env.MONGOHQ_URL, ['articles'])
+
+db.articles.find(indexable: true).toArray (err, articles) ->
+    console.log(articles.length)
+    #console.log(article.thumbnail_image) for article in articles
```
kanaabe: stray log

scripts/reindex_articles.coffee
```diff
@@ -1,7 +1,7 @@
 require('node-env-file')(require('path').resolve __dirname, '../.env')
 mongojs = require 'mongojs'
 path = require 'path'
-{ indexForSearch } = Save = require '../api/apps/articles/model/save'
+{ indexForSearch } = Save = require '../api/apps/articles/model/distribute'
```
kanaabe: ah thanks for catching this

kanaabe:
Looks awesome! Can't wait to see this in action

Issue #1097, opened 2017-06-13
Remove lead paragraph from fullscreen articles


kanaabe:
Related to : https://github.com/artsy/force/issues/1530

kanaabe:
It's actually the `hero_section.intro`, not `lead_paragraph`

PR #1107, opened 2017-06-15
@eessex => Remove intro text from fullscreen component

kanaabe:
Closes https://github.com/artsy/positron/issues/1097

image


eessex:
Looks good from here, merging!

PR #1108, opened 2017-06-16
@wrgoldstein => Send recently added fields to Fulcrum

kanaabe:
I've added a few more fields than just `vertical` and `internal_tags` in here that I think could be useful. For Vertical, I'm just pulling the name of the vertical if there is one -- does that work for you? `internal_tags` will be in the same array of string format as `tags`.

Addresses https://trello.com/c/bmUsuoLm/1125-positron-verticals-and-authors-in-rollups

### DIFF
--- a/scripts/daily_upload_s3.coffee
+++ b/scripts/daily_upload_s3.coffee
@@ -24,7 +24,36 @@ db = mongojs(process.env.MONGOHQ_URL, ['articles'])
 filename = "export_" + moment().format('YYYYMMDDhhmmss') + ".csv"
 dir = 'scripts/tmp/'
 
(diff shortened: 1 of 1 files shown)


wrgoldstein:
Lgtm thanks!

wrgoldstein:
Sometimes the article`thumbnail_image` is actually a raw jpeg encoding, which end up being more characters than are allowed in a single field in our data warehouse. An example is the article with id `54f6498c11c94a06005bb925`, which has a 60327 character image in that column.

```
id                                                      54f6498c11c94a06005bb925
author_id                                               503f86e462d56000020002cc
...
thumbnail_image                data:image/jpeg;base64,/9j/4AAQSkZJRgABAQEASAB...
thumbnail_title                Paintings by John Evans Make Us Question, Has ...
...
description                    John Evans is a painter of nature who does not...
```

Issue #1098, opened 2017-06-13
Reactify fillwidth


eessex:
At suggestion of @craigspaeth, a refactored version of fillwidth that gives widths to components as props, rather than assigning with jquery -- inspired by Cab's version in Reaction Force: https://github.com/artsy/reaction-force/blob/master/src/components/artwork/fillwidth.tsx

eessex:
Should be ported into reaction force as well

Issue #1071, opened 2017-05-30
Drag and drop within image pairs and image sets


owendodd:
- [x]  Ability to reorder images within and image paring (in an image section)
- [x]  Ability to reorder images within an image set

eessex:
@owendodd, our section menu is now shorter - which layout do you prefer?
![screen shot 2017-06-22 at 2 03 04 pm](https://user-images.githubusercontent.com/1497424/27448717-c8fc5ba0-5753-11e7-8389-c445e33ba2c3.png)
![screen shot 2017-06-22 at 2 04 05 pm](https://user-images.githubusercontent.com/1497424/27448718-c8fd68e2-5753-11e7-8ecc-97c39a13951f.png)

owendodd:
@eessex Lets do the single line!

PR #1129, opened 2017-06-22
@kanaabe => Merge image_set into image_collection

eessex:
image-set-switching

Finished, but cannot merge until Reaction-Force is deployed.
Related to: https://github.com/artsy/positron/issues/1098, https://github.com/artsy/positron/issues/1071

This PR adds ability to toggle an image_collection to an image set.  The section type is not converted, but the UI is all housed in image_collection. 

- adds buttons to toggle image_set layout
- adds fill_width component, which returns an array of dimensions without jQuery
- adds Artwork and ImagesetPreview components from Reaction Force
- Removes image_set and artwork components
- Removes outdated features from SectionTool


eessex:
Reaction-force is deployed, this is now ready to go.

client/apps/edit/components/content/sections/image_collection/components/controls.coffee
```diff
@@ -17,10 +17,21 @@ module.exports = React.createClass
...
     @props.section.set layout: e
     @props.onChange()
 
+  toggleImageSet: ->
+    if @props.section.get('type') is 'image_collection'
+      @props.section.unset 'layout'
+      @props.section.set 'type', 'image_set'
+    else
+      @props.section.set 'layout', 'overflow_fillwidth'
+      @props.section.set 'type', 'image_collection'
```
kanaabe: You can call these two lines in a single `set`.

client/apps/edit/components/content/sections/image_collection/test/controls.coffee
```diff
@@ -86,11 +115,14 @@ describe 'ImageCollectionControls', ->
...
-  xit 'saves an artwork by url', ->
+  it 'saves an artwork by url', ->
     input = r.find(@component, 'bordered-input')[1]
-    input.value = 'http://artsy.net/artwork/this-art'
+    input.value = 'http://artsy.net/artwork/skull7'
     r.simulate.change input
     r.simulate.click r.find(@component, 'esis-byurl-button')[0]
+    Backbone.sync.args[0][2].success fabricate 'artwork'
+    @component.props.section.get('images')[0].type.should.eql 'artwork'
+    @onChange.called.should.eql true
```
kanaabe: 👍

client/apps/edit/public/icons/edit_image_set_slideshow.svg
```diff
@@ -1 +1 @@
-<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 44 44"><defs><style>.cls-1{fill:#231f20;}</style></defs><title>Untitled-1</title><path class="cls-1" d="M6760.67,551.72h-32v-32h32v32Zm-30-2h28v-28h-28v28Z" transform="translate(-6728.67 -507.72)"/><polygon class="cls-1" points="38 37 36 37 36 8 7 8 7 6 38 6 38 37"/><polygon class="cls-1" points="44 31 42 31 42 2 13 2 13 0 44 0 44 31"/></svg>
\ No newline at end of file
+<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 44 44"><defs><style>.cls-1{fill:#ffffff;}</style></defs><title>Untitled-1</title><path class="cls-1" d="M6760.67,551.72h-32v-32h32v32Zm-30-2h28v-28h-28v28Z" transform="translate(-6728.67 -507.72)"/><polygon class="cls-1" points="38 37 36 37 36 8 7 8 7 6 38 6 38 37"/><polygon class="cls-1" points="44 31 42 31 42 2 13 2 13 0 44 0 44 31"/></svg>
```
kanaabe: I've mentioned this before but in times like this, feel free to change the classname within the SVG if you need to access the fill or other properties. That way you should be able to remove the embedded style tag entirely and use the classname for setting properties via Styl instead of inline. Not necessary here, but just something to think about.

client/apps/edit/components/content/sections/image_collection/components/artwork.coffee
```diff
@@ -1,37 +1,20 @@
 React = require 'react'
+components = require('@artsy/reaction-force/dist/components/publishing/index').default
+Artwork = React.createFactory components.Artwork
```
kanaabe: Woohoo!

client/apps/edit/public/icons/edit_image_set_slideshow.svg
eessex: This is the svg that was added to reaction-force, I wonder if it makes sense to start using that version instead, and pass color etc as props?  The icons folder in reaction-force isn't exported yet, but could easily be.

kanaabe:
I think a GIF would be nice here to showcase the type switching.

I saw one test for the url changing, but I didn't see anything that tests the "glue", ie: when that change does occur, does the Artwork component render correctly as well?

kanaabe:
Also curious why the lock file grew so big?

eessex:
Artwork component rendering is tested here - https://github.com/eessex/positron/blob/56b0d9755a36a7b347b2a5a7a28689900483c333/client/apps/edit/components/content/sections/image_collection/test/index.coffee#L76

I'm not sure what you mean about glue?  It tests for the presence of rendered images/artworks in both layouts. Not sure about the lock file -- i deleted and rebuilt it after adding reaction-force (it was altered after removing particle2), so maybe it updated lots of versions. 

Also just found a css bug - adding a fix now.

kanaabe:
I'm a little uneasy introducing that many changes to our dependencies with that lockfile. Do you mind pulling it down from master, and only updating reaction-force? The one dependency that caught my eye is mongodb which we've had problems with in the past.

eessex:
Makes sense to me! Just replaced with a new lock file.

kanaabe:
Sorry for the delay -- this is ready to go code-wise but we now have a merge conflict.

Once resolved, please go ahead and merge, test on staging, and deploy. The auto-deploy to staging hasn't changed (except that we handle it by deploying and then merging changes into a staging branch), but the production deploy is changed to the staging -> release branch method like we do in Force. Go ahead and test that out

PR #1130, opened 2017-06-22
@craigspaeth => Refactor API and use custom Joi extension

kanaabe:
- Removes the joi-objectid plugin
- Every `where` method is separated into `where` (with Joi validation) and `mongoFetch` which makes a mongo query
- `Article.save` now receives options, which are used for determining if the Article should save with defaults or not. This is important for distinguishing between a `update` and `create`. Will explain this more below.

See below for more comments!


api/apps/articles/index.coffee
```diff
@@ -7,5 +7,5 @@ app = module.exports = express()
 app.get '/articles', setUser, routes.index
 app.get '/articles/:id', routes.find, routes.show
 app.post '/articles', setUser, authenticated, routes.restrictFeature, routes.create
-app.put '/articles/:id', setUser, authenticated, routes.find, routes.restrictFeature, routes.update
+app.put '/articles/:id', setUser, authenticated, routes.restrictFeature, routes.update
```
kanaabe: We no longer `find` the article first since
1. We already `find` once within `save`
2. For PUT, we don't want to merge an existing article + input _before_ Joi.validate because we'd have to `.toString()` all the `ObjectIds` again. Before, the merge was helpful because it overrode defaults, but now there is an option to exclude that. Ends up being much cleaner.

api/apps/articles/model/index.coffee
```diff
@@ -6,40 +6,47 @@ _ = require 'underscore'
...
+    (cb) ->
+      return cb() unless count
+      cursor.count cb
+  ], (err, [articles, total, articleCount]) ->
+    return callback err if err
+    callback null, {
+      results: articles
+      total: total
+      count: articleCount
+    }
```
kanaabe: These `mongoFetch`es are separated to support JoiQL which we've discussed :)

api/apps/articles/model/index.coffee
```diff
@@ -48,29 +55,38 @@ Q = require 'bluebird-q'
...
         else if unPublishing
-          onUnpublish article, sanitizeAndSave(callback)
-        else if not publishing and not article.slugs?.length > 0
-          generateSlugs article, sanitizeAndSave(callback)
+          onUnpublish modifiedArticle, sanitizeAndSave(callback)
+        else if not publishing and not modifiedArticle.slugs?.length > 0
+          generateSlugs modifiedArticle, sanitizeAndSave(callback)
         else
-          sanitizeAndSave(callback)(null, article)
+          sanitizeAndSave(callback)(null, modifiedArticle)
```
kanaabe: I've cleaned this up a bit to mitigate some of the mutations we were doing.

`modifiedArticle` - extended copy of article with input
`input` - only the new fields that are added
`article` - the original article, or {} if it is new.

This helps simplify/reason in functions like `generateKeywords` where figuring out if we should generate new keywords was kind of a pain since input was just a merged version of input + article.

api/apps/articles/model/save.coffee
```diff
@@ -68,22 +55,26 @@ removeStopWords = (title) ->
...
 
 @generateKeywords = (input, article, cb) ->
+  unless (
+    input.primary_featured_artist_ids or
+    input.featured_artist_ids or
+    input.fair_ids or
+    input.partner_ids or
+    input.contributing_authors or
+    input.tags
+  )
```
kanaabe: This is where it's easier to reason that `input` only contains the new fields we want to update.

api/apps/articles/model/retrieve.coffee
```diff
@@ -1,88 +1,101 @@
...
-    query.auction_ids = ObjectId input.auction_id if input.auction_id
-    query.section_ids = ObjectId input.section_id if input.section_id
-    if input.biography_for_artist_id
-      query.biography_for_artist_id = ObjectId input.biography_for_artist_id
-    query.featured_artwork_ids = ObjectId input.artwork_id if input.artwork_id
-    query.tags = { $in: input.tags } if input.tags
-    query.tracking_tags = { $in: input.tracking_tags } if input.tracking_tags
+  { limit, offset, sort, count } = input
+  query = _.omit input,
+    'limit'
```
craigspaeth: @damassi pro tip: Long vertical lists like this can benefit from alphabetical sorting (F5 in Sublime).

damassi: Oh yeah, I'm kinda a fanatic about alpha 🙂 Everything I come across I tend to sort

damassi: But I think you mean @kana!

api/apps/articles/model/save.coffee
craigspaeth: Nitpick: I find that `unless` reads better as a postfix, e.g.

```coffee
return cb null, article unless (
 input. primary_featured_artist_ids or
  ...
)
```

api/apps/articles/test/integration.coffee
```diff
@@ -83,7 +85,11 @@ describe 'articles endpoints', ->
 
     it 'gets a list of articles by channel', (done) ->
       fabricate 'articles', [
-        { title: 'Winter Is Coming', channel_id: ObjectId('5086df098523e60002000012'), published: true }
+        {
+          title: 'Winter Is Coming'
```
craigspaeth: Less than a month now!!!

craigspaeth:
👏  great refactor! Much cleaner now!

api/apps/articles/model/retrieve.coffee
craigspaeth: EDIT: Pro tip from @damassi 😬

kanaabe:
We lost semaphore for now, but i'm close to switching over to Circle so I'd like to just wait on that!

craigspaeth:
D'oh! I did it again! I didn't realize CI wasn't running. 🤦‍♂️

PR #1131, opened 2017-06-22
@kanaabe => imageset caption bug

eessex:
Adds editing prop to image-set caption (is readonly without it)


kanaabe:
We lost semaphore :(

PR #1132, opened 2017-06-23
@kanaabe => Migrate to CircleCI

kanaabe:
Will likely need to be tweaked, but just to get the builds going to test this...

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,8 @@
 machine:
   node:
     version: 7
+  pre:
+    - mkdir ~/.yarn-cache
 deployment:
   staging:
     branch: master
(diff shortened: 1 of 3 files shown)


circle.yml
```diff
@@ -0,0 +1,27 @@
...
+  staging:
+    branch: master
+    owner: artsy
+    commands:
+      - NODE_ENV=production make deploy env=staging
+  production:
+    branch: release
+    owner: artsy
+    commands:
+      - NODE_ENV=production make deploy env=production
```
orta: This matches with https://github.com/artsy/metaphysics/blob/master/circle.yml - should be 👍

kanaabe: Oh wait but these won't actually push to their respective branches

orta: No, it'd mean your deploy process is `git push upstream master:staging` or `git push upstream master:release`

kanaabe: Right. Wonder which is better -- to use the CIRCLE_SHA or pull a sha with git? https://github.com/artsy/positron/blob/master/Makefile#L52

craigspaeth: You'll want to make sure to update the deploy command to take `$CIRCLE_SHA1` into consideration [here](https://github.com/artsy/positron/blob/master/Makefile#L53). e.g. `git push --force git@heroku.com:positron-$(env).git $CIRCLE_SHA1:refs/heads/master`.

orta: Clever

Issue #1087, opened 2017-06-06
Articles can't save with 0 sections


eessex:
https://github.com/artsy/positron/pull/915
"Problem

If you go from having sections in an article to deleting all the sections and try to save an empty article, it will fall back to its previous attrs.

I think this predates me, but i wonder what the reason we wanted to fallback on attrs was. It seems like the reason was that on the edit router, we need to do article.toJSON() before rendering but we didn't have article.sections defined when first initializing.

I'm sure there was a good reason at the time but if we listen for changes to the article's attrs, we can set article.sections after the sync is triggered."

PR #1135, opened 2017-06-23
@craigspaeth => Save with 0 sections

eessex:
Closes https://github.com/artsy/positron/issues/1087

The collections don't seem to be sync'ing when empty, so this adds a check for an empty Section Collection when removing sections, and manually resets the article.sections to an empty array if true.


craigspaeth:
👍  Pretty straightforward! Thanks!

PR #1137, opened 2017-06-23
@craigspaeth => Move deploy commands to circle and upgrade bucket-assets

kanaabe:
- Upgrade bucket-assets
- Remove make deploy from makefile

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -19,7 +19,9 @@ deployment:
         EOF
       - chmod 600 ~/.netrc
       - "[[ ! -s \"$(git rev-parse --git-dir)/shallow\" ]] || git fetch --unshallow"
-      - NODE_ENV=production make deploy env=staging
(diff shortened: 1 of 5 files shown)


circle.yml
```diff
@@ -19,7 +19,9 @@ deployment:
         EOF
       - chmod 600 ~/.netrc
       - "[[ ! -s \"$(git rev-parse --git-dir)/shallow\" ]] || git fetch --unshallow"
-      - NODE_ENV=production make deploy env=staging
+      - node_modules/.bin/bucket-assets -b positron-staging
+      - heroku config:set COMMIT_HASH=$(shell git rev-parse --short HEAD) --app=positron-staging
```
craigspaeth: This should use ASSET_MANIFEST now like [this](https://github.com/artsy/force/blob/master/scripts/deploy.sh#L9)

`heroku config:set ASSET_MANIFEST=$(cat manifest.json)`

PR #1139, opened 2017-06-24
@kanaabe => Fix CI tests

kanaabe:
Some new failed tests emerged because of [file ordering](https://circleci.com/docs/1.0/file-ordering/).

When this happens again in the future, a simple way to reproduce a failing test from CI (after you've verified the "whole" test suite passes) is to start with the failing test.

`yarn mocha /Users/kanaabe/Development/Artsy/positron/client/apps/switch_channel/test/index.coffee`
If that works, go backwards and follow the file order that CI is picking up instead of the order your file system decides to. So, the file before this one that Circle decided to run was 

`yarn mocha /Users/kanaabe/Development/Artsy/positron/client/apps/settings/test/routes.coffee /Users/kanaabe/Development/Artsy/positron/client/apps/switch_channel/test/index.coffee`

You'll eventually hit the problematic file when the test fails. Unfortunately this process takes a while, but you'll definitely be able to reproduce the issue. 

In this case, the problematic file was `settings/test/client/curations.coffee` even though the test was failing on `switch_channel/test/index.coffee`.

### DIFF
--- a/client/apps/settings/test/client/curations/venice_admin.coffee
+++ b/client/apps/settings/test/client/curations/venice_admin.coffee
@@ -28,7 +28,6 @@ describe 'VeniceAdmin', ->
       $.fn.typeahead = sinon.stub()
       window.jQuery = $
       VeniceAdmin = benv.require resolve __dirname, '../../../client/curations/venice_admin.coffee'
-      # VeniceSection = benv.require resolve __dirname, '../../../client/curations/venice_section.coffee'
       Autocomplete = benv.require resolve __dirname, '../../../../../components/autocomplete_list/index.coffee'
       dropdownHeader = benv.requireWithJadeify(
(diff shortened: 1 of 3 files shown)


client/apps/settings/test/client/curations/venice_admin.coffee
```diff
@@ -81,8 +79,7 @@ describe 'VeniceSection', ->
   it 'Clicking save button saves the curation and removes warnings from save button', ->
     r.simulate.click r.findTag(@component, 'button')[0]
     @component.state.isSaving.should.eql true
-    @component.state.curation.save.callCount.should.eql 1
-    setTimeout( =>
-      @component.state.isSaving.should.eql false
-      @component.state.isChanged.should.eql false
-    , 750)
```
kanaabe: Generally we should avoid using `setTimeout` in tests, especially without a `done` callback. This looks to be the main issue.

PR #1145, opened 2017-06-26
@kanaabe => Autocomplete bug

eessex:
Fixes a bug for AutocompleteList with inline props -- only prints dragContainer if @props.draggable is set to true (rather than just setting the dragContainer to draggable: false).

### DIFF
--- a/client/components/autocomplete_list/test/index.coffee
+++ b/client/components/autocomplete_list/test/index.coffee
@@ -21,6 +21,8 @@ describe 'AutocompleteList', ->
         )
       $.fn.typeahead = sinon.stub()
       @AutocompleteList = benv.require resolve __dirname, '../index'
+      DragContainer = benv.require resolve __dirname, '../../drag_drop/index'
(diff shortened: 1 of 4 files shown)


client/components/autocomplete_list/test/index.coffee
```diff
@@ -21,6 +21,8 @@ describe 'AutocompleteList', ->
         )
       $.fn.typeahead = sinon.stub()
       @AutocompleteList = benv.require resolve __dirname, '../index'
+      DragContainer = benv.require resolve __dirname, '../../drag_drop/index'
+      @AutocompleteList.__set__ 'DragContainer', React.createFactory DragContainer
```
kanaabe: Curious, is it necessary to createFactory manually in this test? I would think that initializing AutocompleteList will create the component automatically. Maybe if we're doing something special to DragContainer but it seems like we're not here?

eessex: Yep that is correct - just removed

kanaabe:
👍  Looks good!

PR #1157, opened 2017-07-07
@craigspaeth => DragDrop QA

eessex:
Published articles were not re-rendering the list on drag-end, this adds a forceupdate to the section list if article is published


craigspaeth:
Works for me 👍 

I wonder though if there's something else gnarlier happening with events (e.g. `props.sections.on(..., @forceUpdate)` causing a race condition withing React's lifecycle. It's not apparent to me how that would happen by looking through the code right now, but food for thought.

Issue #1148, opened 2017-06-28
Deprecate toc


eessex:

eessex:
Double checking -- @owendodd we are planning to remove this feature entirely, including from existing articles?

owendodd:
@eessex Can we check to see how many articles are currently using. Pretty sure its just 1 and if so, then totally fine to remove entirely

kanaabe:
Just 1

PR #1175, opened 2017-07-13
@eessex => Adds new types to layout field + deprecate TOC

kanaabe:
Closes https://github.com/artsy/positron/issues/1148

Migration plan:
- [x] Remove tocs on articles
- [ ] Reset layout to `classic` across the board (after deploy)


client/apps/edit/components/content/sections/text/index.coffee
```diff
@@ -319,7 +289,6 @@ module.exports = React.createClass
   getPlugins: ->
     plugins = []
     plugins.push({label: 'artist', style: 'artist'}) if @hasFollow
-    plugins.push({label: 'toc', style: 'toc'}) if @hasToc
     return plugins
```
eessex: this will set off our menu to an odd number -- maybe we should add back the ordered list icon?  was removed to make space for strikethru, but functionality still exists.

eessex:
This looks ok to me, are you planning to manually remove the toc info from the article that it is used in?

client/apps/edit/components/content/sections/text/index.coffee
kanaabe: Seems like a good swap! I'll add that in. Are there any other text features like `ol` that are hidden atm?

kanaabe:
Yes, already removed 👍

client/apps/edit/components/content/sections/text/index.coffee
eessex: That's the only one -- there's an icon for it somewhere

PR #1186, opened 2017-07-20
@kanaabe => Update reaction image-set import

eessex:
- Changes ImagesetPreview naming to match new Reaction component
- Updates props for Artwork component


kanaabe:
Sweet! FYI the fillwidthDimensions method is also exposed if you wanted to swap that out later.

PR #1188, opened 2017-07-20
@kanaabe => Bare-bones admin 2

eessex:
Adds a route for edit2 that is limited to editorial channels only


kanaabe:
Curious why editorial only when we'll have to support the other channels too?

eessex:
Was thinking to prevent other channels from previewing while it is in progress, maybe i should restrict by user admin instead?

kanaabe:
👍  to Admin-only instead!

PR #1189, opened 2017-07-20
@kanaabe => Use regular attrs for lead paragraph

eessex:
Tiny step towards lead paragraph changes...
- Removes 'setLeadParagraph' and article.leadParagraph from article model
- Uses regular backbone attrs when setting or reading leadParagraph


kanaabe:
Looks good, but I'm trying to think of the reason why we wrapped this in a Backbone model in the first place..? Wonder what the reason was.

eessex:
It was from when this was a scribe / content editable -- the method wrapped plain text in html before saving, should have been deprecated when we switched to draft.

kanaabe:
👍 Makes sense. Thanks!

PR #1191, opened 2017-07-21
@kanaabe => Content2 / Article Header

eessex:
- Renders Content2 app if '/edit2' sd is true
- Sets editorial articles to 'standard' in '/edit2' - temp for preview purposes
- Adds layout name as class to 'edit-section-layout' container (top level of content app)
- Content2 can display both standard and classic header sections
- Displays lead paragraph conditionally if standard layout
- Does not allow hero in standard layout

screen shot 2017-07-21 at 2 05 18 pm

screen shot 2017-07-21 at 2 05 31 pm

### DIFF
--- a/client/apps/edit/client.coffee
+++ b/client/apps/edit/client.coffee
@@ -13,6 +13,7 @@ EditHeader = require './components/header/index.coffee'
 EditDisplay = require './components/display/index.coffee'
 EditAdmin = React.createFactory require './components/admin/index.coffee'
 EditContent = React.createFactory require './components/content/index.coffee'
+EditContent2 = React.createFactory require './components/content2/index.coffee'
 
 async = require 'async'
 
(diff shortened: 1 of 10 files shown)


client/apps/edit/client.coffee
```diff
@@ -27,10 +28,20 @@ async = require 'async'
...
     )
-    ReactDOM.render(
-      EditContent(article: @article, channel: channel)
-      $('#edit-content')[0]
-    )
+    if sd.EDIT_2
+      if channel.get('name') is 'Artsy Editorial'
+        # TODO - replace with better conditional (or channel.isEditorial),
+        # using name allows viewing the test channels in classic for interim.
+        @article.set('layout', 'standard')
```
kanaabe: Confused as to why we can't use `channel.isEditorial()`? Also this kind of backfill is very easy to do so instead of writing a conditional we should go ahead and make the change to `standard`.

eessex: This is for development happiness rather than production -- gives me an easy way to preview both layouts by viewing a test channel or the editorial channel (isEditorial returns true for the test channels). 

Yeah, we could go ahead and convert all editorial articles to standard now to avoid this.

kanaabe: As long as you're in a staging environment, I don't see why you can't develop classic on a partner or team channel instead? The test channel is meant to emulate Editorial as much as possible which is why it's the only other channel with an `editorial` type.

client/apps/edit/components/content2/sections/header/index.styl
```diff
@@ -0,0 +1,81 @@
+.edit-header
+  padding-top 60px
+  .public-DraftEditorPlaceholder-root
```
kanaabe: Is it possible to use an internal class then this rather than the third party's? It's less predictable if their underlying code changes. If not, no worries.

eessex: Not in this case -- is a built in class from draft

client/apps/edit/client.coffee
eessex: ah yeah i guess that makes sense too

eessex: Should I remove the set layout in anticipation of a backfill?

kanaabe:
Great!

Issue #1126, opened 2017-06-21
Unschedule articles


kanaabe:
No longer able to unschedule an article in Admin.

eessex:
Using the 'x' button in the date field sets scheduled_publish_at to null, the unschedule button was removed in the admin panel redesign

![screen shot 2017-06-22 at 4 21 17 pm](https://user-images.githubusercontent.com/1497424/27454134-e2fb1dda-5766-11e7-806d-5046289dbf89.png)

kanaabe:
Wonder what changes we can make for the UI to be less confusing. When I click X the publish date/time defaults to the current time which isn't really indicative of the article being unscheduled. Maybe if it's a draft, show no date/time _unless_ it's scheduled?

owendodd:
@me add button that unschedules

PR #1194, opened 2017-07-21
@kanaabe => Add schedule/unschedule button

eessex:
screen shot 2017-07-21 at 3 36 50 pm

Closes https://github.com/artsy/positron/issues/1190
Closes https://github.com/artsy/positron/issues/1126
- Adds button to schedule/unschedule/update publish time
- Uses moment-timezone to show all publish dates in NY time

### DIFF
--- a/client/apps/edit/components/admin/article/index.coffee
+++ b/client/apps/edit/components/admin/article/index.coffee
@@ -1,7 +1,7 @@
 React = require 'react'
 ReactDOM = require 'react-dom'
 _ = require 'underscore'
-moment = require 'moment'
+moment = require 'moment-timezone'
 sd = require('sharify').data
 { div, label, input, button } = React.DOM
 AutocompleteList = React.createFactory require '../../../../../components/autocomplete_list/index.coffee'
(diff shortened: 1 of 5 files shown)


kanaabe:
Looks good but ❌ . Also I'm thinking it might make more sense to use `moment().local()` for Writer. The only place we need to worry about displaying EDT/EST is for all Editorial articles on the front-end no matter which timezone you're in. For global teams/partners though, we'll be displaying local times, so it can be confusing if their articles in Writer are showing EDT/EST. Those on Editorial would see EDT/EST in Writer anyway if we use `moment.local()`. Does that make sense?

By the way if we use `moment().local()`, it's already available via the `date` method on the [article model](https://github.com/artsy/positron/blob/master/client/models/article.coffee#L60).

kanaabe:
Ok cool so two more things -- 1. we should remove the `moment-timezone` package entirely and 2. I mentioned this before, but the `moment(@props.article.get('published_at')).local()` methods can be condensed to use `article.date('published_at')` which does the same thing.

kanaabe:
👏

PR #1196, opened 2017-07-24
@craigspaeth => New articles bug

eessex:
Recent changes to the 'lead_paragraph' field introduced a bug that resulted in a hang when creating new articles. 
PR adds conditional to check if lead paragraph before assessing length.

### DIFF
--- a/client/apps/edit/components/layout/index.coffee
+++ b/client/apps/edit/components/layout/index.coffee
@@ -116,7 +116,7 @@ module.exports = class EditLayout extends Backbone.View
 
   getBodyText: =>
     @fullText = []
-    if @article.get('lead_paragraph').length
+    if @article.get('lead_paragraph')?.length
       @fullText.push @article.get('lead_paragraph')
     for section in @article.sections.models when section.get('type') is 'text'
       @fullText.push section.get('body')


kanaabe:
Going to merge this since it's urgent.

client/apps/edit/components/layout/index.coffee
```diff
@@ -116,7 +116,7 @@ module.exports = class EditLayout extends Backbone.View
 
   getBodyText: =>
     @fullText = []
-    if @article.get('lead_paragraph').length
+    if @article.get('lead_paragraph')?.length
```
craigspaeth: I wonder if on the server-side we should be stricter—either not allow nulls or not allow blank strings. Generally speaking on think I got a little `allow('', null)` heavy [in the schema](https://github.com/artsy/positron/blob/master/api/apps/articles/model/schema.coffee#L93) and we could be more strict about validation server-side to avoid the consumer apps from having to be extra defensive. e.g. In this case `allow(null).min(3)` would be a more appropriate way to describe "An article can have no lead paragraph, or a lead paragraph at-least 3 characters long". Not suggesting for this PR, but wanted to put that out as food for thought going forward.

kanaabe: Agree 100%! I definitely feel the `allow(null,"")` pain these days and it'd be great to address this as technical debt after/while we're doing this article overhaul. I'll make an issue in the meantime.

PR #1199, opened 2017-07-25
@damassi => Swap URL for webfont.artsy.net

kanaabe:
Use your webfont bucket instead of pointing to Force's CDN. Thanks for putting the bucket together! 💯


damassi:
Nice; I'll update Force with a `WEBFONT_URL` env variable as well (and merge on 🍏 )

PR #1205, opened 2017-07-26
@eessex => Fix scheduler

kanaabe:
Not sure if the stuff described in https://artsy.slack.com/archives/C04GGGDCM/p1500985858521608 is entirely solved by this --

In the old iteration of the scheduler, we set both the `scheduled_publish_at` and `published_at` as the same value when someone scheduled and article which meant that the `onPublish` field wouldn't set the `published_at`. Now that that's changed, we should update the API to not set a default `published_at` if it's scheduled. 

Additionally, we won't have to view drafts with the scheduler anymore because it uses the unix timestamp to generate a slug if it is a draft. This ensures that no two drafts created on the same day have the same slug.

### DIFF
--- a/api/apps/articles/model/save.coffee
+++ b/api/apps/articles/model/save.coffee
@@ -15,7 +15,7 @@ Article = require './index'
 artsyXapp = require('artsy-xapp').token or ''
 
 @onPublish = (article, cb) =>
-  unless article.published_at
+  unless article.published_at or article.scheduled_publish_at
     article.published_at = new Date
   @generateSlugs article, cb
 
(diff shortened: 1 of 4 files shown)


api/apps/articles/model/save.coffee
```diff
@@ -59,7 +59,9 @@ removeStopWords = (title) ->
   # Append published_at to slug if that slug already exists
   db.articles.count { slugs: slug }, (err, count) ->
     return cb(err) if err
-    slug = slug + '-' + moment(article.published_at).format('MM-DD-YY') if count
+    if count
+      format = if article.published then 'MM-DD-YY' else 'X'
+      slug = slug + '-' + moment(article.published_at).format(format)
```
eessex: 👍  to unix

kanaabe:
Going to merge this!

Issue #1204, opened 2017-07-26
Deprecate two fullscreen fields


kanaabe:
There's a `background_image_url` and a `background_url` (video) on the fullscreen header. These should be converted into a single `url` - the client can determine how to render based on the url and this way hero_sections have a consistent asset field.

PR #1206, opened 2017-07-26
@craigspaeth => Adds url field to hero_section

kanaabe:
Starts work for: https://github.com/artsy/positron/issues/1204 

After I backfill this field, I'll get support for Force, then remove the two fields we no longer need.


craigspaeth:
Sounds like a good plan to me!

PR #1209, opened 2017-07-27
@craigspaeth => Deprecate hero_section fields

kanaabe:
Closes https://github.com/artsy/positron/issues/1204

Ok last step! Remove the old fields. The `url` is already backfilled and Force is using this new `url` field too. After this, we just need to unset all the old fields in the database.


craigspaeth:
👍LGTM!

PR #1212, opened 2017-08-01
@kanaabe => Article 2 Text Component

eessex:
There is a lot of green from the rich_text component, which was duplicated as rich_text2 -- additions made to that directory are in the rich_text2/utils/index file, as well as  rich_text2/components/edit_nav and tests. 

Adds text component to Content2:
- Moves more functions into utils folder
- Moves menu into separate component
- Adds article2 styles for classic, standard, as well as partial feature styles
- Updates draft_config to specify block/style maps based on layout
- Allows or limits use of blockquote, h1, artist-link, h3 styling based on layout/channel
- Strips h3 styling based on layout

Standard Layout:
screen shot 2017-08-07 at 11 57 08 am

Classic Layout:
screen shot 2017-08-07 at 11 59 14 am


client/apps/edit/components/content/section_list/test/index.coffee
```diff
@@ -27,7 +28,7 @@ describe 'SectionList', ->
       @SectionList.__set__ 'SectionContainer', React.createFactory @SectionContainer
       @SectionList.__set__ 'DragContainer', React.createFactory DragContainer
       @props = {
-        sections: @sections = new Backbone.Collection [
+        sections: @sections = new Sections [
```
kanaabe: 😏

client/models/section.coffee
```diff
@@ -45,4 +45,12 @@ module.exports = class Section extends Backbone.Model
...
\ No newline at end of file
+      super
+
+   getLayout: ->
+    layout = 'column_width'
+    if @get('type') is 'text'
+      layout = 'overflow_fillwidth' if @get('body')?.includes('<blockquote>')
+    else if @get('layout')
+      layout = @get('layout')
+    return layout
```
kanaabe: I wonder if we'll want an additional layout for blockquote since it's a little larger than the overflow_fillwidth for ImageCollection:

image

This _could_ work if we adjust the width numbers of `overflow_fillwidth` per component, but I tend to feel like that ends up with a lot of CSS confusion especially with the way I'm building a new component in Reaction called `SectionContainer` that wraps every section component. SectionContainer knows about the following layout properties:

`column_width`has a max-width of 680px
`overflow_fillwidth` has 780px

kanaabe:
This looks great! Besides my last comment (which I'm curious of your thoughts), it'd be nice to see some screenshots here.

client/models/section.coffee
eessex: Oh I see- was mostly looking at the standard when building and didn't notice the detail- yeah probably easiest to make one more layout size

kanaabe: Ah I see, so for Standard it would be overflow_fillwidth. In that case it's not necessary for this PR to tackle the Standard/Feature blockquote.

eessex: I am also passing the type to the container, so we should be able to catch it in the feature CSS

PR #1213, opened 2017-08-01
@kanaabe => Add postscript field and UI

eessex:
- Adds postscript field to articles api
- Adds rich-text field for postscript to edit/content/section_list for editorial channels

screen shot 2017-08-07 at 1 56 47 pm

### DIFF
--- a/api/apps/articles/model/schema.coffee
+++ b/api/apps/articles/model/schema.coffee
@@ -141,6 +141,7 @@ denormalizedArtwork = (->
       layout: @string().allow('overflow_fillwidth', 'column_width', null)
       images: @array().items([denormalizedArtwork, imageSection])
   ]).allow(null)
+  postscript: @string().allow('', null)
   primary_featured_artist_ids: @array().items(@string().objectid()).allow(null)
   featured_artist_ids: @array().items(@string().objectid()).allow(null)
   featured_artwork_ids: @array().items(@string().objectid()).allow(null)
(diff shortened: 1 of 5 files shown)


kanaabe:
Will a postscript always exist in the body of an article? And do we only need a postscript to be able to designate where the proper "end" of an article is?

If so, we may want to move this into a real section to make article rendering for the client easier. This is kind of the same problem we've had with `lead_paragraph`. It's a field that should exist as part of the article, but it was never treated as so and has been forgotten about or ignored in the past. The way we render it in Writer doesn't necessarily have to drop-down in SectionList, but I think saving it as a real section in the backend will have benefits for us in the future. Thoughts?

eessex:
This makes sense to me - will look at moving this field into a section today.  

Re: lead paragraphs for editorial -- should be converted into a text section as well, allowing us to remove support altogether from the standard/feature layout. Another backfill for the switch...

eessex:
After looking at this a bit further I'm leaning towards keeping it separate. It is an optional field used only on editorial articles. The end of the article will continue to be determined by the last section, and this should come after. We don't want to render this in the sections_list in writer because it must always be the last section and shouldn't be drag-droppable. 

Because it is different than the text section, we'd also need to build out something similar to hero sections, adding another component as well as article model logic to moves it out of the sections.collection and add its own event hooks for changes, and then re-add it to the collection on save. Seems much more straightforward to optionally render the field on the client if it exists. Are there benefits to using sections that I am missing?

kanaabe:
Ok I see where you're coming from -- on the Positron side, you'd want to keep things orderly by rendering all sections in the sections list and prevent more event hooks/conditionals. From a modelling perspective I still think it makes more sense to treat it as a first class section since it is part of the body of an article, but it looks like it may be more trouble than it's worth. Just keep in mind that we'll need to update all the distribution channels and Force to reflect the new postscript field!

PR #1215, opened 2017-08-08
@craigspaeth => API bug with viewing draft & expose keywords for graphql

kanaabe:
1. Fixes a bug with [this line](https://github.com/artsy/positron/blob/master/api/apps/articles/routes.coffee#L8) where not passing any `published` param will result in skipping the error. Instead of fixing the conditional, we should give the `published` field a default value in the `querySchema` to prevent future mistakes like this.
2. Exposes the `keywords` field in the `inputSchema` (which is normally generated internally) so our GraphQL endpoint can view this field.

### DIFF
--- a/api/apps/articles/index.coffee
+++ b/api/apps/articles/index.coffee
@@ -5,7 +5,7 @@ routes = require './routes'
 app = module.exports = express()
 
 app.get '/articles', setUser, routes.index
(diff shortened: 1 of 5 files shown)


craigspaeth:
Pretty straightforward! LGTM, thanks!

PR #1221, opened 2017-08-12
Simplify development environment setup via Docker Compose

damassi:
This is still WIP as I need to test a few things out, but what this PR does is wrap all of our local development environment services (mongo, elasticsearch, node) into one command `docker-compose up`. This allows us to keep things contained when working locally and free of global installs. Future setup is along the lines of 

- Install Docker for Mac (https://docs.docker.com/docker-for-mac/install/)
- `docker-compose up`

Note that this only applies only to local development environments and does not change existing deployment configuration.

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -7,6 +7,7 @@ ARTSY_URL=http://localhost:5000/__gravity
 SESSION_SECRET=p0s1tr0n
 MONGOHQ_URL=mongodb://localhost:27017/positron-test
 ELASTICSEARCH_URL=http://localhost:9200
+ELASTICSEARCH_PORT=9200
 SALT=$2a$10$PJrPMBadu1NPdmnshBgFbe
 API_MAX=100
 API_PAGE_SIZE=10
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM node:7
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+COPY package.json .
+COPY . .
+RUN yarn install
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,18 @@
+version: '2'
+services:
+  elasticsearch:
+    image: elasticsearch:latest
+    ports:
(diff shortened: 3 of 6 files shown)


.env.test
```diff
@@ -7,6 +7,7 @@ ARTSY_URL=http://localhost:5000/__gravity
 SESSION_SECRET=p0s1tr0n
 MONGOHQ_URL=mongodb://localhost:27017/positron-test
 ELASTICSEARCH_URL=http://localhost:9200
+ELASTICSEARCH_PORT=9200
```
yuki24: This also needs to be added to `.env.example`.

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+version: '2'
+services:
+  elasticsearch:
+    image: elasticsearch:latest
+    ports:
+      - "${ELASTICSEARCH_PORT}:${ELASTICSEARCH_PORT}"
+  mongodb:
+    image: mongo:latest
```
yuki24: I was getting a connection error from mongo client. I might be missing something, but possibly the port for mongodb also needs to be forwarded and exposed like it is for elasticsearch?

```
ports:
  - 27017:27017
```

yuki24:
👏 great PR @damassi! I wish positron had this before I set that up...

docker-compose.yml
bhoggard: You should lock to a specific version. Production is running 3.0.6, so I would set it to

`mongo:3.0`

Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM node:7
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+COPY package.json .
+COPY . .
+RUN yarn install
```
bhoggard: To avoid docker cache invalidation, do `yarn install` before the second `COPY` command.

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+version: '2'
+services:
+  elasticsearch:
+    image: elasticsearch:latest
```
bhoggard: Again, lock to a specific version.

PR #1223, opened 2017-08-14
Add CoffeeLint

damassi:
This PR adds a default [CoffeeLint](http://www.coffeelint.org/) config that will evaluate Positrons CoffeeScript code in a similar way that ESLint evaluates JavaScript. Once configured within your editor, it will pick up the `coffeelint.json` file in the root. Catches all common syntax errors, etc. 

Editor plugins:
- https://github.com/slb235/vscode-coffeelint
- https://atom.io/packages/linter-coffeelint
- https://github.com/SublimeLinter/SublimeLinter-coffeelint

### DIFF
--- a/coffeelint.json
+++ b/coffeelint.json
@@ -1,0 +1,135 @@
+{
+  "arrow_spacing": {
+    "level": "ignore"
+  },
+  "braces_spacing": {
+    "level": "ignore",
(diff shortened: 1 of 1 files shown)


kanaabe:
👍 Thanks!

PR #1224, opened 2017-08-14
Add Babel Pipeline, Better ESLint

damassi:
This PR adds Babel to the flow, along with some common packages as taken from Force:

- babel-cli
- babel-plugin-inline-react-svg
- babel-plugin-module-resolver 
- babel-plugin-rewire
- babel-plugin-styled-components
- babel-plugin-transform-class-properties
- babel-plugin-transform-runtime
- babel-preset-es2015
- babel-preset-react
- babel-preset-stage-3

Notable items from above:
- Updates ESLint setup to follow what we have on Force, with `standard` and `standard-react` being the primary engines. 
- In React code, `.svg` file imports are automatically converted into components that you can use like so:  `<Icon />` 
- Modules written in JS can take advantage of absolute path imports -- e.g., `import baz from 'api/foo/bar/baz'`
- Similar to Force, rewrite for test code
- I know most component work is taking place in `reaction`, but I want to setup server-side react rendering and this, similar to force, is needed to pluck those styles out


index.js
```diff
@@ -0,0 +1,3 @@
+require('babel-core/register')
+require('coffee-script/register')
+require('./app')
```
damassi: Created this boot loader so that from the point of `./app` onwards we can write ES6.

index.js
```diff
@@ -0,0 +1,14 @@
...
+
+if (process.env.NODE_ENV === 'test') {
+  env('./.env.test')
+
+  // Other envs: development, staging, production
+} else {
+  env('./.env')
+}
+
+require('./boot')
```
damassi: Created this boot loader so that from the point of ./app onwards we can write ES6.

boot.js
```diff
@@ -0,0 +1,52 @@
...
+      process.send('listening')
+    }
+  })
+})
+
+// Crash if we can't get/refresh an xapp token
+artsyXapp.on('error', (error) => {
+  console.warn(error)
+  process.exit(1)
+})
```
kanaabe: 🙌

kanaabe:
👏 This is so awesome!

damassi:
😄

PR #1225, opened 2017-08-14
Implement server-side code hot-swapping on file change 

damassi:
This PR sets up code hot-swapping on the server without a full restart (but still requiring a refresh of the browser window). Crazy-fast development speed! 

**Overview of process**:
- `reloadable` is called with a path to an app (`client`, `api`, etc) as well as a callback function that is executed whenever files with that appPath change
- When a source-code change is detected an internal lookup is made to Node, scanning its `require` cache
- If found, it is cleared via `delete require.cache[id]`
- When the browser is reloaded the callback re-requires the initial app and since the module is no longer in the cache node re-caches it, effectively hot-reloading the code during runtime. 

**Example**:

```js
import { createReloadable, isDevelopment } from 'lib/reloadable'

... 

if (isDevelopment) {
  const reloadAndMount = createReloadable(app)

  // Note that if you need to mount an app at a particular root (`/api`), pass 
  // in `mountPoint` as an option.
  app.use('/api', reloadAndMount(path.resolve(__dirname, 'api'), {
    mountPoint: '/api'
  }))

  // Otherwise, just pass in the path to the express app and everything is taken care of 
  reloadAndMount(path.resolve(__dirname, 'client'))
} else {
  app.use('/api', require('./api')
  app.use(require('./client')
}
```


boot.js
```diff
@@ -50,3 +52,11 @@ artsyXapp.on('error', (error) => {
...
   process.exit(1)
 })
+
+// TODO: Possibly a terrible hack to not share `req.user` between both.
+const doNotShareUserHack = (app) => {
+  app.use((req, rest, next) => {
+    req.user = null
+    next()
+  })
+}
```
kanaabe: Ah right this comment can probably be removed since it's pretty old and we actually benefit from not sharing req.user because it keeps our API stateless.

lib/reloadable.js
```diff
@@ -0,0 +1,76 @@
...
+
+      // Node env not 'development', exit
+    } else {
+      throw new Error(
+        '(lib/reloadable.js) NODE_ENV must be set to "development" to use ' +
+        'reloadable.js'
+      )
+    }
+  }
+}
```
kanaabe: Love this! Very clear and thoughtful error handling

boot.js
```diff
@@ -16,23 +18,31 @@ app.use(
...
 
-  // TODO: Possibly a terrible hack to not share `req.user` between both.
-  app.use((req, rest, next) => {
-    req.user = null
-    next()
-  })
+    // Enable server-side code hot-swapping on change
+    app.use('/api', reloadAndMount(path.resolve(__dirname, 'api'), {
+      mountPoint: '/api'
+    }))
```
kanaabe: 👍

boot.js
damassi: Oh yeah, gotta fix this

boot.js
```diff
@@ -50,3 +61,10 @@ artsyXapp.on('error', (error) => {
   console.warn(error)
   process.exit(1)
 })
+
+const invalidateUserMiddleware = (app) => {
```
kanaabe: 👍

kanaabe:
This is 🔥

PR #1226, opened 2017-08-15
Add Babel on the client

damassi:
Missed this from yesterday! Now you can use Babel on the client as well as the server, with everything being coordinated around the main `.babelrc` file. 

Additionally upgraded `browserify-dev-middleware` to latest; it was a major version out of date.


yuki24:
Briliant

PR #1227, opened 2017-08-15
Always use yarn 0.27.x on heroku, stable version as of today

yuki24:
Heroku picks up ` Downloading and installing yarn (0.28.1)...` which is a nightly build. We should be using a stable version.


kanaabe:
🙏  Thank you!

PR #1230, opened 2017-08-15
Add server-side React renderer 

damassi:
[WIP]

This PR moves over our `renderReactLayout` utility so that Positron can take advantage of server-side React, and move away from Jade. 

**NOTE:** This utility will be moved shortly into its own `@artsy/render-react-layout` (or something) NPM package; will remove and replace when ready.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -16,7 +16,7 @@
   "env": {
     "development": {
       "plugins": [
-        ["styled-components", {
+        ["babel-plugin-styled-components", {
           "ssr": true
         }]
       ]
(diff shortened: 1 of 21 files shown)


kanaabe:
Cool! I'm fine with merging this, but there's a part of me that thinks we should just wait until it's packaged or actually need it. In general, I'm not too worried about supporting SSR in Positron -- it makes sense for SEO or Performance in end-user-facing apps but for a CMS it's not a huge deal. Let me know if I'm missing something or if we're actually getting a big performance gain. I do think it's cool to have, but it also feels like more code to maintain/think about when there's already a plan to package it soon.

damassi:
@kanaabe - Starting to agree. There's already been a tiny bit of drift between this and Force which should be unnecessary once the updates I plan to make are made. Will close for now and reopen when ready!

PR #1231, opened 2017-08-15
Fix an 'Outdated Yarn lockfile' error on heroku

yuki24:
It seems like this change was lost while we were making changes to our setup. This also should fix [the deploy issue on heroku](https://dashboard.heroku.com/apps/positron-staging/activity/builds/3d848395-f40c-4d8b-add7-8557ced1cde1).


yuki24:
closing since https://github.com/artsy/positron/pull/1232/files also should fix the issue.

PR #1232, opened 2017-08-15
Setup 💅  StyledComponents 💅 

damassi:
@yuki24 and I paired on this and were running into an issue with `process is not defined` when trying to import `styled-components`. Turned out we needed to add `envify` (similar to Webpack's `DefinePlugin`) to preprocess node ENV vars. 

Added: 
- `styled-components`
- https://github.com/LoicMahieu/react-styled-flexboxgrid, which is a really sweet API port of Bootstraps grid system built in flexbox


yuki24:
🎉

PR #1234, opened 2017-08-15
WIP: Promoted content admin

yuki24:
### 🚧🚧 This is still WIP 🚧🚧

 * The canvas section is not yet implemented
 * Some styles are missing

screen shot 2017-08-15 at 7 40 46 pm


client/apps/settings/client/curations/promoted_admin.jsx
```diff
@@ -0,0 +1,153 @@
...
+                <Col lg={true}>
+                  <label>Title</label>
+                  <input
+                    className="bordered-input"
+                    placeholder="Title"
+                    defaultValue={section.name}
+                    onChange={ ((e) => this.onInputChange(e, index)).bind(this) }
+                    name="name" />
+                </Col>
+                <Col lg={true}>
```
damassi: @yuki24 - for props that are truthy you can omit the value: 

```
<Col lg>
  ...
</Col>

File: client/components/character_limit/index.jsx [hunk 2]
```diff
@@ -0,0 +1,37 @@
...
+      defaultValue: this.props.defaultValue,
+      onChange: this.onInputChange.bind(this),
+      name: this.props.name,
+      maxLength: this.props.limit
+    }
+
+    return (
+      <div>
+        <label>{this.props.label}</label>
+        <div>{this.state.remainingChars} Charactors</div>
```
eessex: Typo on Characters here

client/apps/settings/client/curations/promoted_admin.jsx
```diff
@@ -0,0 +1,152 @@
...
+            <div className="promoted-admin__section">
+              <Row key={index}>
+                <Col lg>
+                  <label>Title</label>
+                  <input
+                    className="bordered-input"
+                    placeholder="Title"
+                    defaultValue={section.name}
+                    onChange={ ((e) => this.onInputChange(e, index)).bind(this) }
+                    name="name" />
```
eessex: Since the onChange events are repeated a few times each, it might make sense to move these into the constructor-- an [example here](https://github.com/eessex/positron/blob/9b458e2c779b2c8390915da3b27f50bba9a7b7f5/client/components/file_input/index.jsx#L8).

client/apps/settings/client/curations/promoted_admin.jsx
```diff
@@ -0,0 +1,152 @@
...
+
+  onInputChange(e, i) {
+    return this.onChange(e.target.name, e.target.value, i)
+  }
+
+  onChange(key, value, index) {
+    const newCuration = this.state.curation.clone()
+    const section = newCuration.get("sections")[index]
+    set(section, key, value)
+
```
eessex: You can also set the curation sections directly here

client/apps/settings/stylesheets/promoted_admin.styl
```diff
@@ -0,0 +1,18 @@
+.promoted-admin
+  &__section-header
+    display flex
+    align-items center
+    border-bottom 1px solid gray-color
+    margin-top 30px
+    padding-bottom 10px
+    h1
+      Garamond()
+      font-size 30px
```
eessex: You can combine the first two lines of h1 by using `garamond-size('headline')`

client/apps/settings/stylesheets/promoted_admin.styl
```diff
@@ -0,0 +1,18 @@
...
+      padding-right 20px
+    .dropdown-header__caret
+      display flex
+  &__section_title
+    Garamond()
```
eessex: minor -- we usually do the classname with hyphens rather then underscore, as `&__section-title`

client/apps/settings/client/curations/promoted_admin.jsx
damassi: Or even better, [omit constructors completely](https://github.com/eessex/positron/blob/master/.babelrc#L14).

Since we're using ES6 and fat-arrow functions bind to the outer scope automatically, you can do this instead: 

```js
class Home extends Component {
  state = {
    count: 0
  }

  handleClick = () => {
    const count = count + 1

    this.setState({
      count 
    })
  }

  render () {
    return (
      <div onClick={this.handleClick}>
        Click!
      </div>
    )
  }
}
```

No more need to use constructors at all except in cases where you need to set the state via props.

PR #1238, opened 2017-08-16
Make undefined check safe

damassi:
Noticed in the previous PR that there was an equality check against `undefined` but JS is super weird and that kind of operation isn't safe. A best practice is do `typeof myVar === 'undefined'`, or use underscore or lodash's `_.isUndefined`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,15 +1,19 @@
 require('babel-core/register')
 require('coffee-script/register')
-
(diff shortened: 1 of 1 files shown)


kanaabe:
Nice catch! Thanks

yuki24:
@damassi looks great! Thanks for catching this 👍

PR #1245, opened 2017-08-22
Autolink artists

kanaabe:
Hackathon project


ashkan18:
after talking to @cavvia this morning, looks like we can do something similar to https://github.com/cavvia/gravity/blob/master/app/reports/duplicate_artists_report.rb#L160 i'll try it tomorrow.

api/apps/search/queries.coffee
```diff
@@ -0,0 +1,19 @@
...
+            fuzziness: 2
+          }
+        }
+        should: {
+          match_phrase: {
+            name: term
+          }
+        }
+      }
+    }
```
cavvia: :+1: nice thing about json queries is you can just copy them between client apps.

client/apps/edit/components/layout/index.coffee
```diff
@@ -133,3 +136,50 @@ module.exports = class EditLayout extends Backbone.View
...
+
+          @article.replaceLink(findText, @getNewLinkFromHits(valid_results) || text)
+          return cb()
+    ), (err, result) =>
+      @article.sections.trigger 'change:autolink'
+      $('#autolink-status').removeClass('searching').html('Auto-Link')
+      $('#edit-content__overlay').removeClass('disabled')
+
+  findValidResults: (hits) ->
+    _.reject(hits, (h) -> h._score < 6.5 || h._source.visible_to_public == false)
```
cavvia: :+1: 6.5 is the threshold we use for detecting artist duplicates. You might get away with lowering it slightly for fuzzier matches, but this should work pretty well.

cavvia:
Looks like a great feature!

api/apps/search/queries.coffee
```diff
@@ -0,0 +1,19 @@
+module.exports =
+  matchAll: (term) ->
+    {
+      bool: {
+        must: {
+          multi_match: {
+            query: term,
+            type: 'best_fields',
+            fields: ['name.*', 'alternate_names.*'],
+            fuzziness: 2
```
cavvia: As I mentioned, `AUTO` will default to 2 for strings of more than five characters, as per [the doc](https://www.elastic.co/guide/en/elasticsearch/guide/current/fuzziness.html). But forcing it to two is good. You can't go higher on ES 2.x afaik.

api/apps/search/routes.coffee
```diff
@@ -0,0 +1,13 @@
...
+queries = require './queries.coffee'
+
+# GET /api/search
+@index = (req, res, next) ->
+  search.client.search(
+      body:
+        query: queries.matchAll(req.query.term)
+      , (error, response) ->
+        return console.log("nothing matched for: #{req.query.term}") if error
+        res.send response.hits
```
cavvia: I assume since you don't specify indexes here, this will hit all the indexes. This could potentially be constrained using the [index parameter](https://www.elastic.co/guide/en/elasticsearch/reference/2.1/multi-index.html).

ashkan18: i tried to limit it to specific types, but then the list got long enough, would this be a performance hit?

api/apps/search/queries.coffee
ashkan18: 👍

client/apps/edit/components/layout/index.coffee
ashkan18: I think less fuzzy works better in our case, we replace existing term with what we received from search results and i can imagine not creating the link would be better than replacing text with wrong result. I should still look into scoring system, how did we came up with 6.5?

api/apps/search/routes.js
```diff
@@ -0,0 +1,17 @@
+import search from '../../lib/elasticsearch.coffee'
```
damassi: @kanaabe - you've now got [absolute URL ability](https://github.com/artsy/positron/blob/master/.babelrc#L12); I would use it! It makes refactoring, restructuring and readability way easier over time:

`import search from 'lib/elasticsearch.coffee'`

cc/ @eessex

Makefile
```diff
@@ -19,6 +19,7 @@ sf:
 # Run all of the project-level tests, followed by app-level tests
 test: assets
 	$(BIN)/mocha $(shell find api -name '*.test.coffee')
+	$(BIN)/mocha $(shell find api -name '*.test.js')
 	$(BIN)/mocha $(shell find client -name '*.test.coffee')
```
eessex: will 22 conflict with 24?  we have jest running on all the .js tests at the moment

eessex: oh nvmd i see it is the api side

client/apps/edit/components/content/sections/text/index.coffee
```diff
@@ -48,6 +48,7 @@ module.exports = React.createClass
     @hasFeatures = @hasFollow
 
   componentDidMount: ->
+    @props.sections.on 'change:autolink', @editorStateFromProps
     if @props.section.get('body')?.length
```
eessex: Sorry should have realized this earlier -- we are deprecating `client/apps/edit/components/content/` in favor of `/content2` very soon, is there a way to port this into the new text component? Maybe in a later PR?

kanaabe: Yep I am aware and working on that part now :)

eessex: awesome!

kanaabe:
Don't think there's anything left to address now -- @ashkan18 want to look over it one more time and merge if 👍 ?

client/apps/edit/components/layout/index.coffee
```diff
@@ -133,3 +136,50 @@ module.exports = class EditLayout extends Backbone.View
...
+    result = results[0]
+    name = result._source.name
+    link = @findLinkFromResult(result)
+    "<a href='#{link}'>#{name}</a>"
+
+  findLinkFromResult: (result) ->
+    switch result._type
+      when "artist" then "https://artsy.net/artist/#{result._source.slug}"
+      when "partner" then "https://artsy.net/#{result._source.slug}"
+      when "city" then "https://artsy.net/shows/#{result._source.slug}"
```
ashkan18: we can probably get `https://artsy.net/shows` to a config if we don't have it as a config. We can do that in separate PR though. when i added that i wasn't sure if there is config for it.

ashkan18:
Had one small comment we can fix in follow up PR! going to merge this since im really excited as you can see!!!!

client/apps/edit/components/layout/index.coffee
kanaabe: Oh true, "https://artsy.net/" can be replaced with the already existing sd.FORCE_URL

ashkan18: Im doing this in a separate PR

client/apps/edit/components/layout/index.coffee
cavvia: I think 6.5 was just by manual observation running the duplicate artist report and seeing if they were real dupe candidates that it was picking up on.

PR #1246, opened 2017-08-22
@kanaabe => Consolidate section layout controls

eessex:
Moves layout icon-toggles into section-container component, so we don't have to repeat across components.


eessex:
Will do the next PR in es6 👍

kanaabe:
Cool, love this idea!

PR #1248, opened 2017-08-23
@eessex => Alternate titles and descriptions in s3 export

kanaabe:
Pass some more fields to the daily s3 export


eessex:
does it make sense to add the new postscript field as well?

PR #1249, opened 2017-08-24
@damassi => [Minor] Compile and look for JS tests

eessex:
Following on our conversation yesterday, includes js tests in the compile process, and looks for them when running yarn test.


damassi:
👍

damassi:
@eessex - I keep trying to hit that rebuild button because the test failures seem spurious, but it keeps going red 😞

eessex:
Yeah has been failing somewhat randomly on every pr lately- going to try and target the culprit today

PR #1253, opened 2017-08-29
@craigspaeth => Fix GraphQL query for objects with no isTypeOf option

kanaabe:
There are a few more sections missing the `isTypeOf` method that's currently breaking GraphQL queries on sections.


craigspaeth:
Pretty straightforward, thanks!

PR #1255, opened 2017-08-29
@kanaabe => Image collection updates

eessex:
- Replaces ImageCollection's `components/artwork` and `components/image` with jsx
- Adds reaction components to ImageCollection for Remove icon, ImageSetPreviews and Image
- Adds new fileUpload component to controls
- Adds size limit (and error alert if over) to fileUpload
- Gets rid of image-specific functions in sectionControls


client/apps/edit/components/content2/sections/image_collection/index.coffee
```diff
@@ -126,25 +134,31 @@ module.exports = React.createClass
...
-                    index: i
-                    key: i
-                    image: item
-                    removeItem: @removeItem
-                    editing:  @props.editing
-                    dimensions: @state.dimensions
-                    imagesLoaded: @state.imagesLoaded
-                  }
+                  React.createElement(
+                    Image.default, {
```
kanaabe: Kinda strange that you'd have to reference the default here. It looks like your image and artwork does export the class as default so I wonder why it needs to be referenced like this?

client/apps/edit/components/content2/sections/image_collection/test/artwork.test.js
```diff
@@ -0,0 +1,64 @@
...
+
+  it('calls removeItem when clicking remove icon', () => {
+    props.editing = true
+    const wrapper = mount(
+      <Artwork {...props} />
+    )
+    wrapper.find('.edit-section__remove').simulate('click')
+    expect(props.removeItem).toBeCalled()
+  })
+})
```
kanaabe: Nice tests!

client/apps/edit/components/content2/sections/image_collection/index.coffee
eessex: Yeah have been toying with it but is the only way I've gotten the reaction components to work in coffeescript- otherwise I get an error that it received a string. In es6 components we do not need the .default to import

client/components/file_input/index.jsx
```diff
@@ -5,10 +5,25 @@ const sd = require('sharify').data
...
+    } else {
+      return 30000000
+    }
+  }
+
+  getSizeError() {
+    const size = this.props.sizeLimit ? this.props.sizeLimit.toString() : '30'
+    return 'Error: Size limit is ' + size + 'MB, selected file is too large.'
+  }
+
```
kanaabe: 👏  Thank you! I noticed that we didn't have this here the other day.

kanaabe:
Nice work! Looks like you're getting into a good flow with new layouts/components/es6.

PR #1261, opened 2017-09-01
Boost name over alternative names, boost index

ashkan18:
# Problem
`==Museum of Modern Art==` is replaced with Arab Museum of Modern Art
`==Kara Walker==` is replaced with nothing

# Solution
Add boost to name match over `alternate_names`, this fixes modern art issue.

Add `indices_boost` to our query which seems to fix Kara Walker issue since artist result comes first.

### DIFF
--- a/api/apps/search/queries.js
+++ b/api/apps/search/queries.js
@@ -5,7 +5,7 @@ export const matchAll = (term) => {
         multi_match: {
           query: term,
           type: 'best_fields',
-          fields: ['name.*', 'alternate_names.*'],
+          fields: ['name.*^4', 'alternate_names.*'],
           fuzziness: 2
         }
       },
--- a/api/apps/search/routes.js
+++ b/api/apps/search/routes.js
@@ -5,7 +5,7 @@ const { NODE_ENV } = process.env
 
 // GET /api/search
 export const index = (req, res, next) => {
-  const env = NODE_ENV === 'production' ? 'production' : 'production'
+  const env = NODE_ENV === 'production' ? 'production' : 'staging'
   let index = ''
   if (req.query.type) {
     index = _.map(req.query.type.split(','), term => {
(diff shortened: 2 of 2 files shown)


api/apps/search/routes.js
```diff
@@ -5,7 +5,7 @@ const { NODE_ENV } = process.env
 
 // GET /api/search
 export const index = (req, res, next) => {
-  const env = NODE_ENV === 'production' ? 'production' : 'staging'
+  const env = NODE_ENV === 'production' ? 'production' : 'production'
```
kanaabe: Should this be changed back to staging?

api/apps/search/queries.js
```diff
@@ -5,7 +5,7 @@ export const matchAll = (term) => {
         multi_match: {
           query: term,
           type: 'best_fields',
-          fields: ['name.*', 'alternate_names.*'],
+          fields: ['name.*^4', 'alternate_names.*'],
```
kanaabe: Oo what does this do?

api/apps/search/routes.js
```diff
@@ -15,6 +15,11 @@ export const index = (req, res, next) => {
   search.client.search({
     index,
     body: {
+      indices_boost: {
+        artists_production: 4,
```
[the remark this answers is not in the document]
kanaabe: 👍  Fine with me

api/apps/search/queries.js
ashkan18: this boosts, `name` field over `alternate_names` in case of Museum of Modern Art, alternate name matching was causing Arab Museum of Modern Art to show up first.

api/apps/search/routes.js
ashkan18: yes, sorry, fixed that, wanted to try with production locally.

PR #1263, opened 2017-09-05
Server side debugging

ashkan18:
@orta showed me how we can do server side debugging using chrome developer tools which seemed pretty cool! 

I added a new task `make dev` to enable server side debugging and added some docs about how to use it. We could also make that part of `make s` let me know which one makes more sense.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,10 @@
 
 BIN = node_modules/.bin
 
+# Start the server with inspect
+dev:
+	DEBUG=app,client,api node --inspect ./index.js
+
 # Start the server
 s:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -8,6 +8,10 @@
 
 BIN = node_modules/.bin
 
+# Start the server with inspect
+dev:
+	DEBUG=app,client,api node --inspect ./index.js
+
```
kanaabe: 🙌  woohoo!

kanaabe:
😎  Thanks @ashkan18 this is great!

PR #1264, opened 2017-09-06
@kanaabe => Plain text component

eessex:
A Draft JS component that creates plain text -- for use instead of textarea for article titles, decks and potentially other places.
 
Why? A `textarea` does not change its size based on the content, meaning a long title could overflow its container, or a short title could be rendering with extra space, changing the layout from what would be expected in force.


kanaabe:
LGTM! Nice set of tests + thanks for the explanation. Feel free to merge when green

PR #1270, opened 2017-09-11
[GraphQL] Add script to dump schema.

alloy:
Using this script you can do `mkdir schema && ./node_modules/.bin/babel-node scripts/dump-schema.js ./schema` and get a result like [this one](https://gist.github.com/alloy/89cc695e4f746bc6f569c92977d9f4a5).

Dunno if you want to merge this at all or if it’s just useful while cleaning up the schema.


kanaabe:
Cool! Thanks for the handy script.

PR #1272, opened 2017-09-11
@kanaabe => Clean up failing/pending tests

eessex:
Fixes tests that were failing intermittently on Circle, and brings the number of pended tests down to 3. 

Also adds a conditional to the User model to prevent [this error](https://sentry.io/artsynet/positron-production/issues/284826139/) from Sentry.


kanaabe:
Awesome job!!

PR #1275, opened 2017-09-11
@eessex => [API] Omit query in articles

kanaabe:
Allows you to pass an array of article ids to omit from an articles retrieval.


eessex:
Does this have the new test stuff merged into it?  Just sent a PR that went green on the first try, but looks like the drag and drop stuff is still unhappy here.

PR #1276, opened 2017-09-12
@kanaabe => Fix Edit rendering error

eessex:
Pages with the new hero_section types are causing errors in the old edit page.  This PR adds a conditional to exclude 'text' and 'split' layouts from rendering in the old layout.


kanaabe:
Thanks!

PR #1277, opened 2017-09-13
Safeguard tests from using .env

kanaabe:
This PR safeguards against a bug where importing/requiring [this test helper](https://github.com/artsy/positron/blob/master/api/test/helpers/db.coffee) _after_ importing/requiring the app will cause the helper to use your .env file instead of .env.test.

### Problem
To replicate the issue, require the `app` before the helper. (Don't try this now unless you're pointing to a local database in your .env)
```
app = require 'api/index.coffee'
{ empty } = require 'api/test/helpers/db.coffee'
```
This causes `app` to be loaded with no `process.env.NODE_ENV` set which means [this line](https://github.com/artsy/positron/blob/master/api/index.coffee#L1) gets executed and all subapps get loaded with your local `.env` settings. If you swap the order of the require, it behaves correctly. This is why:
```
{ empty } = require 'api/test/helpers/db.coffee'
    # in this file, we run:
    # env path.resolve(__dirname, '../../../.env.test')
    # which sets process.env.NODE_ENV
app = require 'api/index.coffee'
    # now that we have process.env.NODE_ENV,
    # https://github.com/artsy/positron/blob/master/api/index.coffee#L1
    # no longer is executed and the subapps use the correct environment
```

We've been miraculously getting by without issues until now, when I accidentally wiped out staging yesterday when running a test locally with the problematic order.

### Solution

We need to ensure that our `NODE_ENV` in a test setting is working properly. By setting the `NODE_ENV` on the `mocha.opts` level, we can ensure that every test that runs with the right environment. The `jest` settings don't need to be updated since this only affects API work.

1. Ensure that we're passing in `NODE_ENV=test` on all mocha tests
2. Update `api/index.coffee` to switch environments based on `NODE_ENV`.

cc @joeyAghion @eessex @damassi


Makefile
```diff
@@ -22,10 +22,10 @@ sf:
...
 # Run all of the project-level tests, followed by app-level tests
 test: assets
-	$(BIN)/mocha $(shell find api -name '*.test.coffee')
-	$(BIN)/mocha $(shell find api -name '*.test.js')
-	$(BIN)/mocha $(shell find client -name '*.test.coffee')
-	$(BIN)/jest $(shell find client -name '*.test.js')
+	NODE_ENV=test $(BIN)/mocha $(shell find api -name '*.test.coffee')
+	NODE_ENV=test $(BIN)/mocha $(shell find api -name '*.test.js')
+	NODE_ENV=test $(BIN)/mocha $(shell find client -name '*.test.coffee')
+	NODE_ENV=test $(BIN)/jest $(shell find client -name '*.test.js')
```
kanaabe: I don't think this is necessary since mocha will pick up on `test/mocha.opts` but JIC..

craigspaeth:
While we're looking at the env file, maybe it'd be better to use [dotenv](https://github.com/motdotla/dotenv#preload) to preload the separate env filed via `mocha -r dotenv/config dotenv_config_path=env.test` and `node -r dotenv/config`. I _think_ that'd solve this as well as removing the `switch` boilerplate.

kanaabe:
@craigspaeth you're right, that's a much cleaner solution 👍

craigspaeth:
Nice! Thanks for embarking on this!

PR #1279, opened 2017-09-13
@craigspaeth => [API] Related articles GraphQL, better omit query, and decaffeinate

kanaabe:
- Adds a `relatedArticles` query to our GraphQL `article` and `articles` resources
- Improve the `omit` query by allowing slugs + objectids
- Decaffeinate GraphQL subapp

PR #1280, opened 2017-09-14
@craigspaeth => Add relatedArticles resolvers

kanaabe:
A follow-up to https://github.com/artsy/positron/pull/1279

I realized the PR above was a bit naive because we need to use the result of the initial articles fetch to query for related articles. We want the query to look like this:
```
{
  articles(published:true, featured: true, limit: 3){
    title
    id
    relatedArticlesCanvas {
      thumbnail_title
    }
    relatedArticlesPanel {
      thumbnail_title
    }
  }
}
```

This allows us to reuse `root` in the `relatedArticlesCanvas` and `relatedArticlesPanel` resolvers which now has the article data. From there we can grab things like `vertical.id` and `tags` and feed it into the subsequent `relatedArticles` queries. 
 
cc @eessex

### DIFF
--- a/api/apps/graphql/resolvers.js
+++ b/api/apps/graphql/resolvers.js
@@ -1,9 +1,11 @@
+import _ from 'underscore'
 const User = require('api/apps/users/model.coffee')
 const Curation = require('api/apps/curations/model.coffee')
 const Channel = require('api/apps/channels/model.coffee')
 const Tag = require('api/apps/tags/model.coffee')
 const Author = require('api/apps/authors/model.coffee')
 const { mongoFetch, present, presentCollection, find } = require('api/apps/articles/model/index.coffee')
+const { ObjectId } = require('mongojs')
 
 export const articles = (root, args, req, ast) => {
   const unpublished = !args.published || args.scheduled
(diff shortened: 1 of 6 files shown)


api/apps/graphql/resolvers.js
```diff
@@ -93,3 +95,45 @@ export const authors = (root, args, req, ast) => {
...
+    featured: true,
+    channel_id: ObjectId(root.channel_id),
+    tags: tags,
+    limit: 3,
+    sort: '-published_at'
+  }
+  return new Promise((resolve, reject) => {
+    mongoFetch(_.pick(args, _.identity), (err, results) => {
+      if (err) {
+        reject(new Error(err))
```
craigspaeth: FYI it's typically good to do `return reject(new Error(err))` here because to prevent the code below from executing. Although this is not [necessary in this case](https://stackoverflow.com/questions/32536049/do-i-need-to-return-after-early-resolve-reject), just FYI. I won't block this PR for it, b/c if you do want to address this I'd prefer updating `mongoFetch` to return a promise allowing async/await so you can do...

````javascript
export const relatedArticlesPanel = async (root) => {
  const args = { ... }
  const { results } = await mongoFetch(...)
  return results
}
````

craigspaeth:
Another nitpick about the async/await opportunity, but otherwise, code LGTM and makes sense. Thanks!

api/apps/graphql/resolvers.js
kanaabe: Sounds good, I'll follow up with the modernized `mongoFetch` next!

PR #1290, opened 2017-09-20
@craigspaeth => Admin: Related articles, layout toggle

eessex:
- Adds related_articles field to the Article schema

In Admin panel:
- Adds UI for toggling article layout on editorial channels
- Adds autocomplete for adding related articles in editorial channels
- Adds confirm warning data can be lost when changing from layout to feature

screen shot 2017-09-20 at 11 13 52 am

### DIFF
--- a/api/apps/articles/model/schema.coffee
+++ b/api/apps/articles/model/schema.coffee
@@ -159,7 +159,7 @@ denormalizedArtwork = (->
       images: @array().items([denormalizedArtwork, imageSection])
   ]).allow(null)
   postscript: @string().allow('', null)
-  related_articles: @array().items(@string().objectid()).allow(null)
+  related_article_ids: @array().items(@string().objectid()).default([])
   primary_featured_artist_ids: @array().items(@string().objectid()).allow(null)
   featured_artist_ids: @array().items(@string().objectid()).allow(null)
   featured_artwork_ids: @array().items(@string().objectid()).allow(null)
--- a/api/apps/articles/model/schema.coffee
+++ b/api/apps/articles/model/schema.coffee
@@ -159,6 +159,7 @@ denormalizedArtwork = (->
       images: @array().items([denormalizedArtwork, imageSection])
   ]).allow(null)
   postscript: @string().allow('', null)
+  related_articles: @array().items(@string().objectid()).allow(null)
   primary_featured_artist_ids: @array().items(@string().objectid()).allow(null)
   featured_artist_ids: @array().items(@string().objectid()).allow(null)
   featured_artwork_ids: @array().items(@string().objectid()).allow(null)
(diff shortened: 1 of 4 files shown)


api/apps/articles/model/schema.coffee
```diff
@@ -159,6 +159,7 @@ denormalizedArtwork = (->
       images: @array().items([denormalizedArtwork, imageSection])
   ]).allow(null)
   postscript: @string().allow('', null)
+  related_articles: @array().items(@string().objectid()).allow(null)
```
craigspaeth: It might be more appropriate to call this `related_article_ids` for clarity, and remove `.allow(null)` in favor of `.default([])`.

I realized a design mistake I made early in Positron's data modeling was the liberal use of `.allow(null)`, which can lead to unnecessary defensiveness on clients down the line (there's actually an example of this [just below](https://github.com/artsy/positron/pull/1290/files#diff-bec8b0f3e27f5cf948898309333d3ed4R19)). I realized later that a better mindset to have here is to default to "empty" values instead of null, and only allow null if it truly makes sense to distinguish between the field being "empty" and the field being missing entirely.

e.g. An article can be related to zero `[]` or many articles `[id, id]`, but there isn't really a third state we need to consider that describes "missing the concept of related articles"—so probably better to not allow that. Conversely, in the case of `scheduled_publish_at `, an article can either be scheduled to publish at a certain date, or _not_ scheduled at all `null`. Admittedly, this is a very subjective/case-by-case thing but generally I'd like to see Positron default to empty values like `[]`, `''`, `false`, etc. before considering allowing null. [Related Stack Overflow question](https://dba.stackexchange.com/questions/5222/why-shouldnt-we-allow-nulls).

THAT said, I won't hold this PR up for it as this alludes to a bigger effort across Positron's schemas that I'd be happy to see addressed wholesale later.

client/apps/edit/components/admin/article/index.coffee
```diff
@@ -39,6 +45,13 @@ module.exports = AdminArticle = React.createClass
     @setState featured: featured
     @onChange 'featured', featured
 
+  onLayoutChange: (e) ->
+    if e.target.name is 'standard' and @props.article.get('layout') is 'feature'
+      canLooseData = confirm 'Some header and section layout data may be lost. Change anyways?'
```
craigspaeth: `canLooseData` > `canLoseData`

client/apps/edit/components/admin/test/article.coffee
```diff
@@ -94,6 +98,19 @@ describe 'AdminArticle', ->
       @component.showActive('tier', 2).should.eql ' active'
       @component.showActive('featured', true).should.eql ''
 
+  describe 'Display: Editorial features', ->
```
craigspaeth: 👌  solid tests as always.

craigspaeth:
Overall LGTM! I left some small comments that you can feel free to address or punt on for this PR. Well done! Feel free to merge when you're ready.

api/apps/articles/model/schema.coffee
eessex: This makes sense to me- going to update to incorporate your comments

PR #1297, opened 2017-09-25
@craigspaeth => Hero section

eessex:
This PR adds a HeroSection component for use on Classic layout articles (in practice, used in internal Team Channels). It also moves towards deprecating the Image component, continuing the trend to es6, and removes some legacy logic specific to Classic layouts.

- Adds ImageCollection to schema for HeroSection options
     * TODO: Remove ImageSection & ArtworkSection entirely (after backfill/articles switch)
- Refactors SectionTool component in es6
- Removes Image component from SectionContainer
- Removes positioning logic for classic articles from SectionControls


client/apps/edit/components/content2/section_tool/index.jsx
```diff
@@ -0,0 +1,146 @@
...
+
+const IconHeroImage = components.Icon.HeroImage
+const IconHeroVideo = components.Icon.HeroVideo
+
+export default class SectionTool extends Component {
+  constructor (props) {
+    super(props)
+
+    this.type = {displayName: 'SectionTool'}
+    this.state = {
```
eessex: Not ideal to declare the displayName like this, but it is used by DragDrop component to identify the SectionTool and exclude wrapping it in draggable properties.  Looking into new logic for this but will be updated in a separate PR.

client/apps/edit/components/content2/section_tool/index.jsx
```diff
@@ -0,0 +1,146 @@
...
+        data-hero={isHero}>
+        <div
+          className='edit-tool__icon'
+          onClick={this.toggleOpen}>
+          <IconEditSection
+            fill={this.state.open || !isHero ? '#000' : '#CCC'}
+            isClosing={this.state.open} />
+        </div>
+        { isHero
+          ? this.renderHeroMenu()
```
craigspaeth: Nice extracting into methods here!

craigspaeth:
Great refactor work! I left some non-blocking comments that you can feel free to address and merge whenever you're ready.

client/apps/edit/components/content2/section_controls/index.styl
```diff
@@ -55,13 +55,23 @@
...
       max-width standard-overflow-w-margin
 
-.classic .edit-section__container
-  &[data-layout=column_width] .edit-controls
-    max-width classic-body-w-margin
-  &[data-layout=overflow_fillwidth],
-  &[data-type=image_set]
-    .edit-controls
-      max-width classic-overflow-w-margin
+.classic
```
craigspaeth: FWIW ideally this would be namespaced by component e.g. `.content2-classic` but probably not worth embarking on that refactor—just playing of-of-context CSS namespace cop.

client/apps/edit/components/content2/section_tool/index.jsx
```diff
@@ -0,0 +1,146 @@
...
+            Embed
+          </li>
+        </ul>
+      )
+    }
+  }
+
+  render() {
+    const { isEditing, isHero, sections } = this.props
+    const isFirstSection = this.props.firstSection && sections.length === 0 ? ' first' : ''
```
craigspaeth: Also, a place where a namespace would be better as `.first` is a very generic name for a css class that could accidentally conflict down the line. Alternatively, the data-attr pattern [you use below](https://github.com/artsy/positron/pull/1297/files#diff-a3b01687d52937b54c16e5ab3a27f233R129) is another way to avoid the problem because it's very unlikely a CSS selector would be written like `[data-state-open] { color: red }`.

Now that we're embarking more full-on-react-world, it might be good to peak at how we mix stylus and React in more greenfield projects like [Prediciton](https://github.com/artsy/prediction/blob/master/src/components/LotEventList.styl#L3) which puts a namespace at the top level and strictly uses the `&_modifying_class` underneath it.

api/apps/articles/model/schema.coffee
```diff
@@ -68,6 +68,16 @@ denormalizedArtwork = (->
     height: @number().allow(null)
 ).call Joi
 
+ImageCollectionSection = (->
+  @object().meta(
+    name: 'ImageCollection'
+    isTypeOf: (data) -> data.type is 'image_collection'
+  ).keys
+    type: 'image_collection'
+    layout: @string().allow('column_width', 'overflow_fillwidth', 'fillwidth', null)
```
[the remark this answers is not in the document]
eessex: ah good call -- just checked the database and 0 records exist without this field, going to go ahead and remove null.  Will put a default in as well.

PR #1304, opened 2017-09-28
@eessex => Deprecate artwork and image sections from article body

kanaabe:
Finally!


api/apps/articles/model/save.coffee
```diff
@@ -136,7 +136,6 @@ sanitize = (article) ->
   if article.sections
     sections = for section in article.sections
       section.body = sanitizeHtml section.body if section.type is 'text'
-      section.caption = sanitizeHtml section.caption if section.type is 'image'
       section.url = sanitizeLink section.url if section.type is 'video'
```
kanaabe: This reminds me, we should do another audit of new fields (postscript comes to mind) that could be an XSS risk.

api/apps/articles/model/schema.coffee
```diff
@@ -107,7 +107,6 @@ ImageCollectionSection = (->
   gravity_id: @string().objectid().allow('', null)
   hero_section: @alternatives().try(videoSection, ImageCollectionSection, imageSection, featureSection).allow(null).default(null)
   sections: @array().items([
-    imageSection
     ImageCollectionSection
```
eessex: I think we still need line 110 until the hero_section backfill

api/apps/articles/model/save.coffee
eessex: Yes for sure!

eessex:
Hooray :clap: Feel free to merge on 🍏

PR #1305, opened 2017-10-03
@craigspaeth => is_super_sub_article support

kanaabe:
We currently do multiple round trips just to decide if an article is part of a super article, which it isn't in most cases. Instead, this takes care of adding/removing a field called `is_super_sub_article` via the main super article's admin.


api/apps/articles/test/model/index/persistence.test.coffee
```diff
@@ -417,6 +417,17 @@ describe 'Article Persistence', ->
         article.super_article.footer_title.should.equal 'Footer Title'
         done()
 
+    it 'saves super sub articles', (done) ->
+      Article.save {
+        author_id: '5086df098523e60002000018'
+        is_super_article: false
+        is_super_sub_article: true
```
craigspaeth: Since this is "derived data" or a "computed property", I think it'd be better if we generated this field on the fly. e.g. in the [`present` function](https://github.com/artsy/positron/blob/master/api/apps/articles/model/index.coffee#L142) you can use... 

```coffeescript
superArticleCount = await db.articles.find(_id: article._id).count()
# ...
_.extend article,
  # ...
  is_super_sub_article: superArticleCount > 0
```

craigspaeth:
This is a decent solution, but it is a denormalization technique that risks this field getting out of sync. For instance, what happens when you delete the super article? You'd need to implement a callback to update this field now, which can get harry. I left a suggestion above that we generate this field on the fly—and if it becomes a performance problem we can revisit this solution. Lmk what you think!

kanaabe:
@craigspaeth ah yea I like your solution much better! Will update

api/apps/articles/routes.coffee
```diff
@@ -54,7 +54,7 @@ User = require '../users/model.coffee'
 
   Article.destroy req.article._id, (err) ->
     return next err if err
-    res.send present req.article
+    present(req.article).then((results) => res.send(results))
```
kanaabe: This was kind of an annoying result of `await`-ing on the db calls, but this works -- do you foresee any issues with this?

api/apps/graphql/index.js
```diff
@@ -12,18 +12,19 @@ import * as resolvers from 'api/apps/graphql/resolvers.js'
...
   relatedArticlesCanvas: array().items(object(Article.inputSchema)).meta({
     name: 'RelatedArticlesCanvas',
     resolve: resolvers.relatedArticlesCanvas
   }),
   relatedArticlesPanel: array().items(object(Article.inputSchema)).meta({
     name: 'RelatedArticlesPanel',
     resolve: resolvers.relatedArticlesPanel
-  })
+  }),
+  is_super_sub_article: boolean()
```
kanaabe: Now that we're not including `is_super_sub_article` directly in the Joi schema, this makes it available in the graphql schema manually.

scripts/reindex_articles.coffee
```diff
@@ -15,15 +16,17 @@ switch process.env.NODE_ENV
...
-    console.log('indexed ' + a.id or a._id)
-    indexWorker(articles, ++i)
-  , 30)
+  Article.present(a).then((results) =>
+    indexForSearch results
+    setTimeout( =>
+      console.log('indexed ' + a.id or a._id)
+      indexWorker(articles, ++i)
+    , 30)
+  )
```
kanaabe: This file doesn't make sense to me since it's calling `indexWorker(articles, 0)` and never iterating on the array itself but I added support for the presentation layer anyway... thoughts @cavvia ?

api/apps/articles/model/index.js
```diff
@@ -0,0 +1,203 @@
...
+
+const getSuperArticleCount = (id) => {
+  return new Promise((resolve, reject) => {
+    if (!ObjectId.isValid(id)) { return resolve(0) }
+    db.articles.find({ 'super_article.related_articles': ObjectId(id) }).count((err, count) => {
+      if (err) { return reject(err) }
+      resolve(count)
+    })
+  })
+}
```
kanaabe: Promisifies the mongo fetch for count.. no promise support in mongojs :(

api/apps/articles/test/routes.test.coffee
```diff
@@ -108,11 +118,13 @@ describe 'routes', ->
...
 
-    it 'delets an existing article', ->
+    it 'deletes an existing article', (done) ->
       @req.article = fixtures().articles
       routes.delete @req, @res
       @Article.destroy.args[0][1]()
-      @res.send.args[0][0].title.should.containEql 'Top Ten'
+      _.defer =>
+        @res.send.args[0][0].title.should.containEql 'Top Ten'
+        done()
```
kanaabe: I think these `defer`'s are also not ideal but I couldn't think of a better way to do something like a `.then` or `await` on the router handler since it doesn't return a promise. When we modernize the route handler I can promisify/await on those in this test.

kanaabe:
This was a good exercise and perhaps it was only painful because we've never had "smart" presentation layers before, but it does make me wonder if we're over-complicating the API and _maybe_ creating performance issues with the extra db call. 

I have another solution in mind where we only allow the field to be presented in our GraphQL endpoint and then introspect on the query to make calls only when needed. Thoughts?

scripts/reindex_articles.coffee
cavvia: I use recursion here with a sleep of 30ms between each call, in order to not overwhelm ES with indexing calls. We pass the index along with the array to the timeout handler `indexWorker` and it selects the next article. There's probably a more intuitive way to do this in js but I don't think there's a native `sleep` function, which is why i use timeouts. This will error when it's done rather than exit gracefully, so it could certainly be improved!

kanaabe: Ahh I see, my bad for not seeing that incrementor. Thanks @cavvia. We can use something like [`async.mapSeries`](https://caolan.github.io/async/docs.html#mapSeries) and run buckets of [parallel](https://caolan.github.io/async/docs.html#parallel) requests depending on how often we'd like to hit `indexForSearch`. I'll pair with you on it sometime if you'd like! For now this should continue to work as usual.

api/apps/articles/model/index.js
craigspaeth: The official driver supports promises, and this might be [worth looking at too](https://github.com/saintedlama/mongoist).

api/apps/articles/routes.coffee
craigspaeth: Yeah I realized that this would change the signature to an async/promise-returning function, resulting in trickling refactors 😞  For a fun related read about this challenge with most languages check out [this article](http://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/).

api/apps/articles/test/routes.test.coffee
craigspaeth: So right now because of Coffeescript implicit returns, these functions actually do return promises since the last line is `present(article).then((results) => res.send(results))` and `present` is returning a promise.

craigspaeth:
Thanks for embarking on this refactor! Overall code LGTM, but I do think you highlight great points about complicating `present` and the db overhead!

> I have another solution in mind where we only allow the field to be presented in our GraphQL endpoint and then introspect on the query to make calls only when needed. Thoughts?

This sounds ideal to me! If we plan to continue moving to GraphQL over RESTish, then this would be even better because it would allow us to lazy load the database call for only clients that request that field (plus one to GraphQL for even giving us this option instead of assuming the client needs `is_super_sub_article `). I don't think you'd even need to introspect the query—you could just add a resolver for it `is_super_sub_article: boolean().meta({ resolve: async () => (await getSuperArticlesCount()) > 0 })`

kanaabe:
Ok so I ripped out the REST-ish way and opted for the GraphQL solution instead which simplified everything tremendously! Our responses look like this:

image


Also @cavvia I went ahead and refactored the reindex script to use `mapSeries` which waits for the callback before continuing on the next index. Let me know if anything I did seems fishy!

api/apps/graphql/index.js
```diff
@@ -12,18 +13,21 @@ import * as resolvers from 'api/apps/graphql/resolvers.js'
...
   relatedArticlesCanvas: array().items(object(Article.inputSchema)).meta({
     name: 'RelatedArticlesCanvas',
     resolve: resolvers.relatedArticlesCanvas
   }),
   relatedArticlesPanel: array().items(object(Article.inputSchema)).meta({
     name: 'RelatedArticlesPanel',
     resolve: resolvers.relatedArticlesPanel
+  }),
+  is_super_sub_article: boolean().meta({
+    resolve: async (root) => (await getSuperArticleCount(root.id)) > 0
```
craigspaeth: 👌

craigspaeth:
Glad that worked out, much simpler! Feel free to merge when 💚

scripts/reindex_articles.coffee
```diff
@@ -15,15 +17,15 @@ switch process.env.NODE_ENV
 # Connect to database
 db = mongojs(process.env.MONGOHQ_URL, ['articles'])
 
-db.articles.find({ }).toArray (err, articles) ->
+db.articles.find({}).toArray (err, articles) ->
   console.log(err) if err
-  indexWorker(articles, 0)
+  console.log(articles.length)
+  async.mapSeries(articles, indexWorker, (err, results) =>
```
cavvia: Does this not risk sending a large number of parallel requests to ES? We want to avoid this. The original code used a 30ms delay between calls to keep the rate down to around 30/s. So we need to be able to rate limit this code too. Admittedly there's not too many articles in the database, but in general indexing bursts can affect ES performance.

kanaabe: Good question, it actually makes calls synchronously now because `mapSeries` waits for [this callback](https://github.com/artsy/positron/pull/1305/files#diff-d19ee1f1541fd71f3d7da42fe15aaf11R139) before moving on to the next index.

cavvia: :+1:

Issue #1308, opened 2017-10-05
Clicking into a text box jumps you up the page


owendodd:
https://cl.ly/092C1P103t2h

PR #1309, opened 2017-10-05
@kanaabe => Scroll bug (and a few other bugs)

eessex:
Closes https://github.com/artsy/positron/issues/1308

Chrome v61 and 63 were have conflicts with the fbjs library, required by Draft JS.  This caused a bug of scrolling to the top of the page whenever `focus()` is called. Here Draft is updated to 10.3, the fix is in its dependency fbjs, which is required at a higher version in 10.3.  [Context is here](https://github.com/facebook/draft-js/issues/14).

Also fixes some css in the hero-section that was not nested properly, and the [handleBackspace bug](https://sentry.io/artsynet/positron-production/issues/356782374/#) in `content`. (This had already been addressed in `content2`).

### DIFF
--- a/client/apps/edit/components/content/sections/text/index.coffee
+++ b/client/apps/edit/components/content/sections/text/index.coffee
@@ -103,7 +103,7 @@ module.exports = React.createClass
     selection = getSelectionDetails(@state.editorState)
     # only merge a section if cursor is in first character of first block
     if selection.isFirstBlock and selection.anchorOffset is 0 and
-    @props.sections.models[@props.index - 1].get('type') is 'text'
+    @props.sections.models[@props.index - 1]?.get('type') is 'text'
       mergeIntoHTML = @props.sections.models[@props.index - 1].get('body')
       @props.sections.models[@props.index - 1].destroy()
       newHTML = mergeIntoHTML + @state.html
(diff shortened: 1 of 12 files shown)


client/apps/edit/components/content/sections/text/test/index.coffee
```diff
@@ -419,7 +420,7 @@ describe 'SectionText', ->
 
   describe '#handleBackspace', ->
 
-    it 'merges section with previous section if at start of block', ->
+    xit 'merges section with previous section if at start of block', ->
```
kanaabe: Can you walk me through why we need to pend tests with the upgrade?

client/apps/edit/components/content/sections/text/index.coffee
```diff
@@ -23,7 +23,7 @@ window.process = {env: {NODE_ENV: sd.NODE_ENV}}
   setSelectionToStart,
   stripGoogleStyles } = require '../../../../../../components/rich_text/utils/index.coffee'
 editor = (props) -> React.createElement Editor, props
-{ div, nav, a, span, p, h3 } = React.DOM
+{ div, nav, a, span, p, h3, h2 } = React.DOM
```
kanaabe: Curious why we include h2 here when it doesn't seem like it's being used in the file? I looked for h3 as well but it doesn't seem to be there either - maybe i'm missing some context with draft tho?

kanaabe:
Good find! 🔎  I had a couple questions but if you'd rather merge this to fix the issue and then go back to fix the tests that's fine with me too 👍

PR #1315, opened 2017-10-09
@kanaabe => Character Limit component

eessex:
One piece of the Display Admin -- a new component with a maxLength, displays a remaining characters count. 

Has types 'input' (default) and 'textarea'.  Use the prop html (boolean) to allow rich-text with bold, italic and links.

### DIFF
--- a/client/components/character_limit/index.jsx
+++ b/client/components/character_limit/index.jsx
@@ -1,0 +1,83 @@
+import Paragraph from 'client/components/rich_text2/components/paragraph.coffee'
+import PlainText from '/client/components/rich_text2/components/plain_text.jsx'
+import PropTypes from 'prop-types'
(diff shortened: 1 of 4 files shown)


client/components/character_limit/index.jsx
```diff
@@ -0,0 +1,83 @@
...
+
+CharacterLimitInput.propTypes = {
+  defaultValue: PropTypes.string,
+  html: PropTypes.bool,
+  label: PropTypes.string,
+  limit: PropTypes.number.isRequired,
+  onChange: PropTypes.func.isRequired,
+  placeholder: PropTypes.string,
+  type: PropTypes.string
+}
```
kanaabe: 👏  for adding propTypes here

PR #1316, opened 2017-10-09
@kanaabe => DropDown component

eessex:
For the Display Admin -- 
Adds a component for DropDown headers that click to reveal children. A similar logic is used in a few places including the Article Admin, these can be replaced in the future with this component.


kanaabe:
Easy peasy 👍

Issue #1118, opened 2017-06-20
Article Type Drop Down


owendodd:
Displays on hover for New Article menu item. 

Relates to: https://github.com/artsy/positron/issues/1117

![screen shot 2017-06-16 at 5 43 56 pm](https://user-images.githubusercontent.com/3196536/27343866-4e84c878-55b2-11e7-8a0e-c022dc467952.png)

owendodd:
Looks good

PR #1319, opened 2017-10-10
@kanaabe: Article type drop down

craigspaeth:
Addresses https://github.com/artsy/positron/issues/1118

This adds the standard vs. classic article links to the New Article menu item. I wasn't totally sure this was the right design for this, but I saw it as an item [in the launch ticket](https://waffle.io/artsy/publishing/cards/59dbdce0b33ffa0050b0273d) that seemed easy to pick off. Lmk what you think.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,8 +41,8 @@ To shut down the process, press `ctrl+c` or execute `docker-compose down`.
 - Install Node 6
 
 ```
-nvm install 6
-nvm alias default 6
+nvm install 8
+nvm alias default 8
 ```
 
 - Fork Positron to your Github account in the Github UI.
--- a/client/components/layout/templates/sidebar.jade
+++ b/client/components/layout/templates/sidebar.jade
@@ -2,13 +2,23 @@
   nav#layout-sidebar
     a#layout-sidebar-home( href='/' )
       include ../public/icons/logo.svg
-    a#layout-sidebar-new-article(
-      href='/articles/new'
-      class=(sd.URL == '/articles/new' ? 'is-active' : '')
-    )
-      include ../public/icons/layout_new_article.svg
-      br
-      | New Article
+    if channel && channel.isEditorial()
+      #layout-sidebar-new-article(
+        class=(sd.URL == '/articles/new' ? 'is-active' : '')
+      )
+        include ../public/icons/layout_new_article.svg
+        br
+        | New Article
+        nav.layout-sidebar-dropdown
+          a( href='/articles/new?layout=standard' ) Standard Article
+          a( href='/articles/new?layout=feature' ) Feature Article
+    else
+      a#layout-sidebar-new-article(
+        class=(sd.URL == '/articles/new' ? 'is-active' : '')
+      )
+        include ../public/icons/layout_new_article.svg
+        br
+        | New Article
     a( href='/articles'
        class=(sd.URL == '/articles' ? 'is-active' : '') )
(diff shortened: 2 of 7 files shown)


eessex:
This should actually be for standard and feature, not classic

craigspaeth:
@eessex Thanks for the help! I updated the PR to set the `layout` depending on the link and only show the drop-down for Editorial + re-assigned to you. Lmk if this makes sense.

client/components/layout/templates/sidebar.jade
```diff
@@ -2,13 +2,23 @@
...
+    if user.isEditorial()
+      #layout-sidebar-new-article(
+        class=(sd.URL == '/articles/new' ? 'is-active' : '')
+      )
+        include ../public/icons/layout_new_article.svg
+        br
+        | New Article
+        nav.layout-sidebar-dropdown
+          a( href='/articles/new?layout=standard' ) Standard Article
+          a( href='/articles/new?layout=classic' ) Classic Article
```
eessex: I think we should still add feature here, or just remove classic -- trying to avoid any new editorial articles getting the classic layout.

client/assets/main.coffee
```diff
@@ -11,6 +11,7 @@ class Router extends Backbone.Router
 
   routes:
     'articles/new': 'articleEdit'
+    'articles/new2': 'articleEdit'
     'articles/:id/edit': 'articleEdit'
```
eessex: Is this used now?

client/components/layout/templates/sidebar.jade
```diff
@@ -2,13 +2,23 @@
...
     a#layout-sidebar-home( href='/' )
       include ../public/icons/logo.svg
-    a#layout-sidebar-new-article(
-      href='/articles/new'
-      class=(sd.URL == '/articles/new' ? 'is-active' : '')
-    )
-      include ../public/icons/layout_new_article.svg
-      br
-      | New Article
+    if user.isEditorial()
```
kanaabe: This works since we're looking at the user's `current_channel` which is always accurate, but it might make more sense if we used [this method](https://github.com/artsy/positron/blob/master/client/models/channel.coffee#L15) on the channel instead. However I don't _think_ channel gets passed into locals atm so this template wouldn't have access to it yet unless each app passed that. Soo down the line it probably makes more sense to pass channel into https://github.com/artsy/positron/blob/master/client/lib/middleware.coffee#L14 but I think this is a good in-between. Not a blocker.

README.md
```diff
@@ -41,8 +41,8 @@ To shut down the process, press `ctrl+c` or execute `docker-compose down`.
 - Install Node 6
 
 ```
-nvm install 6
-nvm alias default 6
+nvm install 8
+nvm alias default 8
```
kanaabe: Thx!

client/components/layout/templates/sidebar.jade
craigspaeth: Oh haha, whoops I just totally overlooked that. Updating..

client/assets/main.coffee
craigspaeth: Nope, removing.

eessex:
Awesome, thanks @craigspaeth! This looks good to me when 🍏

craigspaeth:
Thanks for the comments Eve + Kana. Cleanups made and commits squashed.

PR #1320, opened 2017-10-12
@kanaabe => Display admin

eessex:
Adds a Display Admin to `settings/curations` for creating Panel and Canvas layouts on the display-admin curation. 

This assumes images are still in the standard bucket, and includes minor changes to the admin file-upload component to support images.  Will address using a different bucket in a future PR.

display-admin

### DIFF
--- a/client/apps/settings/client/curations/display/components/canvas.jsx
+++ b/client/apps/settings/client/curations/display/components/canvas.jsx
@@ -64,16 +64,16 @@ export default class Canvas extends React.Component {
                     type='textarea'
                     label='Body'
                     placeholder='Body'
-                    defaultValue={campaign.canvas ? campaign.canvas.body : ''}
+                    defaultValue={campaign.canvas ? campaign.canvas.headline : ''}
                     onChange={(html) => onChange('canvas.body', html, index)}
                     html
-                    limit={90} />
+                    limit={70} />
                 : <CharacterLimitInput
                     label='headline'
                     placeholder='Headline'
                     defaultValue={campaign.canvas ? campaign.canvas.headline : ''}
                     onChange={(e) => onChange('canvas.headline', e.target.value, index)}
-                    limit={25} />
+                    limit={45} />
                 }
             </div>
             <div className='field-group'>
@@ -99,7 +99,7 @@ export default class Canvas extends React.Component {
                 placeholder='Enter legal disclaimer here'
                 defaultValue={campaign.canvas ? campaign.canvas.body : ''}
                 onChange={(e) => onChange('canvas.body', e.target.value, index)}
-                limit={90} />
+                limit={150} />
             </div>
           </Col>
           <Col lg>
--- a/client/apps/settings/client/curations/display/components/canvas.jsx
+++ b/client/apps/settings/client/curations/display/components/canvas.jsx
@@ -5,87 +5,126 @@ import { Col, Row } from 'react-styled-flexboxgrid'
 import ImageUpload from 'client/apps/edit/components/admin/components/image_upload.coffee'
 import CharacterLimitInput from 'client/components/character_limit/index.jsx'
 
-const Canvas = (props) => {
-  const {campaign, index, onChange} = props
-  return (
-    <div className='display-admin__section--campaign'>
-      <Row key={index} className='inputs--text'>
-        <Col lg>
-          <div className='field-group'>
-            {campaign.canvas.layout === 'overlay'
-              ? <CharacterLimitInput
-                  type='textarea'
-                  label='Body'
-                  placeholder='Body'
-                  defaultValue={campaign.canvas ? campaign.canvas.body : ''}
-                  onChange={(html) => onChange('canvas.body', html, index)}
-                  html
-                  limit={90} />
-              : <CharacterLimitInput
-                  label='headline'
-                  placeholder='Headline'
-                  defaultValue={campaign.canvas ? campaign.canvas.headline : ''}
-                  onChange={(e) => onChange('canvas.headline', e.target.value, index)}
-                  limit={25} />
-              }
-          </div>
-          <div className='field-group'>
-            <CharacterLimitInput
-              label='CTA Text'
-              placeholder='Find Out More'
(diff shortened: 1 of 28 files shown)


client/apps/settings/client/curations/display/components/canvas.jsx
```diff
@@ -0,0 +1,75 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import { Col, Row } from 'react-styled-flexboxgrid'
+
+import CanvasImages from './canvas_images.jsx'
+import CharacterLimitInput from 'client/components/character_limit/index.jsx'
+
+export default class Canvas extends React.Component {
+  render () {
```
kanaabe: If there's only a `render` method I prefer the use of stateless functional components

client/apps/settings/client/curations/display/components/canvas.jsx
```diff
@@ -0,0 +1,75 @@
...
+                placeholder='Find Out More'
+                defaultValue={campaign.canvas && campaign.canvas.link ? campaign.canvas.link.url : ''}
+                onChange={(e) => onChange('canvas.link.url', e.target.value, index)} />
+            </div>
+            <div className='field-group'>
+              <CharacterLimitInput
+                type='textarea'
+                label='Disclaimer (optional)'
+                placeholder='Enter legal disclaimer here'
+                defaultValue={campaign.canvas ? campaign.canvas.body : ''}
```
kanaabe: Seeing a bunch of these checks - perhaps this could be cleaned up by caching `campaign.canvas` at the top in a const?

client/apps/settings/client/curations/display/components/canvas_images.jsx
```diff
@@ -0,0 +1,129 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import { Col, Row } from 'react-styled-flexboxgrid'
+import ImageUpload from 'client/apps/edit/components/admin/components/image_upload.coffee'
+
+export default class CanvasImages extends React.Component {
```
kanaabe: Not sure if you saw [this convo](https://artsy.slack.com/archives/C2TQ4PT8R/p1507801960000308) (and it's definitely not necessary for this PR) but it brings up good points about exporting with `default`!

client/apps/settings/client/curations/display/components/panel.jsx
```diff
@@ -0,0 +1,77 @@
...
+            </div>
+            <div className='field-group'>
+              <CharacterLimitInput
+                type='textarea'
+                label='Body'
+                placeholder='Body'
+                defaultValue={campaign.panel ? campaign.panel.body : ''}
+                onChange={(value) => onChange('panel.body', value, index)}
+                html
+                limit={45} />
```
kanaabe: This component is turning out to be so handy!

client/apps/settings/client/curations/display/index.jsx
```diff
@@ -0,0 +1,92 @@
...
+  newCampaign = () => {
+    const newCuration = this.state.curation.clone()
+    const newCampaign = {panel: {assets: []}, canvas: {assets: []}}
+    newCuration.get('campaigns').push(newCampaign)
+    this.setState({
+      curation: newCuration,
+      saveStatus: 'Save',
+      activeSection: newCuration.get('campaigns').length - 1
+    })
+  }
```
kanaabe: Nice :)

client/apps/settings/client/curations/display/index.styl
```diff
@@ -0,0 +1,59 @@
...
+      text-transform none
+      letter-spacing 0
+  select
+    height 2.5em
+  .field-group
+    margin-bottom 20px
+  .save-campaign
+    position fixed
+    top 10px
+    right calc(50vw + 55px - (1120px / 2))
```
kanaabe: What is this? Maybe some vars or a comment would help?

package.json
```diff
@@ -70,6 +70,7 @@
     "knox": "^0.9.2",
     "lodash": "^4.17.4",
     "lodash.clonedeep": "^4.3.1",
+    "lodash.set": "^4.3.2",
```
kanaabe: Ah do you mind removing this along with `lodash.clonedeep` since we have the full lib installed?

client/apps/settings/client/curations/display/components/canvas_container.jsx
```diff
@@ -0,0 +1,53 @@
...
+  constructor (props) {
+    super(props)
+    this.state = {
+      activeLayout: props.campaign.canvas.layout || 'overlay'
+    }
+  }
+
+  setActiveLayout = (layout) => {
+    this.setState({activeLayout: layout})
+    this.props.onChange('canvas.layout', layout, this.props.index)
```
kanaabe: What happens if you switch from a video layout to overlay? Or slideshow with many images to video? Maybe some level of validation on asset type and quantity could be worthwhile here.

eessex: If you switch from a slideshow with many images to another layout it does not delete them, and only shows the new one.  

Video is not allowed to be uploaded unless you are a video section, but it does not delete the files -- was planning to add a warning for not-allowed files but in a separate PR. 

Slideshows limit uploads to 5 images, and hide the upload for more if you have 5.

kanaabe: Yeah it's good that the individual layouts have some validation. For the pilot it's not necessary but I thought it could spark some ideas :). API rules are little more relaxed with Curations, but ideally we could have "layout rules" that handle switching more gracefully -- ie, mapping business rules like "canvas/video has only 1 video asset" to the API so the client doesn't have to think about the possibility of multiple assets or different types.

eessex: Yeah I figured there would be some ability to switch between layouts while you are building, so it seems like we don't want to prevent saving.  However, yeah we do need to validate.  Was planning to do something like image_collections, where an alert tells you if you have too many assets.

PR #1321, opened 2017-10-12
@eessex: Update the display resolver to return a single object instead of a list

craigspaeth:
Since we're implementing the SOV logic server-side this seems like a more appropriate type.

image


eessex:
Looks good!  Feel free to merge when 🍏

PR #1322, opened 2017-10-12
@eessex => Do not send article to Sailthru if it is scheduled

kanaabe:
Closes https://github.com/artsy/publishing/issues/85


eessex:
Easy peasy, feel free to merge when 🍏

PR #1329, opened 2017-10-12
@kanaabe => Update Reaction Paths

eessex:
- Updates Reaction
- Replaces reaction paths with CamelCase
- Adds spacing/propTypes changes to a few of the touched files to appease the linter


client/apps/edit/components/content2/section_controls/index.jsx
```diff
@@ -155,3 +157,14 @@ export default class SectionControls extends Component {
...
+SectionControls.propTypes = {
+  articleLayout: PropTypes.string,
+  channel: PropTypes.object,
+  children: PropTypes.any,
+  disabledAlert: PropTypes.func,
+  isHero: PropTypes.bool,
+  onChange: PropTypes.func,
+  section: PropTypes.object.isRequired,
+  sectionLayouts: PropTypes.bool
+}
```
kanaabe: 😎

kanaabe:
Thank ye very much! There's a 🔴  but looks good after it passes!

P.S. I have another working branch which converts the rest of the defaults in Reaction and I'll follow that up with a Positron PR to support it. ✔️

eessex:
The red all looks to be from a new package, have you seen anything like this error before, or when you switched the force paths?

screen shot 2017-10-12 at 5 39 41 pm

kanaabe:
Hmm I didn't see this error with the Force update - It looks like the path of the error comes from EmailSignup - is the test trying to do something with that?

eessex:
No, weirdly a number of failures are on components that don't even touch reaction.  This is also all passing locally, so I'm guessing it is a problem with the build on circle... possibly not getting the correct env again.

PR #1331, opened 2017-10-13
@eessex: Render a new article link for non-editorial

craigspaeth:
Simple copy and paste error, this fixes it with a test.


eessex:
Thanks!!

PR #1332, opened 2017-10-13
@kanaabe => Feature Header updates

eessex:
- Adds default of 'text' to hero_section.type
- Replaces classnames that changed with reaction namespacing (selectors of this kind are only used here)
- In fullscreen headers, make 'change layout' and 'upload background' buttons black if no background (new gradient makes the white ones invisible)


kanaabe:
Looks straightforward to me! Got a real ❌  though

PR #1334, opened 2017-10-15
@eessex => Update reaction paths and fix Display schema

kanaabe:
(For Monday)
- Updates reaction paths post export default killing
- Fixes Display schema to use `url` as the asset and adds meta information on display schema to support fragments


eessex:
Accidentally merged this -- checking it out now.

eessex:
Looks good, I see no need to revert.

PR #1340, opened 2017-10-16
@kanaabe => Display QA

eessex:
Forgot to test the Campaign file in the last PR for Display Admin, this adds tests for the Canvas component and fixes saving bugs referenced here: https://github.com/artsy/publishing/issues/95

### DIFF
--- a/client/apps/settings/client/curations/display/components/campaign.jsx
+++ b/client/apps/settings/client/curations/display/components/campaign.jsx
@@ -10,6 +10,7 @@ function onChangeDate (name, value, index, onChange) {
 
 export const Campaign = (props) => {
   const {campaign, index, onChange} = props
+
   return (
     <div className='display-admin__section--campaign'>
       <Row key={index}>
@@ -40,7 +41,7 @@ export const Campaign = (props) => {
         <Col lg>
           <label>Traffic Quantity</label>
           <select
-            defaultValue={campaign.sov}
+            value={campaign.sov.toFixed(2)}
             className='bordered-input'
             onChange={(e) => onChange('sov', parseFloat(e.target.value), index)} >
             <option value='0.25'>25%</option>
@@ -58,5 +59,3 @@ Campaign.propTypes = {
   index: PropTypes.number.isRequired,
   onChange: PropTypes.func.isRequired
 }
-
-export default Campaign
--- a/client/apps/settings/client/curations/display/tests/components/canvas.test.js
+++ b/client/apps/settings/client/curations/display/tests/components/canvas.test.js
@@ -1,0 +1,178 @@
+import React from 'react'
+import { Canvas } from '../../components/canvas.jsx'
+import { CharacterLimit } from '/client/components/character_limit'
+import { PlainText } from '/client/components/rich_text2/components/plain_text.jsx'
+import { mount } from 'enzyme'
+import {
+  UnitCanvasImage,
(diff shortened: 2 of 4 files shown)


client/apps/settings/client/curations/display/components/campaign.jsx
```diff
@@ -8,6 +8,15 @@ function onChangeDate (name, value, index, onChange) {
...
 }
 
+function setDefaultSov (value) {
+  var newValue = value.toString()
+  if (newValue.length === 3) {
+    // add a trailing 0 to .5
+    newValue = newValue + '0'
+  }
+  return newValue
+}
```
kanaabe: This makes me wonder why we are saving these values as strings and not floats. They could even be integers with a max of 100. Appending '0' to a string when later we will later want to support things like SOV sum validation seems like we'd regret this pretty quickly imo.

client/apps/settings/client/curations/display/tests/components/canvas.test.js
```diff
@@ -0,0 +1,178 @@
...
+      expect(component.html()).toMatch(props.campaign.canvas.headline)
+      expect(component.html()).toMatch(props.campaign.canvas.link.text)
+      expect(component.html()).toMatch(props.campaign.canvas.link.url)
+      expect(component.text()).toMatch(props.campaign.canvas.disclaimer)
+      expect(component.html()).toMatch(props.campaign.canvas.logo)
+      expect(component.html()).toMatch(props.campaign.canvas.assets[0].url)
+      expect(component.html()).toMatch(props.campaign.canvas.assets[1].url)
+    })
+  })
+})
```
kanaabe: Nice tests!

client/apps/settings/client/curations/display/components/campaign.jsx
eessex: We are saving as floats, it just needs to be a string to render the options on the select

kanaabe: Ah I see, my bad! Well in that case I think [toFixed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) solves the string problem which _should_ remove the need for this method.

eessex: This is actually going the other way around, toString is the desired result.  Turning .75 to string works fine, but .5 returns '0.5' rather than '0.50'.  The select can't render a saved value unless it is an exact match.  The value is saved as an integer here if saved - https://github.com/eessex/positron/blob/7c8dc119a5a452ac68a3b35d0b4142357986fa15/client/apps/settings/client/curations/display/components/campaign.jsx#L54

eessex: oh wait i misunderstood -- yes will use toFixed

client/apps/settings/client/curations/display/components/campaign.jsx
```diff
@@ -10,6 +10,7 @@ function onChangeDate (name, value, index, onChange) {
 
 export const Campaign = (props) => {
   const {campaign, index, onChange} = props
+  console.log(campaign.sov.toFixed(2))
```
kanaabe: stray log!

eessex: good catch

PR #1341, opened 2017-10-17
@craigspaeth => Static Display

kanaabe:
This doesn't include analytics yet because the DisplayPanel does not have react-tracking set up.

image


craigspaeth:
Nice! LGTM!

Issue #1345, opened 2017-10-17
Feature Headers not saving on published articles


eessex:
Feature headers are not saving on published articles.  Additionally this causes 500s in force because hero_section is null.

Flagged by Molly - https://artsy.slack.com/archives/C04GGGDCM/p1508260662000354

PR #1347, opened 2017-10-18
@damassi => Feature header save bugs

eessex:
Closes https://github.com/artsy/positron/issues/1345
Resolves a bug where feature headers were saving inconsistent information. 

- Removes logic for using hero_section title from article model (titles are set directly on the article now)
- Allows hero_section to be saved with 1 or more keys (previously was overwriting 1 or less)
- Adds persistence tests for each hero_section feature layout
- Minor cleanup in the article header component

### DIFF
--- a/api/apps/articles/model/index.js
+++ b/api/apps/articles/model/index.js
@@ -81,9 +81,7 @@ export const save = (input, accessToken, options, callback) => {
 
         // Eventually convert these to Joi custom extensions
         modifiedArticle.updated_at = new Date()
-        if (input.hero_section && input.hero_section.type === 'fullscreen') {
-          modifiedArticle.title = modifiedArticle.hero_section.title
-        }
+
         if (input.author) { modifiedArticle.author = input.author }
 
         // Handle publishing, unpublishing, published, draft
--- a/api/apps/articles/model/index.js
+++ b/api/apps/articles/model/index.js
@@ -20,7 +20,9 @@ const { removeFromSearch, deleteArticleFromSailthru } = require('./distribute.co
 //
 export const where = (input, callback) => {
   return Joi.validate(input, schema.querySchema, { stripUnknown: true }, (err, input) => {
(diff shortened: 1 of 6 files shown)


api/apps/articles/model/index.js
```diff
@@ -81,9 +81,7 @@ export const save = (input, accessToken, options, callback) => {
 
         // Eventually convert these to Joi custom extensions
         modifiedArticle.updated_at = new Date()
-        if (input.hero_section && input.hero_section.type === 'fullscreen') {
-          modifiedArticle.title = modifiedArticle.hero_section.title
-        }
+
         if (input.author) { modifiedArticle.author = input.author }
```
damassi: Conditionals are always a bit easier to read at a glance if broken out like 
```js
if (input.author) { 
  modifiedArticle.author = input.author 
}
```
Also (though i'm not suggesting one does this as it typically leads to subtle bugs) if on a single line no brackets are necessary.

client/apps/edit/components/content2/sections/header/index.jsx
```diff
@@ -22,24 +23,13 @@ export default class SectionHeader extends Component {
...
     this.onChange('hero_section', heroSection.attributes)
-  }
-
-  onChangeLeadParagraph = (html) => {
-    this.onChange('lead_paragraph', html)
-  }
-
-  onUpload = (image, width, height) => {
-    this.onChangeHero('url', image)
+    this.forceUpdate() // re-render reaction header
```
damassi: The [React docs](https://reactjs.org/docs/react-component.html#forceupdate) suggest to avoid using `forceUpdate` when you can as it skips lifecycle methods (which further down the line is weird to debug). A better way to trigger updates in this fashion would be to use `setState`:
```js
  onChangeHero = (key, value) => {
    ...
    this.setState({
      refreshedAt: Date.now()
    })
  }
```

Issue #1343, opened 2017-10-17
Display Admin - Slideshow image bugs


eessex:
- [x] Should limit uploads to 5 total
- [x] First image is not being removed

PR #1353, opened 2017-10-19
@damassi => Display Admin QA 

eessex:
Closes https://github.com/artsy/positron/issues/1343

Addresses some problems with image previews for display canvas slideshow admin, specifically in the image upload component. 

Slideshows use an array of images. When a file in the middle of the array is removed, the other images are expected to shift up into its place.  Because we were setting the src as state on each input component, this wasn't happening. 

- Use props instead of state for rendering the src in image uploads
- Add an error for existing video files if video is not allowed
- Use resizer instead of rendering preview images at full size
- For logos, contain the image in the preview rather than crop


client/apps/edit/components/admin/test/image_upload.test.coffee
```diff
@@ -37,13 +37,15 @@ describe 'ImageUpload', ->
       $(ReactDOM.findDOMNode(@component)).html().should.containEql 'accept="image/jpg,image/jpeg,image/gif,image/png"'
 
     it 'Renders an image preview if file is present', ->
-      @component.setState src: 'http://artsy.net/image.jpg'
-      $(ReactDOM.findDOMNode(@component)).html().should.containEql 'background-image: url(http://artsy.net/image.jpg)'
+      @props.src = 'http://artsy.net/image.jpg'
+      component = ReactDOM.render React.createElement(@ImageUpload, @props), (@$el = $ "<div></div>")[0]
```
damassi: No need to fix, but could just do `$ '<div />[0]'`

client/apps/settings/client/curations/display/components/canvas_images.jsx
```diff
@@ -4,9 +4,9 @@ import { Col, Row } from 'react-styled-flexboxgrid'
 import ImageUpload from 'client/apps/edit/components/admin/components/image_upload.coffee'
 
 export class CanvasImages extends React.Component {
-  onSlideshowImageChange = (imgIndex, url) => {
+  updateImageUrls = (imgIndex, url) => {
     const { assets } = this.props.campaign.canvas
-    if (imgIndex || imgIndex === 0) {
+    if ((imgIndex || imgIndex === 0) && assets.length) {
```
damassi:
```
if (assets && isNumber(imgIndex)) {
  ...
}
```

client/apps/settings/client/curations/display/components/canvas_images.jsx
```diff
@@ -19,13 +19,8 @@ export class CanvasImages extends React.Component {
...
   onImageInputChange = (key, value, imgIndex) => {
-    let newValue
-    if (this.props.campaign.canvas.layout === 'slideshow') {
-      newValue = this.onSlideshowImageChange(imgIndex, value)
-    } else {
-      newValue = value.length ? [{url: value}] : []
-    }
-    this.props.onChange(key, newValue, this.props.index)
+    const assets = this.updateImageUrls(imgIndex, value)
+    this.props.onChange(key, assets, this.props.index)
```
damassi: Minor, but its a good habit to get into is if you use a prop or state value more than once to pluck it out to keep things DRY:

```javascript 
const { 
  onChange, 
  index 
} = this.props 

const assets = this.updateImageUrls(imgIndex, value)
onChange(key, assets, index)
```

PR #1357, opened 2017-10-19
@eessex => Support SOV less than 100%

kanaabe:
Returns `display: null` if there is no ad to show which Force/Reaction can safely handle. On the static endpoint, it returns a 404 error. As far as I can tell, IA doesn't have a great way to handle hiding an ad so I'm going to experiment with this 404 and see if I can get something to work in their [Developer tools](https://business.facebook.com/artsy/publishing_tools/?section=INSTANT_ARTICLES_DEVELOPMENT)

### DIFF
--- a/api/apps/graphql/resolvers.js
+++ b/api/apps/graphql/resolvers.js
@@ -88,17 +88,30 @@ export const channels = (root, args, req, ast) => {
 
 export const display = (root, args, req, ast) => {
   return new Promise((resolve, reject) => {
-    Curation.mongoFetch({ '_id': ObjectId(DISPLAY_ID) }, (err, results) => {
-      if (err) {
-        reject(new Error(err))
-      }
+    Curation.mongoFetch({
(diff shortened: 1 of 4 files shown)


api/apps/graphql/resolvers.js
```diff
@@ -88,17 +88,30 @@ export const channels = (root, args, req, ast) => {
...
   return new Promise((resolve, reject) => {
-    Curation.mongoFetch({ '_id': ObjectId(DISPLAY_ID) }, (err, results) => {
-      if (err) {
-        reject(new Error(err))
-      }
+    Curation.mongoFetch({
+      '_id': ObjectId(DISPLAY_ID)
+    }, (err, { results }) => {
+      if (err) { reject(new Error(err)) }
+
```
eessex: There was some discussion yesterday with @damassi on not using single line conditionals for readability -- minor -- up to you

PR #1362, opened 2017-10-20
@damassi => Resolve null instead of throwing errors on RelatedArticle queries

kanaabe:
All non-editorial articles on article2 pages are currently broken like this one: https://www.artsy.net/article2/gallery-insights-the-pop-up-gallery-checklist

They're sadly breaking on related article queries again. I'm thinking of fixing this in two parts. First, we should stop throwing errors if there are no articles to show -- instead resolve with null. 

The second part will be on this try/catch in Force [here](https://github.com/artsy/force/blob/master/desktop/apps/article2/routes.js#L20-L21). I'm thinking of removing the block and handling errors manually. That way, if one of the resolvers throws an error, we don't have to kill the entire page. What's happening now is that the above article can't reach [this line](https://github.com/artsy/force/blob/master/desktop/apps/article2/routes.js#L28) since the Graphql fetch contains an error.

### DIFF
--- a/api/apps/graphql/resolvers.js
+++ b/api/apps/graphql/resolvers.js
@@ -172,13 +172,16 @@ export const relatedArticlesCanvas = (root) => {
     limit: 4,
     sort: '-published_at'
   }
-  return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


damassi:
👍

PR #1364, opened 2017-10-23
@kanaabe => Fetch data for author_ids

eessex:
- Adds ability to query authors by a list of ids
- Adds resolver for 'relatedAuthors' to return articles by list of ids
- Adds authors to the article metaFields

closes https://github.com/artsy/publishing/issues/110

### DIFF
--- a/api/apps/graphql/index.js
+++ b/api/apps/graphql/index.js
@@ -29,6 +29,9 @@ const metaFields = {
   display: Display.schema.meta({
     args: Display.querySchema,
     resolve: resolvers.display
+  }),
+  authors: array().items(object(Author.schema)).meta({
+    resolve: resolvers.relatedAuthors
   })
 }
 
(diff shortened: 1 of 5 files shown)


api/apps/graphql/resolvers.js
```diff
@@ -64,6 +64,20 @@ export const authors = (root, args, req, ast) => {
   })
 }
 
+export const relatedAuthors = (root) => {
```
kanaabe: Seems kinda funny to call this `relatedAuthors` if we've just been using the term `authors`?

kanaabe: Hmm I guess we've already got `authors` as a resolver -- i suppose this is fine!

api/apps/graphql/index.js
```diff
@@ -29,6 +29,9 @@ const metaFields = {
   display: Display.schema.meta({
     args: Display.querySchema,
     resolve: resolvers.display
+  }),
+  authors: array().items(object(Author.schema)).meta({
```
kanaabe: 👍

kanaabe:
Cool, thx! Good work figuring out the implementation of these "smart" resolvers.

PR #1366, opened 2017-10-26
@kanaabe => Filter Display Ads that are outside of valid range 

damassi:
When a Display's `start_date` is after now or its `end_date` is before now filter it out from the results.


jsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "compilerOptions": {
+    "rootDir": "./",
+    "baseUrl": "./"
+  }
+}
```
damassi: If using VSCode, this allows us to `ctrl+click` directly to anything written in JS, like we would in TypeScript. Unfortunately it doesn't work with `.coffee` files tho :(

kanaabe:
Thanks! Looks great.

PR #1370, opened 2017-10-27
@eessex => Add basic header to editor

damassi:
This PR implements a new Standard Header [from Reaction](https://github.com/artsy/reaction/pull/315) and wires it into `/edit2`. (Right now there's one little bug that I've called out in the comments.)

#### TODO: 
- [x] Write tests
- [x] Publish Reaction and update `package.json`

cc @owendodd 

basic


scripts/mocha.sh
```diff
@@ -6,7 +6,6 @@ mocha \
   --require should \
   --ui bdd \
   --timeout 10000 \
+  --require test/env-config.js \
```
damassi: @kanaabe - this fixes the unsafe test env issue

client/apps/edit/components/content2/index.coffee
```diff
@@ -10,14 +10,15 @@ module.exports = React.createClass
 
   componentWillMount: ->
     @debouncedSave = _.debounce((->
-      @props.article.save()
+      @props.article.save() # TODO: Why is this not saving updates from #19?
```
damassi: @eessex - this thing is almost totally complete but for some reason whenever line 19 fires (for example when uploading an image) it eventually calls `@debouncedSave` but doesn't save the latest values. Logging out in 19 below everything is coming through right, but for some reason when I log here there is no `cover_image` on the model.

client/apps/edit/components/content2/index.coffee
```diff
@@ -10,14 +10,15 @@ module.exports = React.createClass
...
       @onChange()
     ), 800)
     @props.article.on 'change', => @saveArticle()
     @props.article.sections.on 'change add remove reset', => @saveArticle()
     @props.article.heroSection.on 'change remove', => @saveArticle()
+    @props.article.heroSection.on 'change remove', ({changed}) => @saveArticle(changed)
 
-  saveArticle: ->
+  saveArticle: (changed) ->
```
damassi: #fixme - remove

scripts/mocha.sh
```diff
@@ -6,7 +6,6 @@ mocha \
   --require should \
   --ui bdd \
   --timeout 10000 \
+  --require test/config.js \
```
damassi: @kanaabe - this fixes the unsafe test env issue we saw the other day

client/apps/edit/components/content2/sections/header/controls.jsx
```diff
@@ -22,15 +27,62 @@ export default class FeatureHeaderControls extends Component {
...
+    if (isLayoutOpen) {
+      this.toggleLayoutControls()
+    }
+  }
+
+  renderVideoEmbed () {
+    const { article } = this.props
+
+    if (this.state.isVideoOpen) {
+      return (
```
eessex: Since we already have the preview rendering in the main layout, you could probably use just the video controls component here.

client/apps/edit/components/content2/sections/header/controls.jsx
```diff
@@ -22,15 +27,62 @@ export default class FeatureHeaderControls extends Component {
...
+        <div
+          className='edit-section-container'
+          data-editing
+          data-type='video'
+        >
+          <SectionVideo editing
```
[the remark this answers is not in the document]
damassi: Ok just to recap: 
1. I'll skip rendering the entire video component and instead rely on `<VideoControls />`, thereby simplifying the view (since the preview video already renders at full size in the `<BasicHeader />`; no need for duplicates  
2. Move the new video toggle work outside of `<HeaderControls />` and into main index, so that it can be passed in as a child in a similar manner as the other UI items

👍

api/apps/articles/model/schema.coffee
```diff
@@ -22,7 +22,7 @@ videoSection = (->
     name: 'Video'
     isTypeOf: (data) -> data.type is 'video'
   ).keys
-    type: @string().valid('video')
+    type: @string().valid('video', 'basic')
```
damassi: @eessex - Not sure where the model save polling is coming from but this was the issue with saving `cover_image_url`

client/apps/edit/components/content2/sections/header/controls/index.jsx
```diff
@@ -0,0 +1,66 @@
...
+        <VideoControls {...this.props}
+          isOpen={isVideoOpen}
+          onClick={() => this.toggleControls('video')}
+        />
+
+        <LayoutControls {...this.props}
+          isOpen={isLayoutOpen}
+          onClick={() => this.toggleControls('layout')}
+        />
+      </div>
```
damassi: I started working the controls into the structure we discussed earlier today, but realized that it would probably be better to refactor out the controls into something more container-like so as to keep things decoupled from reaction / self-contained and a bit simpler. I wanted to the <BasicHeader /> to just be a simple rendering component for props and not have to worry about the outside world, children, etc. Would be open to refactoring further at a later point but given the time frame this is good to go...

jsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "rootDir": "./",
-    "baseUrl": "./"
+    "baseUrl": "./",
+    "jsx": "react"
```
damassi: Can now `cmd+click` into js / jsx / components and go to definitions.

api/apps/articles/test/model/index/persistence.test.coffee
```diff
@@ -732,3 +732,18 @@ describe 'Article Persistence', ->
...
+          url: 'http://image.com'
+          type: 'basic'
+          deck: 'This year was incredible.'
+          cover_image_url: 'http://some-cover-image.png'
+      }, 'foo', {}, (err, article) ->
+        return done err if err
+        article.hero_section.url.should.equal 'http://image.com'
+        article.hero_section.type.should.equal 'basic'
+        article.hero_section.deck.should.equal 'This year was incredible.'
+        article.hero_section.cover_image_url.should.equal 'http://some-cover-image.png'
```
damassi: @kanaabe - for whatever reason this test only fails on Circle -- thoughts? Since I updated the schema do I need to run a migration of some sort?

damassi:
One thing we need to do is add a button to reset attached cover image (maybe a tiny "reset" button in the attachment interface that only appears when an image is present?) but I don't want that to hold up this pr if possible.

client/apps/edit/components/content2/sections/header/controls/index.jsx
```diff
@@ -0,0 +1,73 @@
...
+          />
+        }
+        <LayoutControls
+          hero={hero}
+          isOpen={isLayoutOpen}
+          onChange={onChange}
+          onClick={() => this.toggleControls('layout')}
+        />
+      </div>
+    )
```
damassi: I started working the controls into the structure we discussed earlier today, but realized that it would probably be better to refactor out the controls into something more container-like so as to keep things decoupled from reaction / self-contained and a bit simpler. I wanted the <BasicHeader /> to just be a dumb rendering component for props and not have to worry about the outside world, children, etc. Would be open to refactoring further at a later point but given the short time frame this is good to go.

api/apps/articles/model/schema.coffee
eessex: Ah cool -- yes we do use a video section for the hero_section on classic articles, this makes sense to me for features.

client/apps/edit/components/content2/index.coffee
```diff
@@ -13,6 +13,8 @@ module.exports = React.createClass
       @props.article.save()
       @onChange()
     ), 800)
+
+    # FIXME: Polling autosave for `last_updated`, not sure where though
     @props.article.on 'change', => @saveArticle()
```
eessex: What's up with the last_updated?  This was added just below in this file from this PR - https://github.com/artsy/positron/pull/1347

client/apps/edit/components/content2/sections/header/controls/VideoControls.jsx
```diff
@@ -0,0 +1,42 @@
...
+            <div
+              className='edit-section-container'
+              data-editing
+              data-type='video'
+            >
+              <SectionVideo editing
+                article={article}
+                section={article.heroSection}
+                channel={{ isArtsyChannel: () => false }} // TODO: is this needed?
+                headerType={'basic-embed'}
```
eessex: The channel prop is for the backbone channel, which has isArtsyChannel as a method on the model.

client/apps/edit/components/content2/sections/header/controls/index.jsx
eessex: This looks like an ok solution to me.

client/apps/edit/components/content2/sections/video/controls.jsx
```diff
@@ -1,54 +1,80 @@
...
-    if (!this.props.section.get('url').length) {
+    if (!this.props.section.get('url')) {
       this.refs.input.focus()
     }
   }
 
   onCoverImageChange = (url) => {
-    this.props.section.set('cover_image_url', url)
+    const isValid = isEmpty(url) || isWebUri(url)
+
```
eessex: :+1:

client/apps/edit/components/content2/sections/video/index.jsx
```diff
@@ -59,29 +81,43 @@ export class SectionVideo extends Component {
   }
 
   renderVideoEmbed () {
-    const { section, article, editing } = this.props
-    if (section.get('url') && section.get('url').length) {
+    const { section, article, editing, headerType } = this.props
+    const isBasicHeader = headerType === 'basic-embed'
+    const hasUrl = Boolean(section.get('url'))
```
eessex: Because 'basic-embed' refers to hiding the preview, it might make sense to use the prop as a boolean rather than a prop specific to the header layout -- i.e.  `<SectionVideo hidePreview />`

jsconfig.json
eessex: :+1:

eessex:
Looks good!  A minor comment about naming, excited to see this in action.  Remove is a minor priority so we can definitely save that for later.

client/apps/edit/components/content2/index.coffee
damassi: Commenting that line out indeed fixes the issue as a listener is no longer registered, but I can't find the root -- meaning, the point in the code that is updating the model every 1 second, triggering this callback.

client/apps/edit/components/content2/sections/header/controls/VideoControls.jsx
damassi: Can you elaborate on what an artsy "channel" is? Isn't everything technically a channel since content can't exist outside of one?

client/apps/edit/components/content2/sections/video/index.jsx
damassi: Ah yes, `hidePreview` is a much better way of doing this 👍

client/apps/edit/components/content2/sections/header/controls/VideoControls.jsx
eessex: A channel is, for lack or a better term, the owner of the article -- for example 'artsy editorial', or 'artsy auctions' -- different channels have different features and permissions available to them.

PR #1375, opened 2017-11-01
@damassi => Move `babel-plugin-transform-decorators-legacy` to `dependencies`

kanaabe:
I had this as a devDependency but it looks like I crashed staging. Sending this your way. Also as a 👋  since Slack is down 😄


damassi:
🍻

PR #1379, opened 2017-11-01
@eessex => Fixes authors query to return null when no authors_ids

kanaabe:
If there are no `author_ids`, we should resolve with null which avoids rendering the Author component. Right now, a query like: 
```
{
  article(id:"editorial-heraldic-patterns-and-funky-variances-in-polly") {
    authors{
      id
      name
    }
  }
}
```

Returns 10 of the latest authors since the null value defaults to a `find({})` on the Authors collection.


eessex:
Looks good to me!

PR #1390, opened 2017-11-02
@kanaabe => Remove edit2 references

eessex:
A soft launch -- this turns on edit2 for everyone, and removes the dual routing system. 

In a separate PR I will `content2`, it's going to require a fine tooth comb and don't want to rush deleting tons of code for tonight.


kanaabe:
Looks good to me! I think this is much safer -- good call!

PR #1391, opened 2017-11-03
Revert "@kanaabe => Remove edit2 references"

kanaabe:
Reverts artsy/positron#1390


kanaabe:
Reverting this so we can deploy it tomorrow -- have the IA PR that needs to go out tonight so this is currently blocking.

PR #1404, opened 2017-11-06
@kanaabe => Add cover_img_url to GraphQL query

damassi:
Addresses https://github.com/artsy/publishing/issues/221. 

Adds a new `cover_img_url` to query and fixes an issue with zero results in display fetches.


kanaabe:
👏 Nice! Thanks for the test also.

PR #1405, opened 2017-11-07
@craigspaeth => Surpass login for static display

kanaabe:
Shuffles middleware around so login is not required for static display panels to work.

### DIFF
--- a/client/lib/setup/authorization.js
+++ b/client/lib/setup/authorization.js
@@ -1,0 +1,16 @@
+
+export const requireLogin = (req, res, next) => {
+  if (req.user) {
+    next()
+  } else {
+    res.redirect('/login')
+  }
+}
+
+export const requireChannel = (req, res, next) => {
+  if (req.user && req.user.get('current_channel')) {
+    next()
+  } else {
+    res.redirect('/logout')
+  }
+}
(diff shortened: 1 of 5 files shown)


client/lib/setup/authorization.js
```diff
@@ -0,0 +1,16 @@
+
+export const requireLogin = (req, res, next) => {
```
craigspaeth: :+1: to refactoring to JS!

craigspaeth:
LGTM feel free to merge when green!

client/lib/setup/authorization.js
```diff
@@ -0,0 +1,16 @@
...
+  if (req.user) {
+    next()
+  } else {
+    res.redirect('/login')
+  }
+}
+
+export const requireChannel = (req, res, next) => {
+  if (req.user && req.user.get('current_channel')) {
```
craigspaeth: Because we require a user _and_ a channel here, the above middleware is actually redundant.

kanaabe: Yeah, that's a good point!

kanaabe: Mm there's a subtle difference - no user means we should redirect to the login screen, whereas no channel means we should log the user out. Logging the user out might be a little extreme for having no channel, but there is a distinction. The logout helps in case the user's session is stale.

craigspaeth: Ah, good call! Thanks for pointing that out.

PR #1408, opened 2017-11-07
Turn on display ads in Instant Articles

kanaabe:
Turns on display, moves to manual placement, uses Force segment key for sending display events


craigspaeth:
Woot! Code LGTM, thanks!

PR #1409, opened 2017-11-07
@kanaabe => Gucci admin

eessex:
Adds admin page for Gucci feature.

Additionally, 
- Save button is moved to its own component that can be used in curation admin pages.
- Moved venice stylesheet into its own directory

screen shot 2017-11-07 at 11 29 27 am


client/apps/settings/client/curations/gucci/components/series.jsx
```diff
@@ -0,0 +1,61 @@
...
+    <div className='admin-form-container series'>
+      <Row>
+        <Col lg={8}>
+          <div className='field-group'>
+            <label>About the Series</label>
+            <div className='bordered-input'>
+              <Paragraph
+                html={curation.get('about') || ''}
+                linked
+                layout='classic'
```
kanaabe: Interesting, could it be worth adding a default layout on the `Paragraph` component? It seems like this could get confusing down the line, like the concept of 'classic' doesn't really apply to this Gucci curation. Not necessary, but on instinct this feels a little hacky -- maybe it just needs to be renamed?

client/apps/settings/client/curations/gucci/index.jsx
```diff
@@ -0,0 +1,86 @@
...
+      curation: props.curation,
+      isSaved: true,
+      activeSection: null
+    }
+  }
+
+  setActiveSection = (i) => {
+    i = i === this.state.activeSection ? null : i
+    this.setState({activeSection: i})
+  }
```
kanaabe: This is more a suggestion for the future (since we have an insane deadline), but I was hoping to see less repeated code throughout these Admin curations. For example, we've seen [this method](https://github.com/artsy/positron/blob/master/client/apps/settings/client/curations/display/index.jsx#L19-L22) implemented a couple times already. This admin, Venice, and YIA share a lot of functionality that could benefit by having a "shell admin" that gets extended by each of the custom ones. This would be a fun one to pair on!

client/apps/settings/client/curations/gucci/test/section.test.js
```diff
@@ -0,0 +1,107 @@
...
+
+  it('Updates published on checkbox click', () => {
+    const component = mount(
+      <SectionAdmin {...props} />
+    )
+    component.find('.flat-checkbox').first().simulate('click')
+    expect(props.onChange.mock.calls[5][0]).toMatch('published')
+    expect(props.onChange.mock.calls[5][1]).toBe(false)
+  })
+})
```
kanaabe: ❤️ These ❤️ Tests ❤️ Are ❤️ Great ❤️

kanaabe:
Great, this looks ready to merge. Feel free to address my comments or not.

Also a reminder to update Heroku with the new `EF_GUCCI` env var.

client/apps/settings/client/curations/gucci/components/series.jsx
eessex: This just changes the font sizes on the component in the UI -- which defaults to the editorial font-size.  Could just change the default size as well to be classic.

client/apps/settings/client/curations/gucci/index.jsx
eessex: Yeah I think we need a container for the dropdown component -- something similar to how the dropDown list works, where these controls can wrap everything inside.  Venice is actually slightly different because it allows multiple panels to be different. Figured it was more labor than makes sense for this timeline but happy to look at this together soon.

PR #1410, opened 2017-11-07
@eessex => Bump reaction and add isMobile flag to static Display

kanaabe:
#minor


eessex:
This looks good to me when 🍏

PR #1415, opened 2017-11-08
Add Webpack

damassi:
This PR adds Webpack to the mix!

- New React code gets sub-second hot reloads, and preserves state
- Nice errors
- Old CoffeeScript code will automatically reload browser on change
- Can now `import 'some-styl-file.styl'` if needed


client/components/autocomplete_channels/index.coffee
```diff
@@ -1,5 +1,6 @@
 _ = require 'underscore'
 Backbone = require 'backbone'
+# TODO: Verify that this still works
 Modal = -> require('simple-modal') arguments...
```
damassi: @kanaabe / @eessex - could one of you verify that this modal still works and get back to me? There was a weird issue that i ran into due to the library (`simple-modal`) using `cssify` internally. I had to exclude it in root `index.js`. But not sure if the css still applies to it

client/lib/webpack-dev-server.js
```diff
@@ -0,0 +1,35 @@
...
+    <head>
+      <title>Webpack Test</title>
+    </head>
+    <body>
+      <div id='react-root' />
+      <script src='/assets/webpack.js'></script>
+    </body>
+  </html>
+  `)
+})
```
damassi: Can use this route `http://localhost:3005/webpack` to test out new configurations

index.js
```diff
@@ -1,3 +1,4 @@
 require('babel-core/register')
 require('coffee-script/register')
+require('ignore-styles') // Fix for require('simple-modal')
```
damassi: See above re `simple-modal`

boot.js
```diff
@@ -9,28 +9,38 @@ import path from 'path'
...
+  id: ARTSY_ID,
+  secret: ARTSY_SECRET
 }
 
 artsyXapp.init(xappConfig, () => {
   app.use(newrelic)
 
   if (isDevelopment) {
-    const reloadAndMount = createReloadable(app, require)
+    app.use(require('./client/lib/webpack-dev-server'))
```
craigspaeth: 👍  nice sub-app pattern use here.

webpack.config.js
```diff
@@ -0,0 +1,175 @@
...
+      return {
+        ...assetMap,
+        ...asset
+      }
+    }, {})
+
+  return assets
+}
+
+module.exports = config
```
craigspaeth: Oh, Webpack and its long config files.

craigspaeth:
Awesome! Super excited to have this upgraded DevX in Positron 💯

webpack.config.js
damassi: LOL yeah -- especially when `$ ezel-assets` with all of its sensible defaults and extensibility via `.babelrc` is two words! The new webpack docs are unreal, too -- its intentionally powerful, but I have literally zero idea how they've brainstormed all of the possible usecases.

eessex:
Wow, so awesome!  Thank you @damassi, can't wait to try this out.

client/components/autocomplete_channels/index.coffee
kanaabe: I'm getting 
image
when I pull this down locally and open the Switch Channel modal. You can replicate it by hovering here:
image

client/components/error/client.coffee
```diff
@@ -1,4 +1,5 @@
 Backbone = require 'backbone'
+# TODO: Verify that this still works
```
kanaabe: Looks like we aren't using this component at all -- feel free to remove it!

client/components/autocomplete_channels/index.coffee
kanaabe: Maybe I'm missing something?

PR #1418, opened 2017-11-08
@eessex => Gucci metadata

kanaabe:
Addresses points in: https://waffle.io/artsy/publishing/cards/5a032dec1b23d70089e701fd

### DIFF
--- a/client/apps/settings/client/curations/gucci/test/section.test.js
+++ b/client/apps/settings/client/curations/gucci/test/section.test.js
@@ -24,7 +24,6 @@ describe('Section Admin', () => {
     thumbnail_image: 'http://thumbnail.jpg',
     email_image: 'http://emailimage.jpg',
     social_image: 'http://socialimage.jpg'
-
   }
 
   beforeEach(() => {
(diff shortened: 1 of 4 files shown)


eessex:
This would cover individual videos, I wonder if we want info for the series at large as well?

kanaabe:
Yess good idea. I'll add that in

client/apps/settings/client/curations/gucci/test/section.test.js
```diff
@@ -6,40 +6,64 @@ import ImageUpload from 'client/apps/edit/components/admin/components/image_uplo
...
+    thumbnail_image: 'http://thumbnail.jpg',
+    email_image: 'http://emailimage.jpg',
+    social_image: 'http://socialimage.jpg'
   }
 
+  beforeEach(() => {
+    props = {
+      section,
+      onChange: jest.fn()
+    }
```
eessex: Good call on putting the mock here

eessex:
Looks great, thanks!

PR #1426, opened 2017-11-09
@eessex => Refactor Gucci Admin metadata and add description

kanaabe:
DRYs up a bit of the metadata work and adds in a general `description` field.


eessex:
Excellent, much cleaner!

PR #1430, opened 2017-11-10
@eessex => Require underscore in a few missing places

kanaabe:
I'm so curious why this just started happening, but let's just fix this for now -- I wonder if with our former Browserify setup `underscore` was a globally available package or something. Seems like an odd thing to start happening now and Webpack is really the only change. 🤔  

Addresses: https://waffle.io/artsy/publishing/cards/5a05f76ee3cf1200c6a673b3


eessex:
LGTM for merge on 🍏  
Yeah, seems like it must be related to Webpack transition -- got a few of these alerts right after merge.

PR #1433, opened 2017-11-13
@eessex => Fixes related articles panel links

kanaabe:
Addresses https://waffle.io/artsy/publishing/cards/59fc727221451400241101bf

It looks like we were wrapping the `articleFeed` results with the presenter, but missed the ones fetched from `related_article_ids`.


eessex:
This looks good to me!

Issue #1412, opened 2017-11-07
DropDown list component


eessex:
A wrapper to contain dropdown headers and their components, that can control what panels are open or not.

PR #1436, opened 2017-11-14
@kanaabe => Dropdown list component

eessex:
Adds component DropDownList, which wraps children the existing DropDownItem component, and handles the open/closed state for each item. 

Also updates Gucci and Display admin to use this component.

Closes https://github.com/artsy/positron/issues/1412


client/components/drop_down/drop_down_item.jsx
```diff
@@ -0,0 +1,67 @@
...
+  to {
+    max-height: 100%;
+    opacity: 1;
+  }
+`
+const DropDownItemContent = styled.div`
+  animation-name: ${navKeyframes};
+  animation-duration: 1s;
+  animation-timing-function: ease;
+`
```
kanaabe: In trying to debug the Gucci header, I found that the use of animation wasn't really necessary for these kinds of CSS movements. When you have two states -- closed + open item, it's a lot easier to manage something like `data-state=open` with `transition` instead. With this, we only get an animation on open, and not for the close. We'd have to create a second animation for the closed state. It also causes unnecessary re-renders of the items.

kanaabe:
I had one comment on the animation, but I think this is good to go otherwise! It's up to you if you want to change it or not.

client/components/drop_down/drop_down_item.jsx
eessex: Tried doing it with data-state, but still ran into some issues because elements are being added/removed from the DOM.  I think https://github.com/reactjs/react-transition-group might offer a solution, but going to save for another time.

kanaabe: Yeah, what I'm proposing is to avoid adding and removing the elements from the DOM. So there's just a simple class or data toggle that shows or hides content with a transition.

client/apps/settings/client/curations/display/index.jsx
```diff
@@ -1,7 +1,7 @@
 import { set } from 'lodash'
 import PropTypes from 'prop-types'
 import React from 'react'
-import DropDownItem from 'client/components/drop_down/index.jsx'
+import { DropDownList } from 'client/components/drop_down/drop_down_list.jsx'
```
damassi: Note that you no longer need to append extensions: https://github.com/artsy/positron/blob/master/webpack.config.js#L111

PR #1441, opened 2017-11-14
@kanaabe => Last section tool is visible

eessex:
Fixes visibility for the last section tool -- also added a check in the component for the last section, and renamed the attr to `data-visible` as opposed to `data-first`.


kanaabe:
Woo! I predict a 10% decrease in support requests 😛

PR #1454, opened 2017-11-20
@damassi => Turn on parallel, use --runInBand for Jest tests

kanaabe:
This turns parallel tests back on for Circle and makes use of `--runInBand` which prevents Jest tests from spawning new processes.

cc @eessex

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "start": "sh scripts/start.sh",
     "dev": "sh scripts/dev.sh",
     "test": "sh scripts/test.sh",
-    "test:watch": "yarn jest -- --watch",
+    "test:watch": "yarn jest -- --watch --runInBand",
     "assets": "sh scripts/assets.sh",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "start": "sh scripts/start.sh",
     "dev": "sh scripts/dev.sh",
     "test": "sh scripts/test.sh",
-    "test:watch": "yarn jest -- --watch",
+    "test:watch": "yarn jest -- --watch --runInBand",
```
damassi: @kanaabe - I think this needs to be on line 11? `test:watch` is just for development mode and should never be run in CI

kanaabe: Yeahh I added it to both. See `scripts/test.sh`

damassi: Ooh 👍

damassi:
LGTM 👍

PR #1464, opened 2017-12-04
@kanaabe => Use Redux for Edit app

eessex:
Reworks edit app to use Redux for actions in the Edit app, in order to maintain a single and consistent article and state between:
- edit/content
- edit/header
- edit/admin
- edit/display

standard-redux

PR #1468, opened 2017-12-05
@kanaabe => Edit header uses React over Backbone

eessex:
Per suggestion of @damassi, this is the first of a few PRs breaking out a pieces from the [WIP Redux PR](https://github.com/artsy/positron/pull/1464/files). 

This code replaces `edit/header`, using React rather than a Backbone view.  It is set up to accept Redux events as props.  This is not yet used and will be implemented in the Redux PR.

### DIFF
--- a/client/apps/edit/components/header2/index.jsx
+++ b/client/apps/edit/components/header2/index.jsx
@@ -1,0 +1,188 @@
+import PropTypes from 'prop-types'
+import React, { Component } from 'react'
+import Icon from '@artsy/reaction-force/dist/Components/Icon'
+import { data as sd } from 'sharify'
+
+export class EditHeader extends Component {
+  static propTypes = {
+    actions: PropTypes.any,
+    article: PropTypes.object,
+    channel: PropTypes.object,
+    edit: PropTypes.object,
+    user: PropTypes.object
+  }
+
+  toggleSection = (activeSection) => {
+    // TODO - connect component to use redux actions directly
+    this.props.actions.changeSection(activeSection)
+  }
+
+  isPublishable = () => {
+    return this.contentIsComplete() && this.displayIsComplete()
+  }
+
+  onPublish = () => {
+    const { actions, article } = this.props
+
+    if (this.isPublishable()) {
+      actions.publishArticle(
+        article,
+        !article.get('published')
+      )
+    }
+  }
+
+  onDelete = () => {
+    const { actions, article } = this.props
+
+    if (confirm('Are you sure?')) {
+      actions.deleteArticle(article)
+    }
+  }
+
+  getSaveColor = () => {
+    const { isSaving, isSaved } = this.props.edit
+
+    if (isSaving) {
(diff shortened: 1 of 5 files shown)


client/apps/edit/components/header2/index.jsx
```diff
@@ -0,0 +1,188 @@
...
+            <button
+              className='avant-garde-button check'
+              onClick={() => this.toggleSection('content')}
+              data-active={activeSection === 'content'}
+            >
+              <span>Content</span>
+              <Icon
+                className='icon'
+                name='check'
+                color={this.contentIsComplete() ? 'limegreen' : '#ccc'}
```
kanaabe: Hmm I think we could do a little better with keeping colors named. I saw it in another PR too, but I generally try to stay away from using hex and stick to colors that are in our style guide. We have a Colors export from Reaction that can be used here.

client/apps/edit/components/header2/index.jsx
```diff
@@ -0,0 +1,188 @@
...
+            <button
+              className='avant-garde-button check'
+              onClick={() => this.toggleSection('display')}
+              data-active={activeSection === 'display'}
+            >
+              <span>Display</span>
+              <Icon
+                className='icon'
+                name='check'
+                color={this.displayIsComplete() ? 'limegreen' : '#ccc'}
```
kanaabe: Same here

client/apps/edit/components/header2/index.jsx
```diff
@@ -0,0 +1,188 @@
...
+                Auto-link
+              </button>
+            }
+          </div>
+        </div>
+
+        <div className='EditHeader__right'>
+          <button
+            className='avant-garde-button'
+            style={{border: 'none'}}
```
kanaabe: The styling seems to be pulling in different directions. In this file alone, I'm seeing three different ways of styling. There's our legacy .styl, then we have this `style={}` format, and then finally we have an Icon that uses styled-components. I feel like at this point for new code we should be sticking to one or two ways.

client/apps/edit/components/header2/index.jsx
```diff
@@ -0,0 +1,188 @@
...
+            className='avant-garde-button'
+            style={{color: this.getSaveColor()}}
+            onClick={() => article.get('published')
+              ? article.trigger('savePublished')
+              : actions.saveArticle(article)
+            }>
+            {this.getSaveText()}
+          </button>
+
+          <a href={`${sd.FORCE_URL}/article/${article.get('slug')}`} target='_blank'>
```
kanaabe: There's a model method for this! You'll want to double check but I think it's `article.getFullSlug()`

eessex: Hot tip!  The function isn't being any used anywhere except now here. Updated to use underscore.string's slugify, which is what we use in API (some special characters were showing in slug).
Also found the 'contentIsComplete' and 'displayIsComplete' existed on the model.

Issue #1459, opened 2017-11-29
Command-i glitch for italicizing captions in Writer


sarahdotson:
When I use command-i to try and italicize text in captions in Writer, it deletes all previous text I typed and doesn't italicize what I type following the command. 

If I highlight the portion I want italicized and use command-i, it does the same thing (deletes all text but the highlighted text)

If I type out a caption, highlight the portion I want, and then use the button for italics in Writer, it works. I never need to use bolded text, but the same goes for command-B

sarahdotson:
Hi, I am still experiencing this issue!

PR #1473, opened 2017-12-07
@kanaabe => Decaf Text nav

eessex:
A new jsx menu for the text editor.
Also addresses https://github.com/artsy/positron/issues/1459


kanaabe:
Woohoo! Looks good 👍

PR #1474, opened 2017-12-08
@craigspaeth => Refactor authorization to use JWT for partner_ids

kanaabe:
## Problem
The original issue was that users who had access to more than 10 partners returned a non-sorted array of partners in `api.artsy.net/v1/api/me/partners`. This caused a redirect loop because the client-side's [`user.isOutdated`](https://github.com/artsy/positron/blob/master/client/components/layout/client.coffee#L40) check would sometimes not match the stored user.

## Solution
We could have paginated this API, but @joeyAghion enlightened me that JWT's blob actually returns `partner_ids`. This field returns the partners a user has access to, and not necessarily the ones it manages (`api.artsy.net/v1/api/me/partners` returns both). However, since all Admins have access to partners anyway, we only need to cover the "user has access to" case, which is primarily used for non-admin Artsy users who manage partner accounts.


api/apps/users/model.coffee
```diff
@@ -48,20 +48,18 @@ bcrypt = require 'bcryptjs'
...
     (cb) ->
       bcrypt.hash accessToken, SALT, cb
   ], (err, results) ->
     return callback err if err
-    user.partner_ids = _.map (results[0]?.body or []), (partner) ->
-      ObjectId partner._id
-    user.channel_ids = _.pluck results[1], '_id'
-    encryptedAccessToken = results[2]
+    partner_ids = jwtDecode(accessToken)?.partner_ids or []
+    user.partner_ids = _.map partner_ids, (id) ->
```
craigspaeth: Code ⛳️  trick, this can be: `user.partner_ids = _.map partner_ids, ObjectId`

client/lib/setup/auth.coffee
```diff
@@ -22,7 +23,8 @@ setupPassport = ->
       headers: 'X-Access-Token': accessToken
       error: (m, err) -> done err
       success: (user) ->
-        id = user.get('channel_ids').concat(user.get('partner_ids'))[0]
+        partner_ids = jwtDecode(accessToken).partner_ids
```
craigspaeth: I don't think the client-side code actually has to be changed because you're already decoding the JWT and saving the partner ids on the server [here](https://github.com/artsy/positron/pull/1474/files#diff-2a1090257f616ec3ff944746d030cda6R58).

craigspaeth:
Yay for JWTs! Glad to see Positron joining the latest in auth practices. I left some non-critical comments that you can feel to address/not and merge when ready. Good stuff!

client/lib/setup/auth.coffee
kanaabe: Ooh yeah, you're totally right. This one is unnecessary. There is on client-side spot [here](https://github.com/artsy/positron/pull/1474/files#diff-ae3098d32d28ef9fa7bc649e32f5482fR48) that I think we will need to keep since it's where we check that the user stored in the session is in sync with the one stored in db.

PR #1477, opened 2017-12-11
@kanaabe => [Article Display] Email preview bug

eessex:
Fixes bug email panel did not display for articles without `email_metatdata` object 
https://sentry.io/artsynet/positron-production/issues/416672960/

Also deprecates email `credit_url` and `credit_line` fields.

### DIFF
--- a/api/apps/articles/model/distribute.coffee
+++ b/api/apps/articles/model/distribute.coffee
@@ -98,8 +98,6 @@ postSailthruAPI = (article, cb) ->
   images: images
   spider: 0
   vars:
-    credit_line: article.email_metadata?.credit_line
-    credit_url: article.email_metadata?.credit_url
     html: html
     custom_text: article.email_metadata?.custom_text
     daily_email: article.daily_email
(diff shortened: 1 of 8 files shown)


kanaabe:
Hmm why do we want to deprecate `credit_url` and `credit_line`?

eessex:
Had a conversation about this last week w/ @owendodd when rebuilding display, they are no longer used in the emails.

kanaabe:
If we do want to truly deprecate those, they should be removed from the API as well!

eessex:
Good point, will do

kanaabe:
Cool thanks! Feel free to merge on green.

craigspaeth:
Worth noting following [that minor outage](https://artsy.slack.com/archives/C02CAB1KW/p1513041611000096) that GraphQL has a first-class [deprecation](https://medium.com/@kevinsimper/how-to-deprecate-fields-in-graphql-52fbd03fb9d7) feature in which one could tag fields as deprecated without removing them and introducing breaking changes right away. That said, often we just need to be careful to deploy updated clients first before dropping fields from APIs that they might use.

PR #1479, opened 2017-12-12
@kanaabe => Allow empty hero

eessex:
Was failing on null hero.


eessex:
Realized that the file this is in does not yet have a test -- making an issue for this!

PR #1483, opened 2017-12-13
@kanaabe => Series edit

eessex:
WIP because still trying to get some tests working on the autocomplete (having trouble with es6 refs combined with mocked document properties).

- Adds new directory in edit - `content/article_layouts`, with parent containers/styles for series and default article
- Adds `series` layout and `series_description` to article schema
- Adds series to new article menu
- Adds RelatedArticleList component, to display a list of ArticleCards and autocomplete input
- Adds Autocomplete2, an es6 version of Bloodhound component that accepts JSX for result formats
- Replaces instances of named layout classes with [data-layout='series'] etc.

RelatedArticleList was set up with the intention of being able to edit articles inline, but for the sake of our deadline they can only be rearranged or deleted ATM.  

series-edit-img


client/apps/edit/components/content/article_layouts/series.jsx
```diff
@@ -0,0 +1,95 @@
...
+        className='EditSeries'
+        url={url}
+      >
+
+        <div className='EditSeries__bg-input'>
+          <FileInput
+            type='simple'
+            onUpload={(src) => onChangeHero('url', src)}
+            prompt={`+ ${url ? 'Change' : 'Add'} Background`}
+            onProgress={() => console.log('uploaded')}
```
kanaabe: oo is this an intentional log?

client/apps/edit/components/content/article_layouts/series.jsx
```diff
@@ -0,0 +1,95 @@
...
+  }
+}
+
+export const EditSeriesContainer = styled.div`
+  ${props => props.url && `
+    background-image: url(${props.url});
+    background-size: cover;
+    background-position: 50%;
+  `}
+`
```
kanaabe: Really love how this translates to the component w/props. What are your thoughts/preferences after doing both children and props?

client/apps/edit/components/content/index.jsx
```diff
@@ -13,42 +11,28 @@ export class EditContent extends Component {
...
+
+    switch (article.get('layout')) {
+      case 'series': {
+        return <EditSeries {...this.props} />
+      }
+      default: {
+        return <EditArticle {...this.props} />
+      }
+    }
+  }
```
kanaabe: Cool cool

client/apps/edit/components/content/sections/related_articles/index.jsx
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  fetchArticles = () => {
+    const { related_article_ids } = this.props.article.attributes
+    let relatedArticles = []
+
+    if (related_article_ids && related_article_ids.length) {
+      return async.each(related_article_ids, (id, cb) => {
+        return request
+          .get(`${sd.API_URL}/articles/${id}`)
```
kanaabe: Ooh we should be using GraphQL! It removes the need for using `async.each` as well. It might be worth merging #1484 first to get it working otherwise you can always pass in `ids` to the normal articles resource.

Even with this REST way, you can fetch with `${sd.API_URL}/articles?ids=[123]` to avoid that `async.each`.

client/apps/edit/components/content/article_layouts/series.jsx
eessex: Definitely a good suggestion to move to props, brings a lot of clarity on both sides.

client/apps/edit/components/content/sections/related_articles/index.jsx
```diff
@@ -0,0 +1,151 @@
...
+        relatedArticles
+      })
+    }
+  }
+
+  onAddArticle = (related_article_ids) => {
+    const { onChange } = this.props
+
+    onChange('related_article_ids', related_article_ids)
+    this.fetchArticles()
```
kanaabe: Interesting, is it necessary to fetch all the articles each time one is added or removed? Ideally we should be able to get the data we need to render the card from the `RelatedArticlesInput` autocomplete so it seems like this additional fetch is redundant.

I see you've added a new version of `Autocomplete` that is designed for an array of `ids` that this component uses though. I think we could rework it a bit to leverage more from the fetch. I don't know how easy this change would be and I'm definitely not asking to make this change now, but this does feel like a worthy optimization for the future!

eessex: Yeah not ideal to fetch all all the time -- and definitely don't need to fetch on removing.

api/apps/articles/model/schema.coffee
```diff
@@ -98,14 +98,15 @@ ImageCollectionSection = (->
...
-  layout: @string().allow('classic', 'standard', 'longform', 'feature').default('classic')
+  layout: @string().allow('classic', 'standard', 'series', 'feature').default('classic')
   updated_at: @date()
   published: @boolean().default(false)
   published_at: @date().allow(null)
   scheduled_publish_at: @date().allow(null)
   lead_paragraph: @string().allow('', null)
   gravity_id: @string().objectid().allow('', null)
   hero_section: @alternatives().try(videoSection, ImageCollectionSection, imageSection, featureSection).allow(null).default(null)
+  series_description: @string().allow(null)
```
kanaabe: Ooh also note we should add tests for these new fields.

client/apps/edit/components/content/sections/related_articles/index.jsx
eessex: This fetch is now updated to use graphql and not repeat itself

kanaabe:
Changes look good! Thanks.

PR #1484, opened 2017-12-13
@kanaabe => Query related articles for series

eessex:
Adds api support for querying all related articles attached to an article.


kanaabe:
Ok niceee code looks good but this isn't anything specific to the Series, right? For example, the video will use the same query, ie the `related_article_ids` field. Do you think it'd make more future sense to rename this to something more generic like `relatedArticles`?

eessex:
Good point, will do!

PR #1485, opened 2017-12-13
@craigspaeth => Add check for empty email_metadata

eessex:
Email metadata form was failing when article has no email_metadata attribute, adds an empty object if this is the case. 

Related convo - https://artsy.slack.com/archives/C04GGGDCM/p1513198964000377

### DIFF
--- a/api/apps/articles/model/schema.coffee
+++ b/api/apps/articles/model/schema.coffee
@@ -178,11 +178,11 @@ ImageCollectionSection = (->
       id: @string().objectid().allow(null)
       name: @string().allow('', null)
   ]).default([])
-  email_metadata: @object().keys
-    image_url: @string().allow('',null)
-    headline: @string().allow('',null)
-    author: @string().allow('',null)
-    custom_text: @string().allow('',null)
+  email_metadata: @object().default({}).keys
+    image_url: @string().allow('')
+    headline: @string().allow('')
+    author: @string().allow('')
+    custom_text: @string().allow('')
   is_super_article: @boolean().default(false)
   super_article: @object().keys
     partner_link: @string().allow('',null)
--- a/client/apps/edit/components/display/components/email.jsx
+++ b/client/apps/edit/components/display/components/email.jsx
@@ -13,7 +13,7 @@ export class DisplayEmail extends Component {
 
   onChange = (key, value) => {
     const { article, onChange } = this.props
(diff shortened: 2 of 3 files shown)


craigspaeth:
This is a good example of comments I've left in the past about defaulting to "empty" data types instead of allowing nullable fields. In this case, we might want to be more strict at the [API schema level](https://github.com/artsy/positron/blob/master/api/apps/articles/model/schema.coffee#L181). Instead of `.allow(null)` everywhere, maybe we instead do:

````coffeescript
  email_metadata: @object().default({}).keys
    image_url: @string().default('')
    headline: @string().default('')
    author: @string().default('')
    custom_text: @string().default('')
````

around [here](https://github.com/artsy/positron/blob/master/api/apps/articles/model/schema.coffee#L181). In this case, we're modeling that all articles must have email metadata—just some of that data might be not filled in yet.

Let me know if that doesn't work easily/make sense. I think it's worth doing this refactor to the API side though to ensure data integrity in the database.

**Comment by eessex** [comment]
Yes makes sense to me, updated to include default in the schema.

File: client/apps/edit/components/display/components/email.jsx [hunk 1]
```diff
@@ -13,7 +13,7 @@ export class DisplayEmail extends Component {
 
   onChange = (key, value) => {
     const { article, onChange } = this.props
-    const emailMetadata = article.get('email_metadata')
+    const emailMetadata = article.get('email_metadata') || {}
```
craigspaeth: This part and the test below shouldn't be necessary with the API changes.

api/apps/articles/model/schema.coffee
```diff
@@ -178,7 +178,7 @@ ImageCollectionSection = (->
       id: @string().objectid().allow(null)
       name: @string().allow('', null)
   ]).default([])
-  email_metadata: @object().keys
+  email_metadata: @object().default({}).keys
     image_url: @string().allow('',null)
```
craigspaeth: Ideally, we'd un-null these too, so...

````coffeescript
    image_url: @string().allow('')
    headline: @string().allow('')
    author: @string().allow('')
    custom_text: @string().allow('')
````

craigspaeth:
Thanks for the update! Left some minor comments, but the important part is in. Feel free to address and merge when ready.

PR #1486, opened 2017-12-13
@kanaabe => Use actual slug to construct full slug

eessex:
The getFullSlug on the article model was not constructing URLs correctly, which is why users were having trouble previewing. (Currently the method is only used in the preview button). 

This opts to use the article's existing slug rather than constructing one. 

We can also at some point transition to this - https://github.com/artsy/reaction/pull/415/files#diff-01463a493f089d969a5da6fe564bf93eR31


kanaabe:
😱 ! Good find!!

PR #1490, opened 2017-12-18
@eessex => Video edit

kanaabe:
With video previewsss

video-edit


client/apps/edit/components/content/article_layouts/video.jsx
```diff
@@ -0,0 +1,234 @@
...
+      <EditVideoContainer>
+        {uploadProgress &&
+          <ProgressBar progress={uploadProgress} />
+        }
+
+        <VideoPreview visible={this.state.showVideo}>
+          <VideoPlayer
+            url={media.url}
+            onLoadedMetadata={this.setDuration}
+          />
```
eessex: setDuration and toggleVideoPreview are declared as component methods but could easily be set inline like on 131 and 133, for consistencies sake.

api/apps/articles/model/schema.coffee
```diff
@@ -157,6 +157,15 @@ ImageCollectionSection = (->
       layout: @string().allow('full', 'mini', null)
       images: @array().items([denormalizedArtwork, imageSection])
   ]).allow(null)
+  media: @object().keys(
```
eessex: In a few conversations @craigspaeth has pushed for not using null in the schema, and giving an empty object or string instead.  Not everything has a media blob, so could go either way, but it would allow you to avoid checks for something existing like here - https://github.com/artsy/positron/pull/1490/files#diff-e0767deec1a38e838fdd9585a07a9720R86

client/apps/edit/components/content/article_layouts/video.jsx
```diff
@@ -0,0 +1,234 @@
...
+    position: fixed;
+    top: 95px;
+    left: 0;
+    right: 0;
+    z-index: 1;
+  }
+
+  ${VideoCoverContainer}, ${VideoAboutContainer} {
+    .public-DraftEditorPlaceholder-root {
+      position: absolute;
```
eessex: We have to declare `.public-DraftEditorPlaceholder-root` positioning frequently, but it is almost always the same -- it might make sense to move this to a general stylesheet.

api/apps/articles/model/schema.coffee
eessex: Also thinking we might want to add an additional field of `series_article_for` similar to the `super_article_for`. Currently we don't have an easy way to find the series from its children, but we need to fetch it for the slug.

kanaabe: Hm yeah that works well if we backfill articles with empty objects because past articles will always be missing the field and GraphQL will return `null`. The null vs empty is something I've gone back and forth with a lot because those empty states tend to bloat the document size but for this case, an empty media object sounds reasonable since it's the core content of some articles. We should probably give `sections` an empty array default as well. 

As for `series_article_for`, this should be added as a query schema and not an input. It's possible we might only expose this on our GraphQL endpoint as well. Since it requires some aggregation I'll leave that for another PR.

client/apps/edit/components/content/article_layouts/video.jsx
kanaabe: Is it safe to make this a default? I'm happy to, but I'm not as familiar with the downstream effects of other components that use this.

eessex: I think it should be fine -- the only place that it is not positioned like above is in the classic and basic titles, where the text is centered.

api/apps/articles/model/schema.coffee
craigspaeth: FWIW I wouldn't worry too much about document size bloat Kana. The current average document size in Positron is ~4kb (which you can see on the [monitoring > status > db.stats() tab](https://5474b8154e676f96c3000017-partners.compose.io/app29923450-heroku-com/mongo/app29923450/status) in compose) and the limit in Mongo is 16mb. As long as we're considerate of the [six rules to designing Mongo schemas](https://www.mongodb.com/blog/post/6-rules-of-thumb-for-mongodb-schema-design-part-3?_ga=2.196332453.888925283.1513627255-1314794917.1513627253) then we'll likely never have to worry about documents getting too large.

I do think liberally allowing nulls was a mistake I made early on in the design of Positron's schema and 👍  to defaulting to empty array for `sections`. The short answer to this is that it forces more uncertainty and over-defensiveness in clients when it can easily be avoided in most cases by defaulting to blank values instead. The dangers of too much nullability have been [written about](https://www.google.com/search?q=null+is+worst+mistake&oq=nu&aqs=chrome.1.69i59l2j69i57j69i60j69i61j69i60.1302j0j1&sourceid=chrome&ie=UTF-8) in depth, and [in database design it's not a sin](https://dba.stackexchange.com/questions/5222/why-shouldnt-we-allow-nulls) to use null but I believe it's better to default to blank unless there's a convincing reason why we should model "a valid state we'd want to model for this field is missing entirely, e.g. an article legitimately might not be scheduled to publish at all, therefore `scheduled_publish_at: null`".

kanaabe:
This is almost ready, just need to add the published media checkbox. Will ping when that's in!

client/apps/edit/components/content/article_layouts/video.jsx
```diff
@@ -0,0 +1,225 @@
...
+        {uploadProgress &&
+          <ProgressBar progress={uploadProgress} />
+        }
+
+        <VideoPreview visible={this.state.showVideo}>
+          <video
+            controls
+            src={media.url}
+            onLoadedMetadata={(e) => this.onMediaChange('duration', e.target.duration)}
+          />
```
kanaabe: Also note that I've switched to using the plain `video` element instead of the player because we need a callback for `onLoadedMetadata` which I realized is not present in the Reaction version of VideoPlayer. Will update that soon after Tuesday.

kanaabe:
This is now updated with the `published` toggle

PR #1491, opened 2017-12-18
@craigspaeth => Exclude non standard or feature articles from instant articles

kanaabe:
We don't want to send `layout: video` or `layout: series` to Instant Articles since they aren't supported yet.

### DIFF
--- a/api/models/article.coffee
+++ b/api/models/article.coffee
@@ -38,6 +38,9 @@ module.exports = class Article extends Backbone.Model
   isFeatured: ->
     @get('featured')
 
+  hasInstantArticle: ->
+    @get('featured') and @get('layout') in ['standard', 'feature']
+
   isEditorial: ->
     @get('channel_id')?.toString() is EDITORIAL_CHANNEL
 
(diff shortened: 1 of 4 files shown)


api/test/models/article.test.coffee
```diff
@@ -38,6 +38,28 @@ describe "Article", ->
         @article.get('sections')[0].images[0].caption.should.equal '<h1>A place for credit</h1>'
         cb()
 
+  describe '#hasInstantArticle', ->
```
craigspaeth: 👍  nice tests!

api/models/article.coffee
```diff
@@ -38,6 +38,9 @@ module.exports = class Article extends Backbone.Model
   isFeatured: ->
     @get('featured')
 
+  hasInstantArticle: ->
```
craigspaeth: 👍  I still find this a clever use of Backbone to add model logic on the server-side. However, I do wonder if it'd be simpler to do a "model library" where we export "pure" functions in this case, e.g.

````javascript
export const hasInstantArticle = (article) => {
  return article.featured && _.contains(["standard", "feature"], article.layout) 
}
````

This more vanilla/functional approach, for instance, feels more natural to the Redux-ey code we're moving towards on the client-side. Just food for though!

craigspaeth:
LGTM! Simple & well tested as usual. I see the in progress label, so feel free to merge when you're ready. My comment about functional stuff isn't meant to block this PR, but just food for thought. Thanks!

PR #1495, opened 2017-12-19
@eessex => Adds release date UI

kanaabe:
Not the ideal UI, but this lets us edit the field on `media`:

image

### DIFF
--- a/client/apps/edit/components/content/article_layouts/test/video.test.js
+++ b/client/apps/edit/components/content/article_layouts/test/video.test.js
@@ -69,4 +69,13 @@ describe('EditVideo', () => {
     component.find(EditVideoPublished).simulate('click')
     expect(props.onChange.mock.calls[0][1].published).toBe(true)
   })
+
+  it('sets release_date on media', () => {
+    const component = mount(
+      <EditVideo {...props} />
+    )
+    const input = component.find('input').at(3)
+    input.simulate('change', { target: { value: '2017-02-07' } })
+    expect(props.onChange.mock.calls[0][1].release_date).toBe('2017-02-07')
+  })
 })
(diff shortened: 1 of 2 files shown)


client/apps/edit/components/content/article_layouts/test/video.test.js
```diff
@@ -69,4 +69,13 @@ describe('EditVideo', () => {
     component.find(EditVideoPublished).simulate('click')
     expect(props.onChange.mock.calls[0][1].published).toBe(true)
   })
+
+  it('sets release_date on media', () => {
+    const component = mount(
+      <EditVideo {...props} />
+    )
+    const input = component.find('input').at(3)
+    input.simulate('change', { target: { value: '2017-02-07' } })
```
eessex: Minor, but i believe finding input[type=date] would work here to give more specificity, in case the order of inputs changes in the UI.

client/apps/edit/components/content/article_layouts/video.jsx
```diff
@@ -146,6 +148,16 @@ export class EditVideo extends Component {
           <label>Video Published</label>
         </EditVideoPublished>
 
+        <EditVideoReleaseDate>
+          <label>Release Date</label>
+          <input
+            type='date'
+            className='bordered-input bordered-input-dark'
+            value={moment(media.release_date).local().format('YYYY-MM-DD')}
+            onChange={(e) => this.onMediaChange('release_date', e.target.value)}
```
eessex: Almost all the dates in positron are saved as [ISO dates](https://github.com/artsy/positron/blob/master/client/apps/settings/client/curations/display/components/campaign.jsx#L6), and reconverted back to this format on the client.

kanaabe: Good call!

client/apps/edit/components/content/article_layouts/test/video.test.js
kanaabe: True true, nice!

kanaabe:
Note that I've bumped the file size limit as well! 100mb is the limit for videos. The videos are only ~4 minutes but they're high res and the lowest I could get without sacrificing quality was about 75mb.

PR #1496, opened 2017-12-19
@eessex => Adds resolver for series article

kanaabe:
This lets us query the seriesArticle within the initial article fetch. Note that the rendering on the Reaction side will have to pull the `seriesArticle` from `article` like: 

```
<Article article={article} seriesArticle={article.seriesArticle} />
```

So we might want to push the `seriesArticle` prop one level down but from a query perspective this should reduce the number of round trips we need to make to render.

image

### DIFF
--- a/api/apps/graphql/resolvers.js
+++ b/api/apps/graphql/resolvers.js
@@ -186,7 +186,7 @@ export const relatedArticles = (root) => {
 
 export const relatedArticlesPanel = (root) => {
   const omittedIds = [ObjectId(root.id)]
-  if (root.related_articles_ids) {
+  if (root.related_article_ids) {
     omittedIds.concat(root.related_article_ids)
   }
   const tags = (root.tags && root.tags.length > 0) ? root.tags : null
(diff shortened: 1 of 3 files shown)


api/apps/graphql/resolvers.js
```diff
@@ -257,6 +257,25 @@ export const relatedArticlesCanvas = (root) => {
   })
 }
 
+export const seriesArticle = (root) => {
+  return new Promise(async (resolve, reject) => {
```
eessex: Does this mean we loop through every series article to find if it includes a child?

kanaabe: Yep, that's how we have to find super articles too. Thankfully there's just one for now, and the real optimization we can make is the `is_series_sub_article` (which we do for super currently). On the Force side, I'm thinking to only make this fetch for video layouts until we have a `is_series_sub_article` ready.

eessex: is `is_series_sub_article` planned to be a bool or an id?

kanaabe: Mmm I suppose it could be either --

But now that I'm thinking again, the `is_series_sub_article` as a bool or id isn't actually an optimization. Because in order to get that, we still need to loop through all `series` articles to find the `id`. So really the best option is to add the series article as `series_article_id` on the sub article and fetch it that way. But as we've discussed, this involves propagating updates to multiple articles within the Series article which we don't have time for.

PR #1513, opened 2017-12-21
@joeyAghion => Adds a `sections` integration test

kanaabe:
Yesterday's failures could have definitely prevented with this simple check

### DIFF
--- a/api/apps/articles/test/integration.test.coffee
+++ b/api/apps/articles/test/integration.test.coffee
@@ -133,6 +133,10 @@ describe 'articles endpoints', ->
         {
           title: 'Cows on the prarie'
           _id: ObjectId '5086df098523e60002000012'
+          sections: [{
+            type: "text",
+            body: "Cows on the lawn"
+          }]
           published: true
         }
       ], (err, articles) =>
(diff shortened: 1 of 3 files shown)


joeyAghion:
This is great, especially if minor package upgrades could have this impact.

Did you confirm that this test would _fail_ on e3281620498dfef72fa50767b81e21e6e5113141?

kanaabe:
Ha good call, it turns out the dependency issue happens in GraphQL, not our REST endpoints. I'll have to set up another test. I'll keep this one open until then.

joeyAghion:
😅 Glad I asked.

And keep this in any case... it's a good addition.

kanaabe:
@joeyAghion this is now updated with an integration test on our GraphQL endpoint 😄

joeyAghion:
Excellent, thanks.

eessex:
@kanaabe was this working with the react 16 setup?  These are failing again on the new React 16 PR, maybe I missed a dependency of something -- 
screen shot 2018-01-10 at 6 50 26 pm

eessex:
Ah just realized i missed a reverted commit -- this is working now 👍

PR #1515, opened 2018-01-04
@kanaaabe => Autocomplete2 tweaks

eessex:
Some changes to the new autocomplete component:

- Adds a `formatSelected` prop to return a custom object on select
- Use async when returning selected item to allow fetches in parent
- Adds a className prop
- Better loading display on results list

Use case: 
For some uses, we store a list of `ids` which was already supported.  In others, we want to push something else to our items list -- for example, a backbone object in the case of article artworks.

### DIFF
--- a/client/components/autocomplete2/index.jsx
+++ b/client/components/autocomplete2/index.jsx
@@ -1,16 +1,18 @@
 import PropTypes from 'prop-types'
 import React, { Component } from 'react'
-import { uniq } from 'lodash'
+import { uniq, compact } from 'lodash'
 import Icon from '@artsy/reaction-force/dist/Components/Icon'
 
 export class Autocomplete extends Component {
   static propTypes = {
+    className: PropTypes.string,
     disabled: PropTypes.bool,
     filter: PropTypes.func,
     formatResult: PropTypes.func,
     items: PropTypes.array,
     onSelect: PropTypes.func,
     placeholder: PropTypes.string,
+    resObject: PropTypes.func,
     url: PropTypes.string
   }
 
@@ -66,11 +68,32 @@ export class Autocomplete extends Component {
     })
   }
 
-  onSelect = (item) => {
-    const { onSelect, items } = this.props
-    items.push(item._id)
-    onSelect(uniq(items))
-    this.onBlur()
+  getResItem = async (selected) => {
+    const { resObject } = this.props
+
+    try {
+      if (!resObject) {
+        return selected.id || selected._id
+      } else {
+        return await resObject(selected)
+      }
+    } catch (error) {
+      console.error(error)
+    }
+  }
+
+  onSelect = async (selected) => {
+    const { items, onSelect } = this.props
+
+    try {
+      const item = await this.getResItem(selected)
+
+      items.push(item)
+      onSelect(uniq(items))
+      this.onBlur()
+    } catch (err) {
+      console.error(err)
+    }
   }
 
   onBlur = () => {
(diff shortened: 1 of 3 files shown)


client/components/autocomplete2/index.jsx
```diff
@@ -66,11 +68,32 @@ export class Autocomplete extends Component {
     })
   }
 
-  onSelect = (item) => {
-    const { onSelect, items } = this.props
-    items.push(item._id)
-    onSelect(uniq(items))
-    this.onBlur()
+  getResItem = async (selected) => {
```
kanaabe: What is res in `getResItem` here? Is it result? If so, we should probably just use the term to avoid confusion. Same goes for `resObject`

eessex: Makes sense -- just updated naming to better reflect what things are used for

client/components/autocomplete2/test/index.test.js
```diff
@@ -47,12 +48,25 @@ describe('Autocomplete', () => {
...
     )
-    component.instance().onSelect(results[0])
-    expect(props.onSelect.mock.calls[0][0][0]).toBe(results[0]._id)
+    await component.instance().onSelect(searchResults[0])
+    expect(props.onSelect.mock.calls[0][0][0]).toBe(searchResults[0].id)
+  })
+
+  it('Returns a custom resObject on select if provided', async () => {
+    const formatSelected = (item) => {
+      return new Article(item)
```
kanaabe: I think I might be a little confused by this -- in order to test the code you wrote, I would think this function needed to return a Promise. If this is the way we intend to write it, perhaps we don't need async/await at all?

eessex: It doesn't wait for component.instance().onSelect (which is also async) to complete before hitting the expect if I don't include async in the test as well.  Async is a new concept to me so perhaps I am doing it wrong?

eessex: Oh i misunderstood your comment -- yes can make the sample prop async too

kanaabe: I think it depends on how we intend to use `formatSelected`. If `formatSelected` is meant to format objects by wrapping it in an `Article` and if it's not intended to make asynchronous requests, then we shouldn't need to use async/await in this PR. The async/await stuff comes in handy when we're dealing with Promises or when we want to avoid callbacks by turning things into Promises. 

For example, the `onSelect` could look like:

```
onSelect = (selected) => {
   const { items, onSelect, formatSelected } = this.props
   const item = formatSelected ? formatSelected(selected) : selected.id

   items.push(item)
   ...
}
```

eessex: we do need it -- was implementing for fetching artworks -- new test coming soon

client/components/autocomplete2/test/index.test.js
```diff
@@ -47,12 +48,53 @@ describe('Autocomplete', () => {
...
+  it('Returns a custom formatted selection on select if provided', async () => {
+    const formatSelected = async (item) => {
+      try {
+        const article = new Article(item)
+        article.fetch = jest.fn().mockReturnThis()
+        return await article.fetch()
+      } catch (error) {
+        return error
+      }
+    }
```
kanaabe: One little tip for mocking this is that Backbone fetches return promises, so there shouldn't be a need to wrap it in the try/catch and async/await on it. You can mock Backbone.sync instead. This might need tweaking, but something like this general pattern should work:

```
Backbone.sync = jest.fn()
const formatSelected = selected => new Article(selected).fetch()
```
I won't let this block you since what you have works, but just some food for thought.

kanaabe:
LGTM! Feel free to merge this yourself. Left a tip on mocking, but it's up to you if you want to implement it.

client/components/autocomplete2/test/index.test.js
eessex: Much better!

PR #1517, opened 2018-01-05
@kanaabe => Admin tags panels

eessex:
Fixes https://github.com/artsy/publishing/issues/285

- Refactors Admin Panel components for Tags and Verticals/Tags in JSX
- Adds AutocompleteInlineList -- wraps Autocomplete and shows a list of selected items


client/apps/edit/components/admin/components/verticals_tags.jsx
```diff
@@ -0,0 +1,112 @@
+import PropTypes from 'prop-types'
+import React, { Component } from 'react'
+import { Col, Row } from 'react-styled-flexboxgrid'
+import { data as sd } from 'sharify'
+import Verticals from '../../../../../collections/verticals.coffee'
+import { AutocompleteInlineList } from '/client/components/autocomplete2/inline_list'
```
kanaabe: Just a style thing, but it would be nice to keep import paths consistent by always using absolute!

client/apps/edit/components/admin/components/verticals_tags.jsx
```diff
@@ -0,0 +1,112 @@
...
+
+  constructor (props) {
+    super(props)
+    const { article } = props
+
+    this.state = {
+      vertical: article.get('vertical') || null,
+      verticals: []
+    }
+  }
```
kanaabe: I think this constructor is unnecessary, right? I'm not so picky about it, up to you.

client/apps/edit/components/admin/components/verticals_tags.jsx
eessex: The coffeescript absolute paths fail for me in tests, so had to use relative here.

client/apps/edit/components/admin/components/verticals_tags.jsx
```diff
@@ -0,0 +1,112 @@
...
+
+  componentWillMount = () => {
+    this.fetchVerticals()
+  }
+
+  fetchVerticals = () => {
+    new Verticals().fetch({
+      cache: true,
+      success: (verticals) => {
+        let sortedVerticals = verticals.sortBy('name')
```
kanaabe: Any reason this has to be a `let`? It doesn't look like we're reassigning it elsewhere. Generally I try to stick with `const` unless necessary.

client/apps/edit/components/admin/test/components/tags.test.js
```diff
@@ -0,0 +1,82 @@
...
+        <AdminTags {...props} />
+      )
+      const input = component.find('.tracking-tags input').first()
+      input.simulate('change', {target: {value: 'New York, Photography'}})
+
+      expect(props.onChange.mock.calls[0][0]).toBe('tracking_tags')
+      expect(props.onChange.mock.calls[0][1][0]).toBe('New York')
+    })
+  })
+})
```
kanaabe: Very nice set of tests as always

client/apps/edit/components/admin/components/verticals_tags.jsx
eessex: good catch

client/components/autocomplete2/inline_list.jsx
```diff
@@ -0,0 +1,51 @@
...
+                key={i}
+              >
+                {item}
+                <button onClick={() => this.onRemoveItem(i)} />
+              </div>
+            )
+          })}
+        </div>
+
+        <Autocomplete {...this.props} />
```
kanaabe: Nice! I like this extension

kanaabe:
I'm happy that modernizing code eliminates bugs! 

For posterity, could you explain what the problem was with React 16 and these tag components? It's worth it to understand root causes so that other can benefit from problems you've already encountered. For example, how did you know that modernizing would fix the bug?

client/apps/edit/components/admin/components/verticals_tags.jsx
kanaabe: Ooh interesting. @damassi does that seem familiar to you? I feel like there should be a way we can get the coffeescript paths to be absolute imports too.

eessex:
The bug was just that a list wasn't wrapped in a div -- relatively simple, but figured it was a good opportunity to start getting the admin panel up to date. 

Have been noticing that React 16 is more strict in its errors!  IE giving WSOD over just console errors.

kanaabe:
Great, thanks for the explanation and proactive refactoring! Oh wow, that's pretty harsh. At least we now have a BrowserStack account to actually test on 😅

client/apps/edit/components/admin/components/verticals_tags.jsx
damassi: Unfortunately we can't do absolute imports with CoffeeScript since we're piping everything through babel (which only transforms .js)

damassi:
@eessex I would investigate the WSOD stuff a bit more (which it seems like you're doing) as console errors are likely error-errors that just got missed.

client/apps/edit/components/admin/components/verticals_tags.jsx
damassi: See https://github.com/artsy/positron/blob/master/.babelrc#L19

eessex:
@damassi the errors are still in the console, but in 15 the same errors would not give WSOD

damassi:
This is precisely what [Error Boundaries](https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html) are for.

eessex:
Ooh interesting, this is new to me

damassi:
Oh yeah, that's the main reason I've been pushing for updates because it can catch all kinds of things in a much more elegant and safe way. A single error boundary at the top of the tree can prevent a crash.

damassi:
@eessex and @kanaabe - i would recommend reading over the [release notes](https://reactjs.org/blog/2017/09/26/react-v16.0.html) 😄

PR #1518, opened 2018-01-05
@kanaabe => Partner draft redirect bug

eessex:
Drafts for partner channels were still erroring, despite having an updated URL -- turns out this was because force is not passing a 'channel_id' when requesting partner drafts. 

hasChannelAccess has been updated not to return false when channel_id is missing.


kanaabe:
Good catch!

PR #1521, opened 2018-01-08
Revert "Fix minor dep updates"

kanaabe:
Reverts artsy/positron#1511


kanaabe:
Haha cool, this PR is failing on the test that was written after this PR fixed the issue with empty sections. Merging since all of it will be reverted in the 16 PR.

PR #1522, opened 2018-01-08
@damassi => Revert React 16

kanaabe:
Hope to see this come back asap! But for now as to not hold back deploys...

I couldn't use the revert UI button here because a file that was modified in that PR was deleted afterwards so this just reverts each commit manually.


damassi:
👍

PR #1523, opened 2018-01-08
@kanaabe => Mutations / Decaf drag source

eessex:
- Removes prop mutations from Drag-Drop and ImageCollection's Controls and ArtworkUrl components
- Decaffeinates DragSource component


kanaabe:
👍 LGTM outside of the test!

PR #1540, opened 2018-01-11
@damassi => Connect Header, SectionList

eessex:
- Adds new `changeSection` reducer and action for managing an article's `activeSection` index. 
- Renames old `changeSection` action (for navigating tabs) to `changeView` (to avoid confusing with article sections)

- Connect section list and use new `changeSection` action
- Connect EditHeader and use `changeView` action for tab navigation

### DIFF
--- a/client/actions/test/editActions.test.js
+++ b/client/actions/test/editActions.test.js
@@ -19,11 +19,11 @@ describe('editActions', () => {
     expect(changeSavedStatus.payload.isSaved).toBe(true)
   })
 
-  it('#changeSection sets the activeSection to arg', () => {
-    const changeSection = editActions.changeSection('display')
+  it('#changeView sets the activeSection to arg', () => {
+    const changeView = editActions.changeView('display')
 
-    expect(changeSection.type).toBe('CHANGE_SECTION')
-    expect(changeSection.payload.activeSection).toBe('display')
+    expect(changeView.type).toBe('CHANGE_VIEW')
+    expect(changeView.payload.activeView).toBe('display')
   })
 
   it('#deleteArticle destroys the article and sets isDeleting', () => {
--- a/client/actions/test/editActions.test.js
+++ b/client/actions/test/editActions.test.js
@@ -19,7 +19,14 @@ describe('editActions', () => {
     expect(changeSavedStatus.payload.isSaved).toBe(true)
   })
 
-  it('#changeView sets the activeSection to arg', () => {
+  it('#changeSection sets activeSection to arg', () => {
+    const changeSection = editActions.changeSection(6)
+
+    expect(changeSection.type).toBe('CHANGE_SECTION')
+    expect(changeSection.payload.activeSection).toBe(6)
+  })
+
+  it('#changeView sets the activeView to arg', () => {
     const changeView = editActions.changeView('display')
 
     expect(changeView.type).toBe('CHANGE_VIEW')
--- a/client/apps/edit/components/content/section_list/index.jsx
+++ b/client/apps/edit/components/content/section_list/index.jsx
@@ -1,18 +1,19 @@
 import PropTypes from 'prop-types'
 import React, { Component } from 'react'
+import { bindActionCreators } from 'redux'
+import { connect } from 'react-redux'
+import * as Actions from 'client/actions/editActions'
 import DragContainer from 'client/components/drag_drop/index.coffee'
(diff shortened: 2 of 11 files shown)


client/apps/edit/components/content/section_list/index.jsx
```diff
@@ -1,41 +1,52 @@
...
+  }
+
+  onDragEnd = (newSections) => {
     const { sections } = this.props
 
-    const editingIndex = sections.indexOf(section)
-    this.setState({ editingIndex })
+    sections.reset(newSections)
+    // force update until sections live in state
+    this.setState({lastUpdated: new Date()})
```
eessex: Backbone events do not trigger re-render in connected components, so need to do it manually for now.  Long-run goal is moving all the backbone actions into redux.

client/actions/test/editActions.test.js
```diff
@@ -19,11 +19,18 @@ describe('editActions', () => {
     expect(changeSavedStatus.payload.isSaved).toBe(true)
   })
 
-  it('#changeSection sets the activeSection to arg', () => {
-    const changeSection = editActions.changeSection('display')
+  it('#changeSection sets activeSection to arg', () => {
+    const changeSection = editActions.changeSection(6)
```
damassi: Wondering -- is there any way to avoid numeric indexes for stuff like this? They don't convey much information to the outside world and tend to be brittle.

client/apps/edit/components/content/section_list/index.jsx
damassi: Maybe add `TODO` to the comment so that it jumps out in editor?

client/actions/test/editActions.test.js
eessex: Planning to move the sections collection into the Redux state eventually, where potentially this could use the section object itself over an index. However that requires changing many components to support, so this is an incremental solution that sticks to our current system of finding a section by index.

client/apps/edit/components/content/section_list/index.jsx
```diff
@@ -86,11 +97,24 @@ export class SectionList extends Component {
...
   }
 }
+
+const mapStateToProps = (state) => ({
+  ...state
+})
+
+const mapDispatchToProps = (dispatch) => ({
+  actions: bindActionCreators(Actions, dispatch)
+})
```
damassi: Ah, you don't necessarily need `bindActionCreators` here as you can assign specific actions and they're bound for you. For example, there are to possible forms:

```javascript
import { changeSection } from 'my/actions'

const mapDispatchToProps = {
  changeSectionAction: changeSection
}
```
(Its nice to add "Action" to the end of the key so that it jumps out in props code, though not required.)

Can also do (similar to what you have): 

```javascript
const mapDispatchToProps = (dispatch) => {
  const dontDispatchRightAway = debounce(() => dispatch(myAction), 1000)

  return {
    someOtherAction: dontDispatchRightAway
  }
}
```
The second form is good if you need to perform intermediate actions before things are dispatched, or mess with the action pipeline in some way.

client/apps/edit/components/header/index.jsx
```diff
@@ -177,3 +176,16 @@ export class EditHeader extends Component {
     )
   }
 }
+
+const mapStateToProps = (state) => ({
+  ...state
+})
+
+const mapDispatchToProps = (dispatch) => ({
+  actions: bindActionCreators(Actions, dispatch)
```
damassi: Same applies here

client/apps/edit/components/content/section_list/index.jsx
eessex: Excellent, thanks!

client/actions/test/editActions.test.js
damassi: Sounds good 👍

client/apps/edit/components/content/section_list/index.jsx
```diff
@@ -86,11 +97,24 @@ export class SectionList extends Component {
...
           isDraggable={false}
         />
       </div>
     )
   }
 }
+
+const mapStateToProps = (state) => ({
+  ...state
+})
```
damassi: Its probably best to pluck out exactly what pieces of state you need from the code for explicitness, but either way is really fine. That said, this can be simplified to:

```javascript
const mapStateToProps = state => state
````

PR #1554, opened 2018-01-19
[WIP] Move Channel & User to Redux

eessex:
Eliminates the need to pass `channel` and `user` down the tree in the `edit/content` app. 

- Channel exists as a backbone model in `store.app`
- User was only being used for an admin check, so `isAdmin` boolean was added to `store.app`
- Cleans up connect declarations to only use needed props


client/apps/edit/components/content/index.jsx
```diff
@@ -1,13 +1,12 @@
 import PropTypes from 'prop-types'
 import React, { Component } from 'react'
-import { EditArticle } from './article_layouts/article.jsx'
+import EditArticle from './article_layouts/article.jsx'
```
l2succes: how come we include the file extension in those imports?

eessex: Good catch! removing

client/apps/edit/client.js
```diff
@@ -14,7 +13,7 @@ import { data as sd } from 'sharify'
 
 export function init () {
   const article = new Article(sd.ARTICLE)
-  const channel = new Channel(sd.CURRENT_CHANNEL)
+  const channel = initialState.app.channel
```
l2succes: concerning the `channel` object in the redux store, it's currently still a backbone model and I believe it should be a plain javascript object. 

https://github.com/eessex/positron/blob/77f35a989f35f94e9e18cbfda3da41a3b528bdb8/client/reducers/appReducer.js#L5

The state should be serializable and hydrated into the client from a local cache or server. In order to do so, all properties on the state object should be plain serializable objects, arrays, and primitives. 

You can read more about it here: https://redux.js.org/docs/faq/OrganizingState.html#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state

l2succes: To be clear, it think it could also be fine to leave backbone models in the store for now especially since the `Channel` model isn't just a wrapper around JSON, there are custom methods on the class. But if we want to have support for things like time-traveling debugging (which is pretty neat), and replaying state eventually. we should think about keeping only plain objects in the store.

If having the following is not feasible for now, it's ok to keep the backbone model as well until we figure out a way to move away from it.
```javascript
import { data as sd } from 'sharify'
export const initialState = {
  channel: sd.CURRENT_CHANNEL,
  isAdmin: sd.USER.type === 'Admin',
  user: sd.USER
}
```

PR #1558, opened 2018-01-23
@kanaabe => Decaf TextInputUrl

eessex:
Small step towards moving Draft components into es6-- decaffeinates the url input used in text sections.


client/components/rich_text/components/input_url.jsx
```diff
@@ -0,0 +1,77 @@
...
+    const { removeLink, selectionTarget } = this.props
+    const { left, top } = selectionTarget
+
+    return (
+      <div
+        className='TextInputUrl'
+        style={{
+          top: top || 0,
+          left: left || 0
+        }}
```
kanaabe: Possible to choose styled-components or CSS on this? I think we've discussed before but it makes it a little hard to track down when we're using 3 different ways to style.

client/components/rich_text/test/components/input_url.test.js
```diff
@@ -0,0 +1,110 @@
...
+  })
+
+  it('Can remove a link', () => {
+    props.urlValue = 'http://artsy.net'
+    const component = getWrapper(props)
+    component.find(RemoveButton).simulate('click')
+
+    expect(props.removeLink.mock.calls.length).toBe(1)
+  })
+})
```
kanaabe: Great tests!

client/components/rich_text/components/input_url.jsx
eessex: Good idea - will add this.

Issue #1403, opened 2017-11-06
Invalid characters break client-side JS


kanaabe:
This article: https://www.artsy.net/article/artsy-editorial-the-most-influential-curators-of-2016

Contained an `LS` character (which I think is line separator in unicode?). This subsequently broke client-side JS and stopped rendering of images, etc. 

Don't have much further info on this, but logging for posterity.

kanaabe:
Happened also with: https://www.artsy.net/article/artsy-editorial-the-top-emerging-artists-of-2016

PR #1560, opened 2018-01-24
@kanaabe => Strip unicode spaces

eessex:
Json stringify seems to have issues with unicode spaces 2028 and 2029.
 (https://stackoverflow.com/questions/16686687/json-stringify-and-u2028-u2029-check)

This adds a check specifically for those characters and removes them on paste.

Related to https://github.com/artsy/positron/issues/1403

### DIFF
--- a/client/apps/edit/components/content/sections/text/index.coffee
+++ b/client/apps/edit/components/content/sections/text/index.coffee
@@ -156,6 +156,7 @@ module.exports = React.createClass
     { editorState } = @state
     unless html
       html = '<div>' + text + '</div>'
+    html = standardizeSpacing(html)
     html = stripGoogleStyles(html)
     blocksFromHTML = convertFromRichHtml html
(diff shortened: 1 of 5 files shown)


kanaabe:
According to the link it doesn't seem like the behavior should be to strip it but instead escape it properly? I'm not too familiar with this, but if the character is presented somewhere like Google Docs correctly, it could cause unexpected changes to the text if we strip it completely.  

Also it looks like the tests are checking if the function is called but doesn't actually test the code change.

eessex:
Rather than escaping linebreaks (which could potentially be outside our spacing rules), added a new function that replaces them with an empty paragraph.  This was the spacing should be preserved, but also stay within our ruleset.  This is called on mount as well as on paste, so hopefully we should be good.

kanaabe:
I'm sure you've tested that this works locally, but I was thinking there's probably a few articles still out there with this character and it'd be a good opportunity to search for them in Robomongo to clean up entirely. Happy to do this part with you sometime!

PR #1562, opened 2018-01-24
@l2succes => Section Actions

eessex:
Sets up actions for editing article sections:

- `setSection` (was `activeSection`) sets currently editing section to store.section from index arg
- `newSection` inserts a section with keys of type at specified article.sections index, and sets it as the editing section
- `removeSection` removes a section from article.sections based on index
- `onChangeSection` updates section and article.sections


l2succes:
👍

eessex:
Ok to merge?

PR #1563, opened 2018-01-24
@damassi => Error boundary

eessex:
Adds a basic error boundary component and wraps:

- SectionContainer

Via edit_container component:
- EditHeader
- EditContent/EditAdmin/EditDisplay


client/components/error/test/error_boundary.test.js
```diff
@@ -0,0 +1,28 @@
...
+        <div>A child component</div>
+      </ErrorBoundary>
+    )
+  }
+
+  it('Displays children', () => {
+    const component = getWrapper()
+    expect(component.text()).toMatch('A child component')
+  })
+
```
eessex: Tried to test the catch mechanism here but had issues simulating the error.  Throwing an error in the children would cause the test to fail, rather than completing the test.  Any ideas on this?

damassi: @eessex, have you tried

```javascript
expect(() => {
   doSomethingThatThrows()
}).toThrow()
```

(This is what I'm thinking: https://facebook.github.io/jest/docs/en/expect.html#tothrowerror)

eessex: The problem I'm looking at is a little different, I'm trying to actually throw an error and confirm that it is caught. 

Something like this -- 
`
it('Can catch errors', () => {
  props.onClick = () => {
     let err = 'An Error'
     throw err
   }
    console.error = jest.fn()
    const component = getWrapper(props)
    const child = component.find('.child').first()

    child.simulate('click')
    expect(console.error.mock.calls[0][0]).toBe('An Error')
  })
`

eessex: The error I throw causes the test to fail, despite the assertions being true: 
screen shot 2018-01-24 at 5 20 06 pm

damassi: Interesting, let me pull down this branch and give it a look as I don't have any ideas off the top of my head. This is all very new and I think a lot of patterns are still being laid down in the community.

eessex: Do you think it is necessary to test this assertion since componentDIdCatch is a built in method, and we're not doing anything with it yet?

damassi: So what I would say about this is that the `<ErrorBoundary />` component has been sufficiently tested in the React codebase, and that one shouldn't worry about testing such low-level behavior. That said, I agree this is a bit annoying but don't think its something to worry about!

eessex: @damassi in that case do you think this is ok to merge?

damassi: I do 👍

damassi:
LGTM 👍

PR #1564, opened 2018-01-24
@essex => Improved article editing

l2succes:
This PR implements a real-time mechanism to only allow one user to edit an article at a time.

Tasks completed:
- We should be able to tell if a user is actively engaging with the page ([IAE-6](https://artsyproduct.atlassian.net/browse/IAE-6))
- There should be an indication that an article is locked on the homepage ([IAE-8](https://artsyproduct.atlassian.net/browse/IAE-8))
- Prevent users from clicking link to a locked article ([IAE-9](https://artsyproduct.atlassian.net/browse/IAE-9))
- Redirect user to writer.artsy.net/articles after 10 seconds ([IAE-3](https://artsyproduct.atlassian.net/browse/IAE-3))
- Error message should display on a locked article (design here: zpl.io/VD7PMn3) ([IAE-2](https://artsyproduct.atlassian.net/browse/IAE-2))

Previews:

Locked row
screen shot 2018-01-26 at 5 30 18 pm

Lock modal
screen shot 2018-01-29 at 6 22 13 pm


client/actions/editActions.js
```diff
@@ -30,6 +35,14 @@ export const changeView = (activeView) => ({
...
 })
 
+export const viewArticles = emitAction(() => ({
+  type: actions.VIEW_ARTICLES,
+  key: messageTypes.articlesRequested,
+  payload: {
+    timestamp: new Date().toISOString()
+  }
+}))
+
```
eessex: If this is for viewing the article list it might make sense to create a new reducer for articles.

client/apps/edit/client.js
```diff
@@ -23,6 +24,7 @@ export function init () {
   new EditLayout({ el: $('#layout-content'), article, channel })
 
   const store = createReduxStore(reducers, initialState)
+  initWebsocket(store)
```
eessex: :raised_hands:

client/actions/editActions.js
l2succes: absolutely, I'm still working on getting things working, but I'll start reorganizing/moveing some code units afterwards

eessex: Awesome, looking forward to seeing where this goes

l2succes: 👍 I'll let you know when it's in a good spot for an initial review 🙂

client/apps/articles_list/components/listView.jsx
```diff
@@ -0,0 +1,147 @@
...
+        </h1>
+        {this.showArticlesList()}
+      </div>
+    )
+  }
+}
+
+const mapStateToProps = (state) => ({
+  ...state
+})
```
damassi: Can do `mapStateToProps = state => state`

client/apps/articles_list/components/listView.jsx
```diff
@@ -0,0 +1,147 @@
...
+
+const mapDispatchToProps = (dispatch) => ({
+  actions: bindActionCreators(Actions, dispatch)
+})
```
damassi: Can alternatively do 
```javascript
import { fetchThing } from './actions'

const mapStateToProps = {
  fetchThingAction: fetchThing
}
```

client/apps/edit/components/edit_container.jsx
```diff
@@ -33,10 +44,23 @@ class EditContainer extends Component {
...
+
+  componentWillUnmount () {
+    //TODO: Send stopEditingArticle action
+  }
+
   onChange = (key, value) => {
     const { article } = this.props
 
     article.set(key, value)
+    this.updateArticleAction(article.toJSON())
```
damassi: should this be `this.props. updateArticleAction`?

client/components/article_list/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+      return (
+        <div style={style} className='article-list__result paginated-list-item' key={article.id}>
+          {checkable
+            ? <div
+                className='article-list__checkcircle'
+                ref={article.id}
+                dangerouslySetInnerHTML={{__html: $(icons()).filter('.check-circle').html()}}
+                onClick={() => this.props.selected(article)}
+              />
+            : null}
```
damassi: This is a matter of preference, but can simplify to 
```
{checkable &&
  <div>
    ...
  </div>}
```

client/apps/websocket/client.js
```diff
@@ -0,0 +1,34 @@
...
+const sd = require('sharify').data
+const socket = io(sd.APP_URL)
+
+const init = (store) => {
+  // add listeners to socket messages so we can re-dispatch them as actions
+  Object.keys(messageTypes)
+    .forEach(key =>
+      socket.on(key, (data) => {
+        const { type, payload } = data
+        console.log(type, payload)
```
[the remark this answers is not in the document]
l2succes: Yeah definitely, I still need to clean up this module a bit after I get all the events hooked up

client/reducers/articlesReducer.js
```diff
@@ -0,0 +1,41 @@
...
+  articlesInSession: sd.ARTICLES_IN_SESSION || {}
+}
+
+export function articlesReducer (state = initialState, action) {
+  switch (action.type) {
+    case actions.EDITED_ARTICLES_RECEIVED: {
+      return {
+        ...state,
+        articlesInSession: action.payload
+      }
```
damassi: I believe there's a lib called [updeep](https://github.com/substantial/updeep) being used elsewhere; what it does is allow for easy / fast / pseudo-immutable updates to nested data:

```javascript
import u from 'updeep'
...
case actions.EDITED_ARTICLES_RECEIVED: {
  return u({
    articlesInSession: action.payload,
    ...
    foo: {
      baz: [
        { ... }
      ]
    }
  }, state)
}

File: client/apps/articles_list/components/listView.jsx [hunk 9]
```diff
@@ -0,0 +1,147 @@
...
+export class ArticlesListView extends Component {
+  static propTypes = {
+    actions: PropTypes.object,
+    articles: PropTypes.array,
+    published: PropTypes.bool,
+    channel: PropTypes.object
+  }
+
+  state = {
+    articles: this.props.articles || [],
```
damassi: I like how state is handled here; rather than dispatching local concerns back to the store just manage it with local state 👍 Makes things clearer and more precise

client/apps/articles_list/components/listView.jsx
damassi: Ideally though, rather than returning the entire state it's best to just pluck what you need, so that the API boundaries of the component are clearly defined:

```javascript
const mapStateToProps = (state) => ({
  foo: state.foo,
  baz: stare.bar.baz
})
```

client/apps/articles_list/components/listView.jsx
```diff
@@ -0,0 +1,147 @@
...
+import $ from 'jquery'
+import { debounce } from 'lodash'
+import FilterSearch from 'client/components/filter_search/index.coffee'
+import request from 'superagent'
+
+require('jquery-on-infinite-scroll')
+const query = require('../query.coffee')
+const icons = () => require('../icons.jade')(...arguments)
+
+export class ArticlesListView extends Component {
```
damassi: I would make sure to map component name to file-name 1:1. At some point Positron snake_casing should go away in favor of more idiomatic modern JS naming (and for consistency across other Artsy apps); e.g.,
 
```javascript
import { ArticlesListView } from 'components/ArticlesListView'` 
```

(or something similar 😄)

But in the meantime file should still match export (`articles_list_view.js`) so that its clear what is being used where. 

cc @eessex

eessex: 👍 yeah ArticlesList is probably a better choice

l2succes: That's precisely what I renamed the class and file to, just haven't pushed it yet 😊

client/apps/edit/components/message/test/index.test.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+import { MessageModal } from '../index'
+import { ModalTypes } from '../modalTypes'
+import { shallow } from 'enzyme'
+
+describe('MessageModal', () => {
+  it('renders lock modal', () => {
+    const type = 'locked'
+    const data = ModalTypes[type]
+    const component = shallow(<MessageModal {...getProps(type)} />)
```
damassi: This is perhaps arguable, but `mount` will always provide more real-world assurances than `shallow` while also being less confusing down the line (when some future dev wants to assert the existence of some random child in the tree but length always unexpectedly returns 0, or events don't bind / trigger, etc). That said, `dive` is really nice / useful.

client/apps/articles_list/client.js
```diff
@@ -0,0 +1,25 @@
...
+const store = createReduxStore(reducers, initialState)
+initWebsocket(store)
+
+export const init = () => (
+  render(
+    <Provider store={store}>
+      <ArticlesList
+        articles={sd.ARTICLES}
+        published={sd.HAS_PUBLISHED}
+        channel={new Channel(sd.CURRENT_CHANNEL)}
```
eessex: The channel is also now in the store here - https://github.com/artsy/positron/blob/master/client/reducers/appReducer.js#L4

l2succes: Oh I must've missed it during the merge, cool I'll remove that

client/reducers/articlesReducer.js
```diff
@@ -11,29 +12,26 @@ export const initialState = {
...
+      const articlesInSession = cloneDeep(state.articlesInSession)
       const session = action.payload
       articlesInSession[session.article] = session
 
-      return {
-        ...state,
+      return u({
         articlesInSession
-      }
+      }, state)
```
damassi: I'm not familiar with the particulars of this data structure (e.g., possible backbone data, etc), but the use of `cloneDeep` in this file should be unnecessary with updeep / redux as each cycle should return [a pseudo-immutable (frozen) copy](https://github.com/substantial/updeep#about). Thinking something like

```javascript
u({
  articlesInSession: {
    [session.article]: session
  }
}, state)
```
would work similarly well.

Updeep offers a number of functional ways to update data, but simple object updates are typically sufficient. However, [`u.constant`](https://github.com/substantial/updeep#uconstantobject) is really handy if you want to replace something outright.

client/apps/articles_list/components/articles_list.jsx
```diff
@@ -0,0 +1,147 @@
...
+          <FilterSearch
+            url={sd.API_URL + `/articles?published=${this.state.published}&channel_id=${sd.CURRENT_CHANNEL.id}&q=%QUERY`}
+            placeholder='Search Articles...'
+            collection={this.state.articles}
+            searchResults={this.setResults}
+            selected={null}
+            contentType='article'
+            checkable={false}
+            isArtsyChannel={this.props.channel.isArtsyChannel()}
+          />
```
eessex: Since the store's channel is a plain object, we don't have the isArtsyChannel method here -- you can use `channel.type !== 'partner'` for the same result.

client/apps/articles_list/components/articles_list.jsx
```diff
@@ -0,0 +1,147 @@
...
+  }
+}
+
+const mapStateToProps = (state) => ({
+  channel: state.app.channel
+})
+
+const mapDispatchToProps = (dispatch) => ({
+  actions: bindActionCreators(Actions, dispatch)
+})
```
eessex: Looks like you've already done this in other files, but mapDispatchToProps can be rewritten to only include the action you are using.

eessex:
It looks like there is an empty file at `client/apps/articles_list/fetchArticles.js`

client/apps/edit/components/tests/edit_container.test.js
```diff
@@ -51,7 +52,16 @@ describe('EditContainer', () => {
       error: {},
       isSaved: false,
       saveArticleAction: jest.fn(),
-      toggleSpinnerAction: jest.fn()
+      toggleSpinnerAction: jest.fn(),
+      startEditingArticleAction: jest.fn(),
+      stopEditingArticleAction: jest.fn(),
+      updateArticleAction: jest.fn(),
+      user: {},
+      currentSession: {
```
eessex: Just want to make sure I'm understanding correctly -- currentSession is the session logged from websockets?  This means it is not necessarily the session of the user on the line above?

client/components/article_list/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+      const attrs = this.getDisplayAttrs(article)
+      const session = activeSessions[article.id]
+      const isCurrentlyBeingEdited = session
+      const isCurrentUserEditing = user && session && user.id === session.user.id
+      const style = isCurrentlyBeingEdited ? {color: colors.grayMedium} : null
+      const shouldLockEditing = isCurrentlyBeingEdited && !isCurrentUserEditing
+      const lockedClass = shouldLockEditing ? 'locked' : ''
+
+      return (
+        <div style={style} className='article-list__result paginated-list-item' key={article.id}>
```
eessex: We've been moving towards using StyledComponents rather than declaring inline styles.

client/components/article_list/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+          {checkable
+            ? <div
+                className='article-list__checkcircle'
+                ref={article.id}
+                dangerouslySetInnerHTML={{__html: $(icons()).filter('.check-circle').html()}}
+                onClick={() => this.props.selected(article)}
```
eessex: I believe there is an icon in Reaction already for the check circle that can be used here (is in top level icons, not publishing)

client/test/models/article.test.coffee
```diff
@@ -89,11 +89,6 @@ describe "Article", ->
 
   describe '#getPublishDate', ->
 
-    it 'returns the current date if unpublished and no scheduled_publish_at', ->
-      @article.set published: false
-      now = moment().format('YYYY-MM-DD')
-      @article.getPublishDate().should.containEql now
-
```
eessex: Was this removed intentionally?

client/components/article_list/index.jsx
damassi: @eessex - thats great 👍 One thing I recently solved is better classname support during development -- i found that `babel-plugin-styled-components` displayName output isn't compatable with `babel-plugin-rewire`, but its easy enough to fix by making sure that babel-plugin-rewire is only loaded when `NODE_ENV=test`: https://github.com/artsy/force/blob/master/.babelrc. Checked positron's .babelrc and it seems like this still needs to be setup.

client/apps/articles_list/components/listView.jsx
l2succes: Yes, definitely. We weren't using the redux state in this component which is why is why I didn't explicitly list out any props.

client/components/article_list/index.jsx
l2succes: Yeah I prefer that as well, that component was converted from coffeescript to ES6 so I made sure it worked.

client/reducers/articlesReducer.js
l2succes: Yes, I've seen in it in `editReducer.js` I went ahead and used it in this reducer module as well

client/apps/edit/components/message/test/index.test.js
l2succes: Right, I didn't think this test warranted the use of `mount` but that may change if we go deeper down the render tree

client/apps/edit/components/tests/edit_container.test.js
l2succes: `currentSession` is the session of the user who is currently editing the article, so it's not necessarily the user authenticated in the browser session.

client/apps/articles_list/components/listView.jsx
damassi: If not using can probably remove for now

client/components/article_list/index.jsx
l2succes: Right, I didn't use styled-components here as this component was converted from coffeescript, I just focused on making it work in ES6 due to time. But I can definitely go back after we this merged in and convert the styling as well.

@damassi I wasn't aware of `babel-plugin-styled-components`, very useful for debugging purposes.

eessex: Ok not a blocker for this PR but maybe add a TODO so we don't forget

client/test/models/article.test.coffee
l2succes: Yeah cause it was failing and I couldn't get it to pass initially but it works now so I added it back

client/apps/articles_list/client/client.coffee
```diff
@@ -1,102 +0,0 @@
...
-onInfiniteScroll = require 'jquery-on-infinite-scroll'
-{ div, nav, a, h1 } = React.DOM
-Article = require '../../../models/article.coffee'
-Channel = require '../../../models/channel.coffee'
-FilterSearch = React.createFactory require '../../../components/filter_search/index.coffee'
-query = require '../query.coffee'
-sd = require('sharify').data
-request = require 'superagent'
-icons = -> require('../icons.jade') arguments...
-
```
eessex: This test file needs to be replaced with a jest test file.

client/apps/articles_list/test/client/client.test.coffee
```diff
@@ -1,60 +0,0 @@
...
-ReactTestUtils = require 'react-dom/test-utils'
-ReactDOMServer = require 'react-dom/server'
-fixtures = require '../../../../../test/helpers/fixtures.coffee'
-Channel = require '../../../../../client/models/channel.coffee'
-r =
-  find: ReactTestUtils.findRenderedDOMComponentWithClass
-  simulate: ReactTestUtils.Simulate
-
-describe 'ArticlesListView', ->
-
```
eessex: This test file needs to be replaced with a jest test file.

client/apps/edit/components/edit_container.jsx
```diff
@@ -80,6 +120,33 @@ export class EditContainer extends Component {
...
+    })
+  }
+
+  sendStopEditing = () => {
+    if (this.state.sentStopEditingEvent) return
+
+    const { article, stopEditingArticleAction, user } = this.props
+    stopEditingArticleAction({
+      article: article.id,
+      user
```
eessex: The EditContainer tests should be updated to cover the new logic in this file.

client/apps/websocket/index.js
```diff
@@ -0,0 +1,92 @@
...
+  const event = articlesRequested
+  socket.emit(event, {
+    type: 'EDITED_ARTICLES_RECEIVED',
+    payload: articlesInSession
+  })
+}
+
+export const onUserStartedEditing = ({io, socket}, data) => {
+  console.log('[socket] onUserStartedEditing, userId: ', data.user.id, ', article: ', data.article)
+  const { timestamp, user, article } = data
```
eessex: Stray console log, or do you want this here?

client/components/article_list/test/index.test.coffee
```diff
@@ -1,78 +0,0 @@
...
-
-describe 'ArticleList', ->
-
-  beforeEach (done) ->
-    benv.setup =>
-      benv.expose $: benv.require 'jquery'
-      window.jQuery = $
-      @ArticleList = benv.requireWithJadeify(
-        resolve(__dirname, '../index')
-        ['icons']
```
eessex: This test file needs to be replaced with jest tests.

client/components/filter_search/test/index.test.coffee
```diff
@@ -1,133 +0,0 @@
...
-r =
-  find: ReactTestUtils.findRenderedDOMComponentWithClass
-  simulate: ReactTestUtils.Simulate
-
-
-describe 'FilterSearch Article', ->
-
-  beforeEach (done) ->
-    benv.setup =>
-      benv.expose
```
eessex: This test file needs to be fixed or replaced with jest tests.

client/components/layout/public/icons/lock.svg
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="8px" height="11px" viewBox="0 0 8 11" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 48.2 (47327) - http://www.bohemiancoding.com/sketch -->
+    <title>Group 9</title>
+    <desc>Created with Sketch.</desc>
```
eessex: This file can be removed if you are using the reaction icons.

eessex:
@l2succes this looks good but there are a number of test files that should be converted to jest or fixed rather than deleted.  Ready to go once those are done.

l2succes:
@eessex Yes I agree, I was planning on re-adding them in a subsequent PR to get this in first but might as well

client/apps/websocket/index.js
l2succes: the `[socket]` logs are intented to be there for server debugging purposes

l2succes:
@eessex Finished addressing your feedback concerning tests

PR #1567, opened 2018-01-30
@l2succes => Move EditLayout save functions to redux

eessex:
Moves all actions related to article saving/setup out of EditLayout backbone view.  

Was not able to deprecate the view yet, as auto-linking and Yoast are still using backbone.

### DIFF
--- a/client/actions/editActions.js
+++ b/client/actions/editActions.js
@@ -7,10 +7,13 @@ export const actions = keyMirror(
   'ERROR',
   'NEW_SECTION',
   'ON_CHANGE_SECTION',
+  'ON_FIRST_SAVE',
   'PUBLISH_ARTICLE',
+  'REDIRECT_TO_LIST',
   'REMOVE_SECTION',
   'SAVE_ARTICLE',
-  'SET_SECTION'
+  'SET_SECTION',
+  'TOGGLE_SPINNER'
 )
 
 export const changeSavedStatus = (article, isSaved) => ({
@@ -75,16 +78,35 @@ export const onChangeSection = (key, value) => {
   }
 }
 
+export const onFirstSave = (id) => {
+  location.assign(`/articles/${id}/edit`)
+
+  return {
+    type: actions.ON_FIRST_SAVE
+  }
+}
+
 export const publishArticle = (article, published) => {
+  if (published) {
+    setSeoKeyword(article)
+  }
   article.set('published', published)
   article.save()
-  article.trigger('finished')
+  redirectToList(published)
 
   return {
     type: actions.PUBLISH_ARTICLE,
     payload: {
       isPublishing: true
     }
+  }
+}
+
+export const redirectToList = (published) => {
+  location.assign(`/articles?published=${published}`)
+
+  return {
+    type: actions.REDIRECT_TO_LIST
   }
 }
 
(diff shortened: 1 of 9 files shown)


client/actions/editActions.js
```diff
@@ -106,6 +132,18 @@ export const saveArticle = (article) => {
   }
 }
 
+export const toggleSpinner = (isVisible) => {
+  if (isVisible) {
+    $('#edit-sections-spinner').show()
+  } else {
```
eessex: This is called on 'componentDidMount', so regrettably needs to remain jQuery (until we have SSR)

client/apps/edit/components/edit_container.jsx
```diff
@@ -1,39 +1,70 @@
...
 
+  componentDidMount = () => {
+    this.props.toggleSpinnerAction(false)
+  }
+
+  setupBeforeUnload = () => {
+    const { article } = this.props
+
+    if (article.get('published')) {
+      article.on(
```
l2succes: This looks like it will add a new event listener on every `change ` event for the current article model, wouldn't `.once` be preferable here?

client/apps/edit/components/edit_container.jsx
```diff
@@ -1,39 +1,70 @@
...
+    toggleSpinnerAction: PropTypes.func
   }
 
   constructor (props) {
     super(props)
 
-    this.state = {
-      lastUpdated: null
-    }
+    this.beforeUnload = this.beforeUnload.bind(this)
```
l2succes: Unless I'm missing something, you shouldn't need to bind `this` here as `beforeUnload` is defined as bound function using ES6's property initializer (hhttps://babeljs.io/docs/plugins/transform-class-properties/l)

l2succes:
@eessex I left a few minor comments but I'll go ahead merge since I've also made some changes in `EditContainer` in my branch including adding a `beforeunload` event handler so I'd like to resolve the merge conflicts and test with that.

client/apps/edit/components/edit_container.jsx
eessex: Ah yes you are right -- once is definitely preferable here.

client/apps/edit/components/edit_container.jsx
eessex: Despite using the es6 binding it was not getting the correct 'this' in this instance.

client/apps/edit/components/edit_container.jsx
l2succes: Actually, why not use the `maybeSaveArticle` or `onChange` event already in the class to check if we should add the listener instead of introducing a new backbone event listener?

https://github.com/artsy/positron/pull/1567/files#diff-b18066acb67f8f3287b04ae47476bc57R83

client/apps/edit/components/edit_container.jsx
```diff
@@ -1,39 +1,70 @@
...
+        () => window.addEventListener('beforeunload', this.beforeUnload)
+      )
+    }
+  }
+
+  beforeUnload = (e) => {
+    const { isSaved } = this.props
+    // Custom messages are deprecated in most browsers
+    // and will show default browser message instead
+    if ($.active > 0) {
```
l2succes: can't we use `state.isSaving` instead of jQuery here?

PR #1569, opened 2018-02-06
@kanaabe => Refactor text component

eessex:
- Reworks text component to use plain objects rather than backbone get/set
- Adds placeholders for new redux actions
- Converts to JSX
- Fixes link input add/remove bugs https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?projectKey=ERB&rapidView=17&selectedIssue=ERB-9


client/apps/edit/components/content/sections/text/test/index.test.js
```diff
@@ -0,0 +1,755 @@
...
+      const component = getWrapper(props)
+      component.instance().onPaste(badHtml)
+
+      expect(component.state().html).toBe('<p>Hello.</p>')
+    })
+  })
+})
+
+// xdescribe '#handleChangeSection', ->
+
```
eessex: somehow missed these -- will add later today

client/apps/edit/components/content/section_container/index.jsx
```diff
@@ -80,13 +86,15 @@ export class SectionContainer extends Component {
 
       case 'text': {
         const { end, start } = this.getContentStartEnd()
+        const textProps = extend(clone(this.props), {
```
kanaabe: Another option without clone: extend({}, this.props)

client/apps/edit/components/content/sections/text/index.jsx
```diff
@@ -0,0 +1,557 @@
...
+  }
+
+  onChange = (editorState) => {
+    const { article, onChange, section } = this.props
+    const html = convertToRichHtml(editorState, article.layout)
+
+    if (section.body !== html) {
+      onChange('body', html)
+    }
+    this.setState({ editorState, html })
```
kanaabe: If we move this up to the if block, I think we could prevent an additional re-render?

eessex: EditorState keeps track of changes in focus and text selection, so we still need to setState -- the if block is to prevent unnecessary saves when those events occur.

client/apps/edit/components/content/sections/text/index.jsx
```diff
@@ -0,0 +1,557 @@
...
+            pluginType={plugin}
+            urlValue={url}
+          />
+        }
+
+      </Text>
+    </div>
+    )
+  }
+}
```
kanaabe: My general feedback on this component is that this is a _lot_ of logic. Generally when a file is getting to be over 500 lines, I would consider breaking it down for easier reading, testing, and consumption. I see a few opportunities to extract some of this into sub components. Thinking something along the lines of `<TextNavWrapper/>` and `<EditorWrapper/>`. Thoughts?

eessex: It grew by ~200 lines when I converted it to JSX, but actually now has less logic than it did before-- moved a number of other functions into the utils files, which could probably go further.  I don't think this can break into any more subcomponents ATM.

client/apps/edit/components/content/sections/text/test/index.test.js
kanaabe: 👍  I trust you have a good sense of the coverage in this file, but on first glance it seems like it's missing some tests like `handleKeyCommand` and the stuff that happens in `componentDidUpdate`? All this to say, I think this would be much simpler to evaluate if these were broken out somehow!

eessex: HandleKeyCommand is here - and further in the keybindings utils tests
https://github.com/artsy/positron/pull/1569/files#diff-09414ab6a2a89e0a22eeff51a9f2077cR342

eessex: The tests are the same as the coffee version, with new ones added for things that were moved to utils. Could be more granular is true. 
Can take a look at using separate test files for the different describe blocks.

client/apps/edit/components/content/sections/text/index.jsx
kanaabe: Could you explain why it couldn't be broken into more subcomponents?

Happy that it has less logic than it did before, but looking holistically, it's still a lot for one file. Most of those methods get passed into components through props and could probably work with an subcomponent that was responsible for managing changes between the two. I'm not saying that it has to be done now or even at all, but instead giving some feedback on the readability of this file from a reviewer's perspective.

eessex: It can't be broken into components because there are no dom elements left that aren't already their own component except the wrapper, but that needs access to the editor ref to focus the editor correctly -- they all have to live side by side except reaction's Text component.  Working on moving more things into Utils, which puts it closer to 400 lines.

eessex: Also we can potentially connect the subcomponents once the new redux setup is up and running, which would remove some of the need to pass props.

kanaabe: Not sure I understand. Wrapper components don't need to have additional dom elements -- we can use them to extract methods and pass the original component into the render method:

```
export class TextInputUrlWrapper extends Component {

   componentWillUpdate() {
      // maybe do some operations to keep state/parents in sync
   }

   removeLink: () => {
      const newEditorState = // get new editorState after removing link
      this.props.onEditorStateChange(newEditorState) //tell parent about change
   }

   render () {
       const {
         selectionTarget,
         plugin,
         url
       } = this.props

       return (
         <TextInputUrl
           removeLink={this.removeLink}
           confirmLink={this.confirmLink}
           selectionTarget={selectionTarget}
           pluginType={plugin}
           urlValue={url}
         />
       )
   }
}
```

I might still be missing something but keeping data like `editorState` and references to the editor in sync across different components should be possible (and likely much easier with Redux). Again, this is more of a response to the readability/responsibility of the component and it's not a blocker in any way but I'd be happy to continue the discussion and pair on this at some point!

PR #1573, opened 2018-02-07
@eveessex => [Socket] filter sessions by channel

l2succes:
To improve security and reduce payload sizes, this modifies websocket handlers to only send to clients sessions for the currently active channel

This also fixes the following bug: https://sentry.io/artsynet/positron-staging/issues/451759495/

### DIFF
--- a/client/apps/articles_list/icons.jade
+++ b/client/apps/articles_list/icons.jade
@@ -1,2 +1,0 @@
-.new-article
-  include ../../components/layout/public/icons/layout_new_article.svg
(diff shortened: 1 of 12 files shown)


client/apps/articles_list/icons.jade
```diff
@@ -1,2 +0,0 @@
-.new-article
-  include ../../components/layout/public/icons/layout_new_article.svg
```
eessex: intentionally removed?

eessex:
Excellent improvements!

PR #1577, opened 2018-02-08
[Articles] fixes unpublished articles not showing

l2succes:
Fixes bug where articles drafts list did not load.

### DIFF
--- a/client/apps/articles_list/components/articles_list.jsx
+++ b/client/apps/articles_list/components/articles_list.jsx
@@ -18,7 +18,9 @@ export class ArticlesList extends Component {
     channel: PropTypes.object,
     apiURL: PropTypes.string,
     user: PropTypes.object,
-    viewArticlesAction: PropTypes.func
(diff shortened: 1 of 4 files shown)


client/apps/articles_list/components/articles_list.jsx
```diff
@@ -142,7 +144,8 @@ export class ArticlesList extends Component {
 
 const mapStateToProps = (state) => ({
   channel: state.app.channel,
-  apiURL: state.app.apiURL
+  apiURL: state.app.apiURL,
+  user: state.app.user
```
l2succes: that line is the fix btw, the `user` object was not being passed into props

eessex: Cool!  Thanks for adding a test for this... ready to go but you have to pass Peril first.

eessex: Also do you mind taking care of the styled-components TODO? Should be a fast fix.

PR #1578, opened 2018-02-09
[WIP] Use Redux for article editing

eessex:
Refactors the edit/content app to use Redux store as data-source for article, and for dispatching actions related to changing articles. 

This is massive, so planning on creating an /edit2 folder and PR'ing these components/sections 1 at a time.


client/actions/editActions.js
```diff
@@ -107,9 +124,65 @@ export const newSection = (type, sectionIndex) => {
...
+  const section = setupSection(type)
+
+  return (dispatch, getState) => {
+    dispatch(changeArticle('hero_section', section))
+  }
+}
+
+export const onChangeArticle = (key, value) => {
+  return (dispatch, getState) => {
+    const article = getState().edit.article
```
damassi: super minor but could maybe do `const { article } = getState().edit` or `const { edit: { article } } = getState()`

PR #1579, opened 2018-02-09
@l2succes => Add /edit2 app

eessex:
Duplicates the `/edit` app as `/edit2`

Will use this to break out small sections from the [Redux project](https://github.com/artsy/positron/pull/1578) into smaller PRs.


eessex:
Merging since trivial and doesn't changes are not used anywhere.

PR #1580, opened 2018-02-09
@l2sucess => Edit actions/reducers

eessex:
- Duplicates the `editActions` and `editReducers` files as `editActions2` and `editReducers2` 
- Updates article actions to use article from store (rather than passing attributes to actions)

These files will replace the existing `editActions` and `editReducers` once the redux component changes are merged.


client/actions/test/editActions2.test.js
```diff
@@ -0,0 +1,296 @@
...
+    window.location.assign = jest.fn()
+    article = clone(FeatureArticle)
+  })
+
+  document.body.innerHTML =
+    '<div>' +
+    '  <div id="edit-sections-spinner" />' +
+    '  <input id="edit-seo__focus-keyword" value="ceramics" />' +
+    '</div>'
+
```
damassi: Can possibly do 

```javascript
document.body.innerHTML = `
  <div>
    <div id="edit-sections-spinner" />
    <input id="edit-seo__focus-keyword" value="ceramics" /> 
  </div>
`
```

client/reducers/editReducer2.js
```diff
@@ -0,0 +1,151 @@
...
+        section: null,
+        isSaved: false
+      }, state)
+    }
+
+    case actions.RESET_SECTIONS: {
+      const article = cloneDeep(state.article)
+      const { sections } = action.payload
+
+      extend(article, { sections })
```
damassi: This seems a little weird to mutate in place. Maybe find a way to assign it to variable and then pass it in?

eessex: That makes sense, can take care of it in the actions instead

damassi: Cool, yeah a good rule of thumb is to clone rather than mutate

PR #1581, opened 2018-02-09
@eveessex => [Articles] converted stylus to styled components

l2succes:
This converts styling for `ArticlesList` from global css styling to using styled-components based styling


eessex:
LGTM!

PR #1582, opened 2018-02-12
@l2succes => Edit header uses Redux

eessex:
Converts the header menu to use plain objects and updated Redux actions.


client/apps/edit2/components/header/index.jsx
```diff
@@ -170,17 +182,20 @@ export class EditHeader extends Component {
...
+            color={this.getSaveColor()}
             onClick={this.onSave}
           >
             {this.getSaveText()}
-          </button>
+          </SaveButton>
 
-          <a href={article.getFullSlug()} target='_blank'>
+          <a
+            href={`${sd.FORCE_URL}/article/${article.slug}`}
```
l2succes: I added `forceURL` to our redux store (https://github.com/artsy/positron/blob/master/client/reducers/appReducer.js#L8). 
As a rule of thumb, I think it's a good idea to limit usage of `sd` in components and try to limit its use to set up `initialState` in our reducers. That'll increase portability

eessex: Sounds good, I'll update this now

PR #1583, opened 2018-02-12
@kanaabe => Section footer uses Redux actions

eessex:
Updates section footer to use Redux's article data and article change action.

### DIFF
--- a/client/apps/edit2/components/content/sections/footer/test/index.test.js
+++ b/client/apps/edit2/components/content/sections/footer/test/index.test.js
@@ -1,6 +1,6 @@
 import React from 'react'
-import Backbone from 'backbone'
 import { mount } from 'enzyme'
+import { clone } from 'lodash'
 import { Fixtures } from '@artsy/reaction-force/dist/Components/Publishing'
 import Paragraph from '../../../../../../../components/rich_text/components/paragraph.coffee'
 import { SectionFooter } from '../index.jsx'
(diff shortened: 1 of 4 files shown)


client/apps/edit2/components/display/test/email.test.js
```diff
@@ -6,7 +6,7 @@ import { CharacterLimit } from '../../../../../components/character_limit'
 import { DisplayEmail } from '../components/email'
 import ImageUpload from 'client/apps/edit/components/admin/components/image_upload.coffee'
 
-describe('DisplayEmail', () => {
+xdescribe('DisplayEmail', () => {
```
kanaabe: Intentional?

kanaabe:
Looks good! So much better breaking this out into smaller PRs 👏 . Just had one question about a commented test suite.

client/apps/edit2/components/display/test/email.test.js
eessex: Yes, this will be upended once all the changes are in

client/apps/edit2/components/content/sections/footer/test/index.test.js
```diff
@@ -11,9 +11,9 @@ describe('SectionFooter', () => {
 
   beforeEach(() => {
     props = {
-      article: new Backbone.Model(FeatureArticle),
+      article: clone(FeatureArticle),
```
kanaabe: This probably doesn't affect the test but it might be preferred to use cloneDeep over clone for objects with deep nesting as clone only does a shallow copy.

eessex: Sounds good, will update that now

PR #1584, opened 2018-02-12
@damassi => Section controls use redux

eessex:
Refactors SectionControls to use Redux data and actions.  Also moves layout controls into a sub-component.

### DIFF
--- a/client/apps/edit2/components/content/section_controls/index.jsx
+++ b/client/apps/edit2/components/content/section_controls/index.jsx
@@ -1,29 +1,39 @@
+// A container for section inputs
+// Position changes on scroll to stick to section top
+
+import styled from 'styled-components'
 import PropTypes from 'prop-types'
 import React, { Component } from 'react'
 import { connect } from 'react-redux'
-import { IconImageFullscreen } from '@artsy/reaction-force/dist/Components/Publishing'
+import LayoutControls from './layout'
 
 export class SectionControls extends Component {
   static propTypes = {
-    articleLayout: PropTypes.string,
     channel: PropTypes.object,
     children: PropTypes.any,
     disabledAlert: PropTypes.func,
     isHero: PropTypes.bool,
-    onChange: PropTypes.func,
-    section: PropTypes.object.isRequired,
-    sectionLayouts: PropTypes.bool
+    showLayouts: PropTypes.bool
   }
 
-  state = { insideComponent: false }
+  state = {
+    insideComponent: false
+  }
 
-  componentDidMount () {
+  componentDidMount = () => {
     this.setInsideComponent()
-    window.addEventListener('scroll', this.setInsideComponent)
+
+    window.addEventListener(
+      'scroll',
+      this.setInsideComponent
+    )
   }
 
-  componentWillUnmount () {
-    window.removeEventListener('scroll', this.setInsideComponent)
+  componentWillUnmount = () => {
+    window.removeEventListener(
+      'scroll',
+      this.setInsideComponent
+    )
   }
 
   setInsideComponent = () => {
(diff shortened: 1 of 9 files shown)


client/apps/edit2/components/content/section_controls/index.jsx
```diff
@@ -1,29 +1,39 @@
+// A container for section inputs
+// Position changes on scroll to stick to section top
```
damassi: This is a small nit, but when documenting at the top of a file its more idiomatic to use multi-line comments, which all editors will create by typing `/**`, and then hit enter. This will expand to 

```javascript
/**
 * A container for section inputs. Position changes on scroll to stick to section top.
 * ... 
 */ 
```


```javascript

File: client/apps/edit2/components/content/section_controls/index.jsx [hunk 2]
```diff
@@ -34,143 +44,122 @@ export class SectionControls extends Component {
...
+    if (this.controls) {
+      const controlsHeight = $(this.controls).height()
+      const offsetTop = $section.offset().top
 
-  toggleImageSet = () => {
-    if (this.props.section.get('type') === 'image_collection') {
-      this.props.section.set({type: 'image_set', layout: 'mini'})
-      this.forceUpdate()
+      const scrollPlusHeader = window.scrollY + this.getHeaderHeight()
+      const offsetMinusControls = offsetTop - controlsHeight
```
damassi: 👍

client/apps/edit2/components/content/section_controls/index.jsx
```diff
@@ -34,143 +44,122 @@ export class SectionControls extends Component {
...
+      return scrollPlusHeader > offsetMinusControls
     }
-    this.props.onChange && this.props.onChange()
   }
 
-  hasImageSet () {
-    const { type } = this.props.channel
-    return this.sectionIsImage() && type === 'editorial' || type === 'team'
-  }
+  isScrolledPast = ($section) => {
```
damassi: Its common to prefix booleans with `is`, but can also use other forms such as `hasScrolledPast` if it reads a bit better. Either way; `is` is definitely a noticeable flag.

client/apps/edit2/components/content/section_controls/index.jsx
```diff
@@ -34,143 +44,122 @@ export class SectionControls extends Component {
...
 
     return (
-      <div
-        ref='controls'
-        className={'edit-controls' + isSticky}
-        style={{
-          position: insideComponent ? 'fixed' : outsidePosition,
-          bottom: this.getPositionBottom()
+      <SectionControlsContainer
+        innerRef={(node) => { this.controls = node }}
```
damassi: What does `innerRef` refer to?

client/apps/edit2/components/content/section_controls/index.jsx
```diff
@@ -34,143 +44,122 @@ export class SectionControls extends Component {
...
 })
 
 export default connect(
   mapStateToProps
 )(SectionControls)
+
+const SectionControlsContainer = styled.div`
+  bottom: ${props => props.bottom};
+  position: ${props => props.position};
+`
```
damassi: I think it's ~ ok to start mixing in styled-components with old CSS but I would also be wary of creeping drift as this kind of thing is typically refactored out all at once (in phases) and can get increasingly unmaintainable otherwise. Def don't want future devs adding stuff to both systems when a single system can be used.

client/apps/edit2/components/content/section_controls/index.jsx
eessex: innerRef is required to use refs with StyledComponents

client/apps/edit2/components/content/section_controls/layout.jsx
```diff
@@ -0,0 +1,123 @@
...
+}
+
+const mapStateToProps = (state) => ({
+  article: state.edit.article,
+  channel: state.app.channel,
+  section: state.edit.section
+})
+
+const mapDispatchToProps = {
+  onChange: onChangeSection
```
damassi: Maybe `onChangeSectionAction` so that its clear exactly whats going on?

client/apps/edit2/components/content/section_controls/index.jsx
eessex: This comes from ongoing conversation with @kanaabe -- rather than using inline styles to pass style-related props we have been trying to transition to StyledComponents.

client/apps/edit2/components/content/section_controls/layout.jsx
```diff
@@ -0,0 +1,123 @@
...
+    }
+    onChange('layout', layout)
+  }
+
+  toggleImageSet = () => {
+    const { section, onChange } = this.props
+
+    if (section.type === 'image_collection') {
+      onChange('type', 'image_set')
+      onChange('layout', 'mini')
```
damassi: Is it necessary to fire two back-to-back actions (leading to two render cycles), or could this be batched like 

```javascript
onChange({
  type: 'image_set',
  layout: 'mini'
})
```

client/apps/edit2/components/content/section_controls/layout.jsx
```diff
@@ -0,0 +1,123 @@
...
+
+    return type.includes('image')
+  }
+
+  sectionHasFullscreen = () => {
+    const { article, section } = this.props
+
+    const isFeature = article.layout === 'feature'
+    const isImage = this.sectionIsImage()
+    const isMedia = ['embed', 'video'].includes(section.type)
```
damassi: Ah, was not aware of `.includes` existing on `Array` (ty)

client/apps/edit2/components/content/section_controls/layout.jsx
eessex: This is on my list of TODOs, but was not planning on changing in this PR.

client/apps/edit2/components/content/section_controls/test/index.test.js
```diff
@@ -1,188 +1,107 @@
...
+const { StandardArticle } = Fixtures
 
 describe('Section Controls', () => {
   let props
 
+  SectionControls.prototype.isScrollingOver = jest.fn().mockReturnValue(true)
+  SectionControls.prototype.isScrolledPast = jest.fn().mockReturnValue(false)
+  $.fn.offset = jest.fn().mockReturnValue({top: 200})
+  $.fn.height = jest.fn().mockReturnValue(200)
+  $.fn.closest = jest.fn().mockReturnThis()
```
damassi: Nice, mocking little jquery plugins for the test :)

client/apps/edit2/components/content/section_controls/test/index.test.js
```diff
@@ -1,188 +1,107 @@
...
+      app: {
+        channel: props.channel
+      },
+      edit: {
+        article: props.article,
+        section: props.section,
+        sectionIndex: props.sectionIndex
+      }
+    })
+
```
damassi: I think i've mentioned this before, but if you check out [here](https://github.com/artsy/force/blob/master/desktop/apps/auction/components/__tests__/Layout.test.js#L21) and [here](https://github.com/artsy/force/blob/master/desktop/apps/auction/__tests__/utils/renderTestComponent.js) I've written a redux test wrapper that you could adapt that makes most boilerplate go away. (It could probably be cleaned up some too):

API is: 

```javascript
const { wrapper } = renderTestComponent({
  Component: Layout,
  options: { renderMode: 'render' }, // from enzyme; shallow, render, mount
  // data is what is passed into the store before it is passed into the provider, and could  
  // be renamed to `initialState` for clarity
  data: {
    app: {
      auction: {
        name: 'An Auction'
      }
    }
  },
  // Props are whats passed into the component, and can potentially override 
  // data in initial state. 
  props: {
    ... 
  }
})
```

No pressure at all to use, just a suggestion. Since you're going down the redux route it would be good to set up a utility helper sooner rather than later that can hide some redundancy.

client/apps/edit2/components/content/section_controls/test/layout.test.js
```diff
@@ -0,0 +1,153 @@
...
+    })
+
+    return mount(
+      <Provider store={store}>
+        <section>
+          <LayoutControls {...props} />
+        </section>
+      </Provider>
+    )
+  }
```
damassi: All that said, I like this pattern and it seems to be as straightforward as can be; anyone arriving at this code would be able to immediately understand and tweak, without having to navigate through intermediary files / libs / etc  👍

client/apps/edit2/components/content/section_controls/index.jsx
```diff
@@ -34,143 +46,122 @@ export class SectionControls extends Component {
...
-        {
-          this.hasImageSet() &&
-          <a
-            name='image_set'
-            className='layout'
-            onClick={this.toggleImageSet}
-            data-active={section.get('type') === 'image_set'} />
+  insideComponent = () => {
+    const { isHero } = this.props
+    if (this.controls) {
```
damassi: Minor, but is there a way to exit early from this without the need for a guard? I noticed that in a few spots this is blocking execution. Would it be possible to move it up a bit, outside of this component and into the parent? Something like 

```javascript
{showControls && 
  <SectionControls /> }
```

client/apps/edit2/components/content/section_controls/index.jsx
damassi: Sounds good 👍, just something to be cautious about and not abuse.

client/apps/edit2/components/content/section_controls/index.jsx
eessex: My understanding of the new ref style is that the component doesn't have access to the ref until after mount -- so these are checking to see if it exists before attempting to performing jquery actions on the dom.

client/apps/edit2/components/content/section_controls/test/index.test.js
eessex: Is true there are a number of places where this code is repeated -- will look into adding something to our test utils for these cases.  There are a few examples that require custom implementations -- for example in these tests we need to wrap the component in `<section>` for jQuery actions to work. So perhaps we'll end up with a mix of both.

damassi: In terms of tests I don't necessarily think repetition is bad so either way 👍  whats more important is just keeping things literal, simple and clear, which current code definitely is.

client/apps/edit2/components/content/section_controls/index.jsx
damassi: I see 👍

PR #1586, opened 2018-02-12
@kanaabe => Embed section uses Redux

eessex:
Converts embed section to use plain objects and Redux actions.


kanaabe:
Woohoo! This is great.

PR #1590, opened 2018-02-12
@eveessex => fixes server crash when invalid session is updated

l2succes:
Fixes: https://sentry.io/artsynet/positron-staging/issues/461198203/


eessex:
Awesome, looks good when green!

PR #1594, opened 2018-02-14
@kanaabe => AutocompleteList component

eessex:
A wrapper component for Autocomplete2, emulates the coffee autocomplete-select-list used in Admin panel. 

screen shot 2018-02-14 at 2 20 09 pm

### DIFF
--- a/client/components/autocomplete2/list.jsx
+++ b/client/components/autocomplete2/list.jsx
@@ -1,0 +1,87 @@
+import { clone } from 'lodash'
+import PropTypes from 'prop-types'
+import React, { Component } from 'react'
+import { Autocomplete } from '/client/components/autocomplete2/index'
+
+export class AutocompleteList extends Component {
+  static propTypes = {
+    className: PropTypes.string,
+    disabled: PropTypes.bool,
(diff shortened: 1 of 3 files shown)


client/components/autocomplete2/list.jsx
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  render () {
+    const { className, formatSelected } = this.props
+    const { items } = this.state
+
+    return (
+      <div className={`Autocomplete--list ${className || ''}`}>
+
+        <div className='Autocomplete__list'>
```
kanaabe: Ah took a minute to understand the BEM logic in this className and the one above and not sure if I still do 😛 . Wondering if this could benefit from changing to `AutocompleteList` as the parent and `AutocompleteList__list` or `autocomplete-list` and `autocomplete-list__list` if you prefer dash style spacing?

eessex: Good call, will change to AutocompleteList -- was using the bem style of variations because the Autocomplete classes are shared between different components, but your suggestion of using component name is clearer.

PR #1596, opened 2018-02-14
@kanaabe => Redux Display

eessex:
- Updates article Display app to use redux store and actions
- adds `client/models/article.js` -- a file of utils to replace backbone model 

For the model replacement file, unsure if there is a better place for this? Happy for second opinions.  It is not truly a model and more like Article utils.


client/apps/edit/components/admin/test/components/tags.test.js
```diff
@@ -1,6 +1,6 @@
 import { mount } from 'enzyme'
 import React from 'react'
-import Article from '../../../../../../models/article'
+import Article from '../../../../../../models/article.coffee'
```
kanaabe: Ah while you're in these js ones, might be nice to make them absolute? would make searching for these much easier in the future.

kanaabe: But I'm also curious now - were these just not working?

client/apps/edit2/components/display/components/magazine.jsx
```diff
@@ -1,61 +1,78 @@
...
+})
+
+const mapDispatchToProps = {
+  onChangeArticleAction: onChangeArticle
 }
+
+export default connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(DisplayMagazine)
```
kanaabe: These are so nice now 😭

client/apps/edit/components/admin/test/components/tags.test.js
eessex: Coffee absolute paths don't work in tests because of how compiling is set up.

client/apps/edit2/components/display/components/preview/social_preview.jsx
```diff
@@ -25,18 +34,18 @@ export const SocialPreview = (props) => {
...
+          {image
+            ? <img src={crop(image, {width: 300, height: 155})} />
             : <div className='edit-display__prev--x' />
           }
           </div>
 
           <div className='edit-display__prev-social--meta'>
             <div className='edit-display__prev-social--headline'>
-              {article.getThumbnailTitle('social_title')}
+              {social_title || thumbnail_title}
```
kanaabe: Interesting this sort of makes things less reliable now that business logic isn't tied to a model. I was hoping that we could make some utilities or a "model-like" file for managing this logic into real methods. I've seen this kind of problem before where denormalizing business logic creates inconsistencies and makes things difficult to find later on.

client/apps/edit2/components/display/index.jsx
```diff
@@ -1,63 +1,55 @@
...
+              className='admin-form-container max-width-container'
+              activeSections={[0]}
+              openMany
+              sections={sections}
+            >
+              <DisplayMagazine />
+              <DisplaySocial />
+              <DisplaySearch />
+              <DisplayEmail />
+            </DropDownList>
```
kanaabe: Oof so nice to clean this up!

client/apps/edit2/components/display/components/preview/social_preview.jsx
eessex: Right this is why I started the other file.  In the case above, I think that actually using a conditional makes it clearer what is happening, because it isn't actually complex logic. One thing i noticed is that we have model logic that isn't used, or used in one place, but there are cases like the byline where it actually does make sense to move out.

eessex: I also tried to move some of the model functions into redux actions so that we'd have access to the data and logic everywhere, but it didn't seem to make sense in the end, because it wants you to store the return value from the methods in state.

client/apps/edit/components/admin/test/components/tags.test.js
eessex: also most of these references will not exist after we merge the rest of the redux things

client/apps/edit2/components/tests/edit_container.test.js
```diff
@@ -161,14 +160,14 @@ describe('EditContainer', () => {
...
-      it('Can render the admin activeView', () => {
+      xit('Can render the admin activeView', () => {
         props.activeView = 'admin'
         const component = getShallowWrapper(props)
 
         expect(component.find(EditAdmin).exists()).toBe(true)
       })
 
-      it('Can render the admin activeView', () => {
+      xit('Can render the display activeView', () => {
```
kanaabe: These are intentional?

client/models/article.js
```diff
@@ -0,0 +1,12 @@
...
+
+export const getArticleByline = (article) => {
+  const { contributing_authors, author } = article
+
+  if (contributing_authors && contributing_authors.length) {
+    return _s.toSentence(pluck(contributing_authors, 'name'))
+  } else if (author.name) {
+    return author.name
+  }
+}
```
kanaabe: Ah yeah, this is was I was thinking about, sorry didn't see it until now. I guess if we aren't using the other methods more than once, it's fine to keep the logic inline. But I do like having this as a home for this kind of business logic 👍

client/apps/edit2/components/tests/edit_container.test.js
eessex: Yes haven't set up this component to use the connected children yet, depends on other PRs to come

client/models/article.js
eessex: Does this seem like an appropriate place to store these files?  Wondering if it is correct to call it a model if it is not really a class but just a collection of utils files.

client/apps/edit2/components/display/components/preview/social_preview.jsx
kanaabe: Yeah if the logic in `getThumbnailTitle` isn't used anywhere else I think it's fine to keep here.

kanaabe:
Woohoo! This is awesome.

PR #1597, opened 2018-02-15
[Edit] fixes draft editor to never show lock modal

l2succes:
Fix for a bug that affects editing a new draft. It should not show the lock modal in that case

### DIFF
--- a/client/apps/edit/components/edit_container.jsx
+++ b/client/apps/edit/components/edit_container.jsx
@@ -48,7 +48,7 @@ export class EditContainer extends Component {
       lastUpdated: null,
       isOtherUserInSession: !!props.currentSession && !isCurrentUserEditing,
       inactivityPeriodEntered: false,
-      shouldShowModal: true,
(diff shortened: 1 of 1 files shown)


l2succes:
It's a pretty trivial fix so self-merging this.

client/apps/edit/components/edit_container.jsx
```diff
@@ -48,7 +48,7 @@ export class EditContainer extends Component {
       lastUpdated: null,
       isOtherUserInSession: !!props.currentSession && !isCurrentUserEditing,
       inactivityPeriodEntered: false,
-      shouldShowModal: true,
+      shouldShowModal: !props.article.isNew(),
```
damassi: @l2succes - Since this is a bug could possibly use a test

l2succes: Yes for sure. I'll add tests in the actual PR. Just wanted to get this out to unblock the editor team

PR #1600, opened 2018-02-20
@l2succes => Redux section container

eessex:
Updates section container to use data and actions from Redux.


eessex:
The actions are now in `/editActions2`

PR #1604, opened 2018-02-21
@kanaabe => Move fetch mentioned to article model

eessex:
In anticipation of redux admin panel, duplicated functions for fetching mentioned artists and artworks out of section model and sections collection into article model. 

The Backbone functionality can be removed after the new admin panel is finished (working through some metaphysics/gravity issues before this is finished)


kanaabe:
Cool, I'm digging this game plan

PR #1605, opened 2018-02-22
[QA] fixes queue section & force edit URL

l2succes:
This fixes 
- the queue tab which was broken due to porting the `ArticleList` component from CoffeeScript to ES6 and the recent introduction of Redux.

-  the lockout modal not rendering when routed from slug based urls (force)


kanaabe:
LGTM!

PR #1609, opened 2018-02-26
@eessex => Bump reaction, rename reaction-force, bring edit2 up to date

kanaabe:
- Bumps reaction to the latest and switches from `reaction-force` -> `reaction`
- Brings in the latest socket changes to edit2 (use of `getSessionsForChannel`)

This doesn't fix the local "Page not found" issue, but I've determined there's currently no issue with Reaction versions + Positron. Instead, the problem with developing locally is a race condition (or initial mount issue, not sure yet) between `express-reloadable` and some recent additions to lock out. This is what happens: 

1. App/express-reloadable mounts for NODE_ENV=development
2. This fetch fails -> https://github.com/artsy/positron/blob/master/client/apps/websocket/index.js#L16
3. Because it fails, we catch it here and `next` it https://github.com/artsy/positron/blob/master/client/apps/articles_list/routes.coffee#L29

In the meantime, any client-side changes will trigger a reload and fix the fetching issue.


client/lib/setup/session.coffee
```diff
@@ -11,9 +11,9 @@ sess = exports.defaultSession = session
...
   io.use sharedsession(sess)
-  
+
   # saves having to type "socket.request.session.user" everywhere
   io.use (socket, next) ->
     user = get socket, "request.session.user"
     socket.user = user if user
-    next()
\ No newline at end of file
+    next()
```
eessex: 👍

PR #1611, opened 2018-02-28
[Dev] update @artsy/express-reloadable

l2succes:
Includes the fix for obfuscated errors. Read more here https://github.com/artsy/express-reloadable/pull/8

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "react-relay": "https://github.com/alloy/relay/releases/download/v1.5.0-plugin.2/react-relay-1.5.0-plugin.2.tgz"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


kanaabe:
Just tried to no avail :/

damassi:
Ah, ok will look into it. Sorry for the delay

l2succes:
Hmm, I'm not seeing on my end right now. @kanaabe Are you still seeing the following error?
`Error: Can't set headers after they are sent.`

kanaabe:
I think the issue we were seeing was separate, had to do with load order right? Not muted errors. Nw, thanks. There's no urgency btw since we have a workaround.

damassi:
@kanaabe -  yup it did, but just wanted to be 100% sure

PR #1613, opened 2018-03-02
Bump reaction

kanaabe:
0.26.4


kanaabe:
@eessex note that I've pended a few tests on edit2 here due to passing in Backbone articles to Reaction components. Don't have the time rn to fix them but since they aren't being used anywhere atm and we need to ship a fix for display ads, I figured I'd pend.

PR #1617, opened 2018-03-07
@kanaabe => Redux admin

eessex:
- Converts  edit2/admin app to use redux data and actions
- Converts coffee admin components to jsx


client/apps/edit2/components/admin/components/verticals_tags.jsx
```diff
@@ -1,18 +1,20 @@
...
 import { Col, Row } from 'react-styled-flexboxgrid'
-import { data as sd } from 'sharify'
+import { connect } from 'react-redux'
 import Verticals from '../../../../../collections/verticals.coffee'
 import { AutocompleteInlineList } from '/client/components/autocomplete2/inline_list'
+import { onChangeArticle } from 'client/actions/editActions'
 
 export class AdminVerticalsTags extends Component {
   static propTypes = {
+    apiURL: PropTypes.string,
```
kanaabe: Curious why you want to make this into a prop instead of pulling from Sharify?

eessex: Really could go either way -- some of these urls worked their way into the store during the lockout project, which also meant they were available within the redux actions without having to loop in sd.  Since they were already there, I started using them rather that requiring another library.

eessex: they are in the app reducer: https://github.com/artsy/positron/blob/master/client/reducers/appReducer.js

client/apps/edit2/components/admin/components/metaphysics_autocomplete.jsx
```diff
@@ -0,0 +1,130 @@
...
+import { clone, uniq } from 'lodash'
+import { connect } from 'react-redux'
+import { difference, flatten, pluck } from 'underscore'
+import PropTypes from 'prop-types'
+import React, { Component } from 'react'
+import { onChangeArticle } from 'client/actions/editActions'
+import { AutocompleteList } from '/client/components/autocomplete2/list'
+import * as Queries from 'client/queries/metaphysics'
+
+export class MetaphysicsAutocomplete extends Component {
```
kanaabe: Don't _love_ the naming here -- maybe `AutocompleteListMetaphysics`? Something to indicate a list would be nice.

client/apps/edit2/components/admin/components/metaphysics_autocomplete.jsx
```diff
@@ -0,0 +1,130 @@
...
+      model,
+      onChangeArticleAction,
+      placeholder
+    } = this.props
+
+    return (
+      <div className='field-group'>
+        <label>{label}</label>
+        <AutocompleteList
+          fetchItems={(fetchedItems, cb) => this.fetchItems(fetchedItems, cb)}
```
kanaabe: Re-remembering this pattern now and still kinda found it hard to follow 😄 . I don't know what a simpler solution is atm but it does feel like something we can improve on in the future.

client/apps/edit2/components/admin/test/components/metaphysics_autocomplete.test.js
```diff
@@ -0,0 +1,186 @@
...
+        return cb(response.body.data.fairs)
+      })
+      const component = getWrapper(props)
+      component.instance().fetchItems(['123'], cb)
+
+      expect(cb.mock.calls[0][0][0]._id).toBe('123')
+      expect(cb.mock.calls[0][0][0].name).toBe('NADA New York')
+    })
+  })
+})
```
kanaabe: Legit set of tests!

client/reducers/appReducer.js
```diff
@@ -1,12 +1,14 @@
...
-  user: sd.USER,
+  apiURL: sd.API_URL,
   appURL: sd.APP_URL,
+  artsyURL: sd.ARTSY_URL,
+  channel: sd.CURRENT_CHANNEL,
   forceURL: sd.FORCE_URL,
-  apiURL: sd.API_URL
+  isAdmin: sd.USER.type === 'Admin',
+  metaphysicsURL: sd.GRAPHQL_ENDPOINT,
+  user: sd.USER
```
kanaabe: Nice ok starting to see how this is helpful. I can get behind this for sure.

client/apps/edit2/components/admin/components/metaphysics_autocomplete.jsx
eessex: 👍

client/apps/edit2/components/admin/components/appearances.jsx
```diff
@@ -0,0 +1,49 @@
...
+            placeholder='Search by fair name...'
+          />
+        </Col>
+
+        <Col xs={6}>
+          <MetaphysicsAutocomplete
+            field='biography_for_artist_id'
+            label='Extended Artist Biography'
+            model='artists'
+            placeholder='Search by artist name...'
```
kanaabe: Would this know to only allow one entry?

client/apps/edit2/components/admin/components/metaphysics_autocomplete.jsx
eessex: Yeah the autocomplete involves a lot of variables, which was part of the motivation for metaphysics-specific component- that makes it easy to just pass a model, label, field and placeholder. This component could probably be improved so that you just pass a model and field and everything else is a default. We could also do a positron specific version so that we don’t ever really need to touch the original autocomplete fetches etc.

client/apps/edit2/components/admin/components/appearances.jsx
eessex: The autocomplete accepts an array of ids, but it could probably be modified to limit one.

eessex: Ooh I see now, looks like artist bio is limited to one 👌

client/apps/edit2/components/admin/components/article/index.jsx
```diff
@@ -0,0 +1,211 @@
...
+}
+
+export default connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(AdminArticle)
+
+const Button = styled.button`
+  padding: 10px 20px;
+  border: 2px solid #ccc;
```
kanaabe: Is this available in `colors._` ?

client/apps/edit2/components/admin/components/article/index.jsx
```diff
@@ -0,0 +1,211 @@
...
+                <ButtonGroup>
+                  <Button
+                    active={article.layout === 'standard'}
+                    onClick={() => onChangeArticleAction('layout', 'standard')}
+                  >
+                    Standard
+                  </Button>
+                  <Button
+                    active={article.layout === 'feature'}
+                    onClick={() => onChangeArticleAction('layout', 'feature')}
```
kanaabe: My bad if you address this later on, but does still have the `alert()` feature somewhere?

client/actions/editActions2.js
```diff
@@ -284,6 +287,65 @@ export const saveArticlePending = () => {
   }
 }
 
+export const onAddFeature = (model, item) => {
```
kanaabe: Funny I had a conversation with Jun yesterday about our use of the word "Feature" on publishing and here is another instance! Could we be a little more verbose here? `onAddFeaturedEntity` or `onAddLinkedEntity`? And maybe below to match Mentioned too?

client/components/autocomplete2/list.jsx
```diff
@@ -54,34 +57,51 @@ export class AutocompleteList extends Component {
...
+  display: flex;
+  justify-content: space-between;
+  letter-spacing: 0;
+  line-height: 26px;
+  margin-bottom: 10px;
+  overflow: ellipsis;
+  padding: 5px 20px 5px 10px;
+  position: relative;
+  text-transform: none;
+`
```
kanaabe: Nice. Glad you followed up with this

client/actions/editActions2.js
eessex: Yes great suggestion

client/apps/edit2/components/admin/components/featuring/featuring_input.jsx
```diff
@@ -0,0 +1,94 @@
...
+})
+
+const mapDispatchToProps = {
+  onAddFeatureAction: onAddFeature
+}
+
+export default connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(FeaturingInput)
```
kanaabe: I know this is replicating how it it's currently implemented, but reading `FeaturingList` and `FeaturingInput` again and I'm wondering if we could just replace both of these with an Autocomplete? Is there any reason not to?

client/apps/edit2/components/admin/components/featuring/mentioned_list.jsx
```diff
@@ -0,0 +1,153 @@
...
+      this.getMentionedArtists()
+    } else {
+      this.getMentionedArtworks()
+    }
+  }
+
+  getMentionedArtists = () => {
+    const { article, setMentionedAction } = this.props
+    const artists = new Artists()
+    const ids = ArticleUtils.getMentionedArtistSlugs(article)
```
kanaabe: Cool 🆒 Cool 🆒

client/apps/edit2/components/admin/index.jsx
```diff
@@ -60,9 +52,7 @@ export class EditAdmin extends Component {
           }
 
           <AdminArticle />
-
-          <AdminFeaturing {...this.props} />
-
+          <AdminFeaturing />
```
kanaabe: Must feel nice to strip {...this.props} 😸

client/reducers/editReducer2.js
```diff
@@ -112,6 +112,7 @@ export function editReducer (state = initialState, action) {
 
     case actions.CHANGE_SECTION: {
       const { key, value } = action.payload
+      console.log(key, value)
```
kanaabe: Stray log

kanaabe:
Whew took a little while to review but I gotta say 💯 job on this and congrats! 🎊 . Can't wait to look back on this in 100 years and wonder how we ever lived without proper state management in Positron.

Seems like you have a few things to update and tests. Let me know when I can do the second pass. Also really appreciate the broken out commits. Made this much more enjoyable and consumable to read.

client/apps/edit2/components/admin/components/article/index.jsx
eessex: Ooh good point - I will add this back

client/apps/edit2/components/admin/components/featuring/featuring_input.jsx
eessex: This does make more sense than adding from a url, will explore that

client/apps/edit2/components/admin/components/featuring/mentioned_list.jsx
eessex: Didn’t get around to it, but we could add metaphysics/gravity support For fetching by [:id] and eliminate the backbone models at some point.

client/apps/edit2/components/admin/components/article/index.jsx
eessex: We are not actually removing any data here now that things are in redux-- do you think it is better to remove it (potentially will never display if changed to standard), or add an alert and manually delete info?

client/apps/edit2/components/display/test/social.test.js
```diff
@@ -6,9 +6,9 @@ import { Fixtures } from '@artsy/reaction/dist/Components/Publishing'
 import { Provider } from 'react-redux'
 import { CharacterLimit } from 'client/components/character_limit'
 import { DisplaySocial } from '../components/social'
-import ImageUpload from 'client/apps/edit2/components/admin/components/image_upload.coffee'
+import ImageUpload from '../../admin/components/image_upload.coffee'
 
-describe('DisplaySocial', () => {
+xdescribe('DisplaySocial', () => {
```
kanaabe: I see we're pending quite a few tests but I presume these will work when we swap edit2?

eessex: Yeah they are passing locally, so i think will be easier to debug when everything is switched over.

client/actions/editActions2.js
```diff
@@ -289,50 +288,12 @@ export const saveArticlePending = () => {
...
-    dispatch(changeFeaturedItems(newFeatured))
-  }
-}
-
-export const changeFeaturedItems = (featured) => {
-  return {
-    type: actions.CHANGE_FEATURED_ITEMS,
-    payload: {
-      featured
-    }
```
kanaabe: Nice. Really glad we did this.

PR #1618, opened 2018-03-08
@kanaabe => Redux series and video layouts

eessex:
A few loose ends that I forgot to PR separately -- 
- Series layout uses redux data and store
- Video layout uses redux data and store
- Section list uses redux data and store
- EditArticle uses redux components


eessex:
This is the last step!  Have a branch ready to go to turn on all redux when this is merged 🎉

PR #1622, opened 2018-03-12
@eessex => Redux QA

kanaabe:
See commit messages for fixes -- ⬇️


client/models/user.coffee
```diff
@@ -48,10 +48,14 @@ module.exports = class User extends Backbone.Model
...
 
+
       for attr in ['id', 'type', 'name', 'email']
         return callback true unless _.isEqual user[attr], @get(attr)
       for attr in ['channel_ids', 'partner_ids']
-        return callback true unless _.isEqual(user[attr].sort(), @get(attr).sort())
+        return callback true unless _.isEqual(
+          _.clone(user[attr]).sort(),
+          _.clone(@get(attr)).sort()
+        )
```
kanaabe: Seemed to just happen locally, but we were getting an error that sorting wasn't allowed on these read-only properties: 

image

client/apps/edit/components/admin/components/autocomplete_list_metaphysics.jsx
```diff
@@ -38,7 +38,7 @@ export class AutocompleteListMetaphysics extends Component {
         return Queries.PartnersQuery
       }
       case 'partner_shows': {
-        return Queries.PartnersQuery
+        return Queries.ShowsQuery
```
kanaabe: In addition to this, there's an error with this query because MP does not yet support `ids` as an argument. I feel like you talked about this before so lmk if you already have a working PR otherwise I'm happy to implement in MP

client/models/user.coffee
eessex: 👍

client/apps/edit/components/admin/components/autocomplete_list_metaphysics.jsx
eessex: Hm maybe I missed shows -- can look at implementing tomorrow

client/apps/edit/components/admin/test/components/autocomplete_list_metaphysics.test.js
```diff
@@ -97,7 +97,7 @@ describe('AutocompleteListMetaphysics', () => {
       const component = getWrapper(props)
       const query = component.instance().getQuery('partner_shows')
 
-      expect(query).toBe(Queries.PartnersQuery)
+      expect(query).toBe(Queries.ShowsQuery)
     })
```
eessex: ooh thanks for catching

PR #1623, opened 2018-03-13
@eessex/@sarahscott => Basic News layout

kanaabe:
Didn't want to unveil the `News` option in the dropdown menu, so I hid it under a `DEV_FEATURES` flag.

- Adds the dropdown to create a News article
- Supports basic version of editing a News article
- More strict on SectionTool - exclude Video/Embed for News and exclude Embed for lower tier channels (realized this got dropped somewhere along the way)

NOTE: This does not include the `social_embed` component. I was thinking it'd be a nice one for @sarahscott to work on!

news-writer


client/apps/edit/components/content/article_layouts/news.jsx
```diff
@@ -0,0 +1,73 @@
...
+  margin: 40px auto;
+
+  .SectionText {
+    max-width: 680px;
+  }
+
+  .SectionContainer {
+    padding: 20px 0;
+  }
+`
```
eessex: Omg exciting to see how easy this looks!

client/apps/edit/components/content/article_layouts/news.jsx
```diff
@@ -0,0 +1,73 @@
...
+    return (
+      <EditNewsContainer>
+
+        <NewsHeadline
+          article={article}
+          editTitle={this.editTitle()}
+        />
+        <SectionList />
+
+        <NewsByline article={article} />
```
eessex: I've noticed that on the other headlines, social links are still functional in writer -- we should consider adding a prop to mute these links.

.env.test
```diff
@@ -15,6 +15,7 @@ EDITORIAL_CHANNEL=5086df098523e60002000018
 FB_PAGE_ID=12345
 SEGMENT_WRITE_KEY=12345
 FORCE_COLOR=1
+DEV_FEATURES=1
```
eessex: a clever solution

kanaabe:
Good catch, updated!

client/components/stylus_lib/index.styl
```diff
@@ -27,6 +27,8 @@ feature-body-width = 680px
 feature-body-w-margin = 720px
 feature-blockquote-width = 940px
 
+news-body-width = 780px
+
```
eessex: 👍 we should also think about adding some of these values to reaction as constants like colors, might make it easier to standardize.

PR #1630, opened 2018-03-14
@kanaabe => Redux QA

eessex:
- Move redirect from new articles to `saveArticle` action (from client, fixes bug of not forwarding)
- Set author_id in editReducer rather than client
- Delete unused redux/backbone sync actions in client (should have been removed at /edit2 merge)


kanaabe:
👍

PR #1637, opened 2018-03-16
Compress responses

yuki24:
Similar to https://github.com/artsy/metaphysics/pull/948 but for Positron. This will reduce approximately 80KB of data transfer for the top page between the user's browser and the Publishing API:

screen shot 2018-03-16 at 11 09 54 am

### DIFF
--- a/boot.js
+++ b/boot.js
@@ -3,6 +3,7 @@
  */
 
 import artsyXapp from 'artsy-xapp'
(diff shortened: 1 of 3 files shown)


eessex:
👏 Good idea!

kanaabe:
👍Thanks!

PR #1639, opened 2018-03-16
@kanaabe => Redirect new articles before change is allowed

eessex:
When creating a new article, waiting for the save callback while typing would cause multiple articles to be created, because the window had not yet finished forwarding to the article's static url.

This adds a save on the client `init()` if there is no id, and then waits for the article to have an id before removing the loading spinner.


kanaabe:
Noice

PR #1640, opened 2018-03-16
@eessex => Debounced autosave on articles

kanaabe:
This debounces the dispatch to `saveArticle` on autosaves. The problem was that we were piling up PUTs on the server which caused a race condition and sometimes old data was the last thing to have actually saved. 

You can see the example here -- it debounces the actions related to `SAVE_ARTICLE` but still allows `CHANGE_SECTION` to dispatch therefore making the UI update smoothly.

debounced-save

### DIFF
--- a/client/actions/editActions.js
+++ b/client/actions/editActions.js
@@ -133,7 +133,7 @@ export const newHeroSection = (type) => {
   }
 }
 
-const dispatchDebouncedSave = debounce((dispatch) => {
+const debouncedSaveDispatch = debounce((dispatch) => {
   dispatch(saveArticle())
 }, 500)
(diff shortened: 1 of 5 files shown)


client/actions/editActions.js
```diff
@@ -147,7 +151,7 @@ export const onChangeArticle = (key, value) => {
     }))
 
     if (!article.published) {
-      dispatch(saveArticle())
+      debouncedSaveDispatch(dispatch)
     }
```
eessex: ooo nice this makes more sense, had been messing with debounce in the backbone 'sync' callback and was not a functional choice.

PR #1641, opened 2018-03-17
@kanaabe => [QA] Text bugs

eessex:
The text component was not connected directly to the store, and receiving props passed from SectionContainer -- causing it to get out of sync when new sections were inserted to the article from the SectionText component. 

This PR connects the SectionText component, using data/actions from Redux.  

There is still a bit of weirdness around content-end dots, but will address in a separate pr.


kanaabe:
Nice, this is a good move 👍

eessex:
I guess a good rule of thumb is to connect anything where it is possible for props data to change outside the component. Expected that looking for a props change via `componentDidUpdate` would be enough, but it wasn't catching everything.

PR #1646, opened 2018-03-19
@kanaabe => First section of news does not allow embeds

eessex:
- SectionTool does not show social_embed for first section of news
- SectionList does not allow dragging social_embed to first section of news
- SectionList shows an error message if attempting to drag social_embed to first section
- Adds hasFeatures prop to store.app
- Adds more granular tests for SectionTool

### DIFF
--- a/client/apps/edit/components/content/section_list/index.jsx
+++ b/client/apps/edit/components/content/section_list/index.jsx
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types'
 import React, { Component } from 'react'
 import { compact } from 'lodash'
 import { connect } from 'react-redux'
-import { resetSections, setSection } from 'client/actions/editActions'
+import { logError, resetSections, setSection } from 'client/actions/editActions'
 import SectionContainer from '../section_container'
 import SectionTool from '../section_tool'
 import DragContainer from 'client/components/drag_drop/index.coffee'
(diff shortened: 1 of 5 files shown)


client/reducers/appReducer.js
```diff
@@ -6,6 +6,7 @@ export const initialState = {
   artsyURL: sd.ARTSY_URL,
   channel: sd.CURRENT_CHANNEL,
   forceURL: sd.FORCE_URL,
+  hasFeatures: sd.CURRENT_CHANNEL.type !== 'partner',
```
kanaabe: Oh no another `features` 😛 . Should we call this `notPartnerChannel`?

client/apps/edit/components/content/section_list/index.jsx
```diff
@@ -24,8 +25,17 @@ export class SectionList extends Component {
...
+      article: { layout },
+      logErrorAction,
+      resetSectionsAction
+    } = this.props
 
+    if (newSections[0].type === 'social_embed' && layout === 'news') {
+      return logErrorAction({
+        message: 'Embeds are not allowed in the first section.'
+      })
+    }
```
kanaabe: This is slick 👍

client/reducers/appReducer.js
eessex: We use the 'hasFeatures' flag in many components, mind if I make a new pr that renames and replaces all instances of it, instead of changing here?

client/apps/edit/components/content/section_list/index.jsx
eessex: One thing to note is that @l2succes made a new error modal during the lockout project, will make a new issue to convert it to use with general client errors as well. (The existing logError uses a flash error)

client/reducers/appReducer.js
eessex: Actually it doesn't affect other components when used from store.  Will update now but still add a follow-up pr.

PR #1651, opened 2018-03-20
@kanaabe => Remove content end

eessex:
HOLD pending bump of Reaction in Force.

- Stops adding content-end spans to articles
- Removes content-end spans if they exist (can be removed after backfill)
- Bumps Reaction


eessex:
PR for force side is here: https://github.com/artsy/force/pull/2286

eessex:
This is ready to review, we'll just need to deploy force before positron (there is a deploy waiting now for force, still waiting for 👍 from @xtina-starr)

PR #1654, opened 2018-03-21
@l2succes => Debounces UPDATE_ARTICLE (getSessions)

kanaabe:
Addresses: https://artsyproduct.atlassian.net/browse/GROWTH-360

Debounce the dispatch that takes care of keeping session info updated so it doesn't fire on every keystroke!


l2succes:
👍

PR #1655, opened 2018-03-21
@eessex => Add AutocompleteSingle to limit artist bio field

kanaabe:
Still WIP but the idea is that we should be able to select items that aren't saved array fields. It becomes too much hair splitting to hack this into AutocompleteList so this creates a new type which is functionally the same as the old AutocompleteSelect. Currently biography_for_artist_id is the only place that uses this, but this field is currently broken so we need to support it.

TODO
- [x] Fix remaining bugs with save/add/remove
- [x] Fix bug with removing an item in inline-list
~~- [ ] Rename AutocompleteListMetaphysics to be more generic~~ (going to remove this TODO because I feel like all the types are variants of a "list")
- [x] Move AutocompleteMetaphysics to /components
- [x] Tests

artist-bio-fixed


client/apps/edit/components/admin/components/autocomplete_list_metaphysics.jsx
```diff
@@ -18,7 +19,12 @@ export class AutocompleteListMetaphysics extends Component {
     model: PropTypes.string,
     onChangeArticleAction: PropTypes.func,
     placeholder: PropTypes.string,
-    user: PropTypes.object
+    user: PropTypes.object,
+    type: PropTypes.string
+  }
```
eessex: I know this comes from the legacy code but I find 'select' versus list to be pretty vague, and wonder if it might make sense to use a boolean of `limitOne` to invoke the select component, rather than a type.

kanaabe: Yeah, that's a good point. This is a good opportunity to be more explicit! In that case I would like to rename the sub-component too -- what are your thoughts on `AutocompleteSingle` or `AutocompleteSingleList` and then still use `limitOne` for interfacing with the general `AutocompleteMetaphysics` component?

kanaabe: `type` might still be a good option here -- if we want to include inline-list, for example.

src/client/components/autocomplete2/index.jsx
```diff
@@ -85,7 +85,12 @@ export class Autocomplete extends Component {
 
   onSelect = async (selected) => {
     const { items, onSelect } = this.props
-    const newItems = clone(items)
+    let newItems
+    if (items) {
+      newItems = clone(items)
+    } else {
+      newItems = []
+    }
```
eessex: 👍

src/client/components/autocomplete2/list_metaphysics.jsx
```diff
@@ -5,7 +5,8 @@ import { difference, flatten, pluck } from 'underscore'
 import PropTypes from 'prop-types'
 import React, { Component } from 'react'
 import { onChangeArticle } from 'client/actions/editActions'
-import { AutocompleteList } from '/client/components/autocomplete2/list'
+import { AutocompleteList } from 'client/components/autocomplete2/list'
+import { AutocompleteSingle } from 'client/components/autocomplete2/single'
 import * as Queries from 'client/queries/metaphysics'
```
eessex: This naming is great!

PR #1656, opened 2018-03-22
@kanaabe => Move source files to /src

damassi:
This follows our other JS codebases in that source-files are now located in /src. This also makes it easier to use `include` rather than `exclude` in webpack and other various file-watching utilities.


scripts/ga_airtable_transfer.js
```diff
@@ -26,6 +26,7 @@ const base = new Airtable({
 const doc = new Sheet(process.env.GOOGLE_SHEET_ID)
 
 if (process.env.NODE_ENV === 'development') {
+  // TODO: Where is this file?
```
damassi: Didn't see this file anywhere?

kanaabe: This gets dropped in outside of the project -- it's the JWT token for accessing Google Sheets but the better way to do this is to use keys (in the else statement) so it's fine to drop this conditional actually. You can leave it in for now if you want and I can remove later.

damassi: 👍

kanaabe:
Cool, thanks for this. 👍

damassi:
I think we're getting really close to parity across all of our JS codebases

PR #1658, opened 2018-03-22
Jest: Re-add --runInBand

damassi:
From looking at https://1670-23808195-gh.circle-artifacts.com/0/tmp/memory-usage.txt remembered why we initially added this a while back -- this prevents zombie process from spinning up.


damassi:
superceded by https://github.com/artsy/positron/pull/1659#issuecomment-375501598

PR #1662, opened 2018-03-26
@kanaabe => Backspace bugs

eessex:
Addresses bug of backspace not acting as expected when text is selected
- adds explicit change of section when merging sections on backspace
- checks if text is selected before manipulating
- adds explicit not-handled return value


kanaabe:
Whoops, missed this one. LGTM!

PR #1665, opened 2018-03-28
[wip] Edit Source

sarahscott:
Will move this into its own file soon, in TypeScript 😄 

screen shot 2018-03-28 at 8 08 39 pm


package.json
```diff
@@ -117,12 +117,13 @@
     "sailthru-client": "^2.0.3",
     "sharify": "^0.1.6",
     "socket.io": "^2.0.4",
-    "styled-components": "^2.1.2",
+    "styled-components": "2.4.0",
```
kanaabe: Not necessary to do this now, but I think we should be able to bump this up to 3.x without problems. We did so in Force and Reaction: https://github.com/artsy/force/pull/2136/files

kanaabe: Reaction wasn't bumped due to jest-styled-components issues which we wouldn't see here..

sarahscott: Done! 🎉

src/client/apps/edit/components/content/article_layouts/news.jsx
```diff
@@ -27,6 +34,26 @@ export class EditNews extends Component {
...
 
+  editSource = () => {
+    const { article, onChangeArticleAction } = this.props
+    const { isEditSourceOpen } = this.state
+    const { title } = article.news_source
+
+    return (
+      <AddSource onClick={() => { this.setState({ isEditSourceOpen: !isEditSourceOpen })}}>
+        {title && title.length ? title : 'Add Source'}
+      </AddSource>
```
eessex: One q that maybe requires @owendodd to weigh in -- are we requiring that both a title and URL are included?  Is there such thing as a title with no link?  If we need both, or just one, we should validate this somehow.

src/client/apps/edit/components/content/sections/news/EditSourceControls.tsx
```diff
@@ -0,0 +1,92 @@
...
+          placeholder={"Enter source name"}
+          onChange={(event) => this.setState({title: event.target.value})}
+        />
+        <ApplyInputContainer>
+          <LinkInput
+            name={"url"}
+            value={this.state.url}
+            placeholder={"Paste or type a link"}
+            onChange={(event) => this.setState({url: event.target.value})}
+          />
```
eessex: Is it possible to do some kind of validation here to make sure it is a valid url?

src/client/apps/edit/components/content/article_layouts/news.jsx
owendodd: Title should be required, link is optional, but agree we should validate the url if one is input

src/client/apps/edit/components/content/sections/news/EditSourceControls.tsx
eessex: The use case i'm thinking of is for having/not having http://, utm or something like that -- I know we do a quick check for links being valid in body text here-- https://github.com/artsy/positron/blob/master/src/api/apps/articles/model/save.coffee#L166
maybe something along these lines, or sanitize the link on the save side.

PR #1666, opened 2018-03-28
@sarahscott => Do not show layout buttons for news

eessex:
Removes toggle from 'standard' to 'feature' layout from Admin panel for articles with 'news' layout.


sarahscott:
👍

PR #1672, opened 2018-04-02
@kanaabe => Updates to Sailthru distribution

eessex:
Addresses [GROW-393](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-393)

- Removes 'magazine' tag from articles
- Adds 'layout' to custom vars
- Uses layout to construct correct article slug
- Adds 'artsy-editorial' tag to articles from the editorial channel

It wasn't completely clear from the ticket what changes were needed, please let me know if there is something missing.

### DIFF
--- a/src/api/apps/articles/model/distribute.coffee
+++ b/src/api/apps/articles/model/distribute.coffee
@@ -56,7 +56,6 @@ getArticleUrl = (article) ->
 postSailthruAPI = (article, cb) ->
   return cb() if article.scheduled_publish_at
   tags = ['article']
-  tags = tags.concat ['magazine'] if article.featured is true
   tags = tags.concat article.keywords if article.keywords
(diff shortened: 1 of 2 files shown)


kanaabe:
LGTM! Note two things:
1. This requires a backfill on Sailthru articles -- you should be able to use the backfill script and tailor it for just distribution -- run that on every Editorial article.
2. The emails team needs to know when all the backfill and new fields are in place so they know to update their feeds. I would notify them when backfill + this deploy is complete with a warning a few hours before.

eessex:
@kanaabe sounds good, should we hold this PR until tomorrow and I can touch base with email team in the AM?

kanaabe:
Yep!

PR #1675, opened 2018-04-05
@kanaabe => Use layout-based slug in deleteArticleFromSailthru

eessex:
Additional updates based on recent changes to Sailthru distribution -- uses `getArticleSlug` method for `deleteArticleFronSailthru` method as well.


kanaabe:
LGTM!

PR #1676, opened 2018-04-09
@kanaabe => [DO NOT MERGE] Backfill logic for new Sailthru tags

eessex:
Planning to run deleteArticleFromSailthru and distributeArticle as separate backfills, but wanted to make sure I'm not missing any steps here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "assets": "sh scripts/assets.sh",
     "deploy": "sh scripts/deploy.sh",
     "mocha": "sh scripts/mocha.sh",
-    "task": "node -r coffee-script/register -r babel-core/register -r dotenv/config"
+    "task": "node -r coffee-script/register -r @babel/register -r dotenv/config"
   },
   "resolutions": {
     "babel-core": "^7.0.0-0",
--- a/src/api/apps/articles/model/index.js
+++ b/src/api/apps/articles/model/index.js
@@ -13,7 +13,12 @@ import moment from 'moment'
 const db = require('../../../lib/db.coffee')
 const { onPublish, generateSlugs, generateKeywords,
   sanitizeAndSave, onUnpublish } = require('./save.coffee')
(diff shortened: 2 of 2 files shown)


src/api/apps/articles/model/index.js
```diff
@@ -289,6 +294,18 @@ export const backfill = (callback) => {
...
       */
+
+      const url = getArticleUrl(article)
+
+      const deleteCb = () => {
+        console.log('article deleted')
+      }
+
+      // remove existing article from sailthru
+      deleteArticleFromSailthru(url, deleteCb)
```
kanaabe: Do we need this? Isn't this covered in `cleanArticlesFromSailthru` already in distribute?

package.json
```diff
@@ -13,7 +13,7 @@
     "assets": "sh scripts/assets.sh",
     "deploy": "sh scripts/deploy.sh",
     "mocha": "sh scripts/mocha.sh",
-    "task": "node -r coffee-script/register -r babel-core/register -r dotenv/config"
+    "task": "node -r coffee-script/register -r @babel/register -r dotenv/config"
```
kanaabe: Thanks! Def follow up and make a PR to include this in master too!

src/api/apps/articles/model/index.js
eessex: I believe cleanArticlesFromSailthru deletes all but the last article.  Added this on the assumption that we want to delete all existing articles and resend with new args.

kanaabe: Oh is this to cover the case of /video and /series? If we delete the article entirely, i wonder if it would remove analytics on it too? It might be worth checking the docs

kanaabe: Usually you can update an article by passing in the same url and giving it new args

eessex: I'm not very familiar with analytics related to email, which docs do you mean?  Is this internal artsy or Sailthru specific?

PR #1679, opened 2018-04-10
@kanaabe => Use full url for deleteArticleFromSailthru

eessex:
Missed one last spot that should use the full slug for sailthru.
Having some trouble with rewire as I try to test this, I wonder if we could pair on that?

### DIFF
--- a/src/api/apps/articles/model/index.js
+++ b/src/api/apps/articles/model/index.js
@@ -13,7 +13,7 @@ import moment from 'moment'
 const db = require('../../../lib/db.coffee')
 const { onPublish, generateSlugs, generateKeywords,
   sanitizeAndSave, onUnpublish } = require('./save.coffee')
-const { removeFromSearch, deleteArticleFromSailthru } = require('./distribute.coffee')
+const { removeFromSearch, deleteArticleFromSailthru, getArticleUrl } = require('./distribute.coffee')
 
 //
 // Retrieval
@@ -204,7 +204,9 @@ export const destroy = (id, callback) => {
     if (!article) {
       return callback(new Error('Article not found.'))
     }
-    deleteArticleFromSailthru(_.last(article.slugs), () => {
+
+    const url = getArticleUrl(article)
+    deleteArticleFromSailthru(url, () => {
       db.articles.remove({ _id: ObjectId(id) }, (err, res) => {
         if (err) {
           return callback(err)


kanaabe:
Would be great to see what code you're working with so far for the test. Push it here so I can see it maybe?

eessex:
Just pushed up the in progress tests -- it looks like it might be a conflict between coffee/es6

screen shot 2018-04-10 at 5 08 17 pm

eessex:
Requiring a second version of the Article `index.js` for rewiring seems to have solved compile issues locally.

src/api/apps/articles/test/model/index/index.test.coffee
```diff
@@ -1,10 +1,12 @@
 _ = require 'underscore'
+rewire = require 'rewire'
 sinon = require 'sinon'
 { db, fabricate, empty, fixtures } = require '../../../../../test/helpers/db'
 gravity = require('antigravity').server
 express = require 'express'
 app = require('express')()
 Article = require '../../../model/index.js'
+RewiredArticle = rewire '../../../model/index.js'
```
kanaabe: Hm this seems a little odd to do. It's worth figuring out why we need to do this instead of putting in confusing code that doesn't have context. At the least, we should figure out why we need to rewire separately and add a comment.

eessex: This reminds me of the cases of Force when things are exported twice -- do you know why we need do this here?
https://github.com/artsy/force/blob/9f74e9b0a228b4bd9e5701878c89b8beeecd7592/src/desktop/apps/article/components/InfiniteScrollArticle.js#L14

eessex: cc @damassi

kanaabe: Not sure exactly why, but this could help: https://github.com/artsy/force/pull/2282#discussion_r176533644

eessex: That matches the version of `babel-plugin-rewire` that we are currently using.

damassi: One thing to note is that babel plugin rewire is now Babel 7 ready and is already [set up in MP](https://github.com/artsy/metaphysics/blob/master/.babelrc#L19) -- also [here](https://github.com/artsy/metaphysics/blob/master/package.json#L110). I know this doesn't apply to CoffeeScript files, but just FYI. 

In Force, the reason why this was necessary is due to ES6 and how exports are static `const`s, which can't be "rewired" or rewritten on the fly. In CoffeeScript -- which is built in node on `module.export` (or CommonJS) this was possible, but not in modern JS. And so there was a little dance around having to work with ES6 and rewire in order to introduce Babel 7. I plan on going back through all of those FIXME's and bring babel-plugin-rewire back so as to remove all of that noise.

eessex: Looking at the docs here, https://github.com/speedskater/babel-plugin-rewire/blob/master/README.md#test-code-3 it seems like they are indeed importing two versions of the component - one with and one without rewire.  The syntax here has to be different because it is coffee, so seems like it does make sense for there to be two imports.

kanaabe: Hmm I don't follow - it looks like they're importing two versions to illustrate how you can rewire default and named exports. I believe they're trying to show how the automatic rewire is applied to the default as well - in this case we are manually rewiring everything.

eessex: Ok, do you mind pairing with me on this tomorrow?  I have spent a few hours debugging this test which seems like way to much for a small bugfix.  It is not completely clear to me what the blocking issue is.

kanaabe: Agree, happy to 👍. Maybe put some time on the cal since tomorrow will be pretty busy.

PR #1685, opened 2018-04-16
@kanaabe => Edit Series / onChangeArticle updates

eessex:
Took this opportunity for a small refactor to the `onChangeArticle` redux action so that it can accept objects (similar to backbone's set), and update nested object fields. 
 
- Adds ability to edit a custom `series.sub_title`
- Uses `series.description` field for editing rather than `series_description`
- Updates reaction
- Adds `onChangeArticleData` action to parse different kinds of args passed to `onChangeArticle`

TODO: Backfill existing `series_description` fields to new `series` object before deploying

### DIFF
--- a/src/client/actions/editActions.js
+++ b/src/client/actions/editActions.js
@@ -129,7 +129,7 @@ export const newHeroSection = (type) => {
   const section = setupSection(type)
 
   return (dispatch, getState) => {
-    dispatch(changeArticle('hero_section', section))
+    dispatch(changeArticleData('hero_section', section))
   }
 }
 
@@ -148,7 +148,7 @@ export const onChangeArticle = (key, value) => {
       edit: { article }
     } = getState()
 
-    dispatch(changeArticle(key, value))
+    dispatch(changeArticleData(key, value))
 
     debouncedUpdateDispatch(dispatch, { channel, article: article.id })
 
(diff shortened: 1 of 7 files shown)


src/client/actions/editActions.js
```diff
@@ -158,13 +158,36 @@ export const onChangeArticle = (key, value) => {
...
+    const { edit: { article } } = getState()
+    let data = {}
+
+    if (typeof key === 'object') {
+    // extend article with an object of keys
+      data = key
+    } else if (key.split('.').length > 1) {
+    // change a nested object value
+      const parentKey = key.split('.')[0]
+      const childKey = key.split('.')[1]
```
kanaabe: Minor improvement:

```
const keys = key.split('.')
const parentKey = keys[0]
const childKey = keys[1]
```

src/client/actions/editActions.js
```diff
@@ -158,13 +158,36 @@ export const onChangeArticle = (key, value) => {
...
+      const newObject = clone(article[parentKey]) || {}
+
+      newObject[childKey] = value
+      data[parentKey] = newObject
+    } else {
+    // change a single key's value
+      data[key] = value
```
kanaabe: Very into this! Following that Backbone set pattern to make it an optional shorthand is nice.

src/client/actions/editActions.js
damassi: Could also destructure right inline: 

```javascript
const [parentKey, childKey] = key.split('.')
```

PR #1686, opened 2018-04-17
@l2succes => Canvas-based Thumbnail Image

kanaabe:
This uses the Canvas API to "draw" some custom text to a canvas, generate an image out of it, converts it to a Blob type, and uploads it to S3 using `gemup`.

The screenshots below show a mismatch in cropping, but that's actually intentional. This format is optimized for social which is seen in the social tab as I scroll down. 

Generates Image:
canvas-image-generator

Can use scheduled publish date:
canvas-image-generator-2

### DIFF
--- a/src/client/components/image_generator/index.tsx
+++ b/src/client/components/image_generator/index.tsx
@@ -16,12 +16,8 @@ interface Props {
 }
 
 export class ImageGenerator extends Component<Props, State> {
-  constructor(props) {
-    super(props)
-
-    this.state = {
-      text: props.article.thumbnail_title || ""
-    }
+  state = {
+    text: this.props.article.thumbnail_title || ""
   }
 
   onChange = (text) => {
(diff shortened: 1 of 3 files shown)


orta:
Cool 👍

l2succes:
Nice! Is the plan to use the article image for the background or it's meant to remain a black?

src/client/components/image_generator/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+interface Props {
+  article: any
+  onChangeArticleAction: any
+}
+
+export class ImageGenerator extends Component<Props, State> {
+  constructor(props) {
+    super(props)
+
+    this.state = {
```
l2succes: you can also do 

```typescript
export class ImageGenerator extends Component<Props, State> {
  state = {
      text: this.props.article.thumbnail_title || ""
  }
...

kanaabe:
@l2succes We got new designs and they're all black background!

src/client/components/image_generator/index.tsx
kanaabe: 👍 Good catch

kanaabe:
@l2succes any more comments on this one? They wanted to start QA'ing by EOD.

l2succes:
@kanaabe Nope! Looks good!

PR #1690, opened 2018-04-19
@xtina-starr => Article List Layout Tag

kanaabe:
Signed-off-by: Christina <christina@artsymail.com>

### DIFF
--- a/src/client/apps/articles_list/components/articles_list.jsx
+++ b/src/client/apps/articles_list/components/articles_list.jsx
@@ -84,7 +84,6 @@ export class ArticlesList extends Component {
   state = {
     articles: this.props.articles || [],
     published: this.props.published,
-    layout: 'feature',
     offset: 0
   }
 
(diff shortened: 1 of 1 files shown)


src/client/apps/articles_list/components/articles_list.jsx
```diff
@@ -99,16 +100,22 @@ export class ArticlesList extends Component {
     // TODO: remove jQuery
     if ($('.filter-search__input').val()) { return }
```
eessex: re: the loading spinner TODO, we have an action for this in Redux -- I believe it is scoped to the edit page by selector, but could probably be adjusted to work here: https://github.com/artsy/positron/blob/master/src/client/actions/editActions.js#L338

eessex: It actually uses jquery too but could be refactored pretty easily into react.

kanaabe:
Ok @xtina-starr I removed the filter work to just include the layout tag!

PR #1693, opened 2018-04-24
@l2succes => Upgrade a few scripts used in Heroku Scheduler

kanaabe:
There are three scripts we use in Heroku Scheduler -- 

- Daily Upload to S3: uploads some fields to a CSV in S3 which Fulcrum then consumes
- Scheduler: finds articles that are scheduled (those with a scheduled_publish_at field) and publishes them
- Unqueue: resets the `daily_email` and `weekly_email` fields daily

This PR modernizes those scripts and adds Sentry reporting so we know when they break next time! Note that this requires an update to some fields here: https://scheduler.heroku.com/dashboard


l2succes:
Sorry! Looks good

PR #1696, opened 2018-04-26
@eessex => Remove unused scripts

kanaabe:
We have a bunch of dead code in the scripts folder that can be deleted -- I kept the reindexing one since I wasn't sure if it was still being used by @cavvia


kanaabe:
Merging this one @eessex!

PR #1697, opened 2018-04-26
@damassi => Move coffee-loader to dep

kanaabe:
I wasn't sure if this was the optimal move, but we're seeing a problem in production where scripts aren't being executed with Heroku Scheduler because it can't find `coffee-loader` which is a dev dependency.

When I bash into the dyno on prod and run the script manually:
```
~ $ yarn task scripts/scheduled_posts.js
yarn run v1.6.0
$ node -r coffee-script/register -r @babel/register -r dotenv/config scripts/scheduled_posts.js
module.js:491
    throw err;
    ^

Error: Cannot find module 'coffee-script/register'
    at Function.Module._resolveFilename (module.js:489:15)
    at Function.Module._load (module.js:439:25)
    at Module.require (module.js:517:17)
    at Function.Module._preloadModules (module.js:669:12)
    at preloadModules (bootstrap_node.js:461:38)
    at startup (bootstrap_node.js:157:9)
    at bootstrap_node.js:598:3
error Command failed with exit code 1.
```

vs Staging which works fine --

```
(master) ~/Development/Artsy/positron ♥ heroku run bash --app=positron-staging
Running bash on ⬢ positron-staging... up, run.3527 (Standard-1X)
~ $ yarn task scripts/scheduled_posts.js
yarn run v1.6.0
$ node -r coffee-script/register -r @babel/register -r dotenv/config scripts/scheduled_posts.js
Completed Scheduling 0 articles.
Done in 4.86s.
```


package.json
```diff
@@ -58,6 +58,7 @@
     "body-parser": "^1.14.1",
     "bucket-assets": "^1.0.2",
     "cheerio": "^0.19.0",
+    "coffee-loader": "^0.9.0",
     "coffeescript": "1.11.1",
```
damassi: The issue is that `coffee-script` isn't in the dependency list -- since they renamed the package there might be some mismatch. `yarn add coffee-script@1.11.1` should fix things.

package.json
```diff
@@ -58,6 +58,7 @@
     "body-parser": "^1.14.1",
     "bucket-assets": "^1.0.2",
     "cheerio": "^0.19.0",
+    "coffee-loader": "^0.9.0",
```
damassi: `coffee-loader` is likely pulling in `coffee-script` as a dep, and that's why it works -- but strictly speaking `coffee-loader` is only for webpack (and thus a client-side compile step, during development)

package.json
kanaabe: Ohh I see. In that case I think updating the task to 

`"task": "node -r coffeescript/register -r @babel/register -r dotenv/config"`

should do it?

damassi: Yup! i think that should work

kanaabe: Cool, thanks!

PR #1700, opened 2018-05-07
@l2succes => Return default for recognized key commands

eessex:
Fixes [GROW-243](https://artsyproduct.atlassian.net/browse/GROW-243)

if/else statement was stopping early for commands with modifier key, updates statement to return default key binding unless key command is 'link'.


eessex:
@l2succes bumping

PR #1705, opened 2018-05-14
Remove reporting script

wrgoldstein:
As part of ongoing efforts to centralize data processing code into one location (see e.g. https://github.com/artsy/fulcrum/pull/781), let's replace this script with something equivalent in Fulcrum. The other side of https://github.com/artsy/fulcrum/pull/917.

Note the Heroku scheduler task should be dropped if/when this is merged.

Thanks!


eessex:
This looks good to me, will check out the heroku task. Are there any other steps that should be taken before putting this in production?  Do I need to wait on Fulcrum deploys?

wrgoldstein:
Nothing else! Should be good to go. I'll remove the `"public"."positron_articles"` table in favor of the new `"positron"."articles"` one in Redshift.

eessex:
Cool, I'll give you a heads up when we deploy 🚀

Issue #1576, opened 2018-02-08
Title not saving in drafts unless you also add in Display


sarahdotson:
If you make a draft and add a title, it won't save the title and will appear in the drafts page as "Missing Title." You have to go and add it in the display panel under the "Magazine Headline" if you want the title to appear in the drafts page. 

I added this:
<img width="819" alt="screen shot 2018-02-08 at 11 42 55 am" src="https://user-images.githubusercontent.com/31517643/35985943-c8ea28e0-0cc5-11e8-8738-46beceacc9db.png">

And this is what it looks like in the draft page if you don't also put it in the Magazine Headline field:
<img width="563" alt="screen shot 2018-02-08 at 11 43 08 am" src="https://user-images.githubusercontent.com/31517643/35985959-d4fa4a84-0cc5-11e8-97f7-09e3012f4f9d.png">

eessex:
@sarahdotson this is technically not a bug -- the list view is built to show the magazine headline rather than the article title.  I believe this might be as an additional indicator that the field is not completed. Perhaps there is a way to change the UI so this is more obvious, cc @owendodd

PR #1709, opened 2018-05-29
Fix title ambiguity

meitros:
Explicitly use Page Title and Magazine Title in placeholders. This should make things more intuitive, e.g. when an article list says Missing Title and the page title was actually set.

Also correctly vertical align the missing thumbnail text

Issue #1576

### DIFF
--- a/src/client/apps/articles_list/components/articles_list.jsx
+++ b/src/client/apps/articles_list/components/articles_list.jsx
@@ -157,7 +157,7 @@ export class ArticlesList extends Component {
   showArticlesList () {
     const { channel, apiURL, checkable, selected } = this.props
     const isArtsyChannel = (type) => {
-      return type in ['editorial', 'support', 'team']
+      return ['editorial', 'support', 'team'].indexOf(type) >= 0;
     }
 
     if (this.props.articles && this.props.articles.length) {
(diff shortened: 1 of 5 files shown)


meitros:
This fixes the linked issue but may still be incomplete because there are other places where we ambiguously say "Missing Title" (hard to fix globally since I don't know how everything is used)

Remaining ambiguities from `git grep Missing`
src/client/apps/settings/client/curations/venice/venice_admin.coffee:          section: s.title || 'Missing Title'
src/client/components/drop_down/drop_down_item.jsx:            {title || 'Missing Title'}
src/client/components/paginated_list/index.jade:          .paginated-list-missing-img Missing Thumbnail
src/client/components/paginated_list/index.jade:            h1.paginated-list-missing-title Missing Title

kanaabe:
Ah yeah, the right solution might be to accommodate for both since Partners don't have the same subtabs. This is what they see: 

image

We can change the naming to be `Missing Thumbnail Title` and `Missing Thumbnail Image` if the current channel is a `partner` type. Otherwise show the `Magazine` version you have. Does that make sense?

meitros:
Sorry for the delay, updated!

src/client/apps/articles_list/components/articles_list.jsx
```diff
@@ -157,7 +157,7 @@ export class ArticlesList extends Component {
   showArticlesList () {
     const { channel, apiURL, checkable, selected } = this.props
     const isArtsyChannel = (type) => {
-      return type in ['editorial', 'support', 'team']
```
meitros: Coffeescript only :P

kanaabe:
LGTM! Thanks for the contribution @meitros!

PR #1710, opened 2018-05-29
@eessex => Allow refreshes on changes to partner_ids 

kanaabe:
In `fetchFromAccessToken`, we previously returned the user if we have one already, but this is too limiting for users who didn't used to have access to Writer but now do. This checks for differences between what's saved in the database and the JWT and allows for a user's info to be refreshed if it has changed.

### DIFF
--- a/src/api/apps/users/model.coffee
+++ b/src/api/apps/users/model.coffee
@@ -26,21 +26,17 @@ jwtDecode = require 'jwt-decode'
     db.users.findOne { access_token: encryptedAccessToken }, (err, user) ->
       return callback err if err
       if user
-        savedPartnerIds = user.partner_ids #TO STRING THIS TO COMPARE
+        savedPartnerIds = user.partner_ids.map((a) => a.toString())
         latestPartnerIds = jwtDecode(accessToken)?.partner_ids or []
-        console.log(jwtDecode(accessToken))
-        console.log(savedPartnerIds)
-        console.log(latestPartnerIds)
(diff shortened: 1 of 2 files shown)


src/api/apps/users/model.coffee
```diff
@@ -25,16 +25,18 @@ jwtDecode = require 'jwt-decode'
     return callback err if err
     db.users.findOne { access_token: encryptedAccessToken }, (err, user) ->
       return callback err if err
-      return callback null, user if user
+      if user
+        savedPartnerIds = user.partner_ids.map((a) => a.toString())
+        latestPartnerIds = jwtDecode(accessToken)?.partner_ids or []
+        return callback null, user if _.isEqual(savedPartnerIds, latestPartnerIds)
       # Otherwise fetch data from Gravity and flatten it into a Positron user
-      async.parallel [
```
kanaabe: This async array is useless so I've removed it in favor of normal callbacks.

src/api/apps/users/model.coffee
```diff
@@ -25,16 +25,18 @@ jwtDecode = require 'jwt-decode'
     return callback err if err
     db.users.findOne { access_token: encryptedAccessToken }, (err, user) ->
       return callback err if err
-      return callback null, user if user
+      if user
+        savedPartnerIds = user.partner_ids.map((a) => a.toString())
```
eessex: A comment here on the difference between the saved and latest might be helpful for future context.

PR #1714, opened 2018-06-05
@l2succes => Cleanup /editActions

eessex:
The `editActions` file was super long and hard to parse, as were related tests. 
This PR has no new code, it breaks out existing methods and tests from `editActions` into a subdirectory (`actions/edit`), with files for the following concerns:
- articleActions
- errorActions
- sectionActions
There is also an index file that exports the key-mirror from each file as a single `action` const for the `editReducer`


src/client/actions/edit/articleActions.js
```diff
@@ -0,0 +1,211 @@
+import { clone, cloneDeep, debounce, set } from 'lodash'
```
damassi: IMO this is exactly the right path -- take an actions file and when it gets too large group it under an appropriate subfolder 💯. The same process typically happens at the reducer level as well, so that things maintain a 1:1 pairity.

src/client/actions/edit/index.js
```diff
@@ -0,0 +1,13 @@
...
+import { actions as editActions } from './editActions'
+import { actions as errorActions } from './errorActions'
+import { actions as sectionActions } from './sectionActions'
+
+export const actions = extend({},
+  articleActions,
+  editActions,
+  errorActions,
+  sectionActions
+)
```
damassi: Can just do 

```
export const actions = {
   ... 
}
```
since there's no need to merge with another object (only using `{}`). Using `extend` will create an intermediate object in memory and copy over by reference, creating a bit of indirection.

eessex:
@damassi do you mean that the reducer should also be broken out into a subdirectory?  This sounds ok to me and can do in another PR -- will update the `extend` comment now.

damassi:
@eessex yup -- typically there's something like `reducers/someReducer` and `actions/someActions` where `some` is the category and they match. That way, just by looking at the filename from where the action comes from, one can know exactly what corresponding file the state is being updated in.

PR #1716, opened 2018-06-13
@kanaabe => Debounce PlainText change, fix tab-toggle

eessex:
Taking a stab at fixing up our plainText saving issues -- scrapped the redux state idea for a simpler solution, debouncing changes to the redux state from the text component.  I believe the culprit of this bug is a [race condition](https://draftjs.org/docs/advanced-topics-editorstate-race-conditions.html) between the local editorState and the redux action `onChangeArticle`.

This also fixes a bug introduced in [this PR](https://github.com/artsy/positron/pull/1714), where `changeView` was not being imported correctly. It also removes the use of `bindActionCreators` in favor of explicitly defining each action.

### DIFF
--- a/src/client/apps/edit/components/header/index.jsx
+++ b/src/client/apps/edit/components/header/index.jsx
@@ -1,21 +1,24 @@
 import styled from 'styled-components'
 import PropTypes from 'prop-types'
 import React, { Component } from 'react'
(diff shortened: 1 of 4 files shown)


kanaabe:
Nice clean up! I like the debounce experiment.

PR #1719, opened 2018-06-26
@damassi => Bump Reaction

eessex:
- Adds support for consuming `.mjs` files
- Upgrades Typescript to v2.9.1
- Matches Relay-related resolutions to Force
- Bumps reaction to latest

### DIFF
--- a/src/client/apps/edit/components/content/sections/news/EditSourceControls.tsx
+++ b/src/client/apps/edit/components/content/sections/news/EditSourceControls.tsx
@@ -1,6 +1,6 @@
 import React, { Component } from 'react'
 import styled from 'styled-components'
-import { borderedInput } from '@artsy/reaction/dist/Components/Mixins'
+// import { borderedInput } from '@artsy/reaction/dist/Components/Mixins'
 import Button from "@artsy/reaction/dist/Components/Buttons/Default"
 
 interface Props {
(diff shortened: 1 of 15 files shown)


src/client/apps/edit/components/content/sections/news/EditSourceControls.tsx
```diff
@@ -57,28 +59,28 @@ export class EditSourceControls extends Component<Props, State> {
...
   margin-bottom: 10px;
+  background: white;
 `
 
 const LinkInput = Input.extend`
   margin-right: 0px;
 `
 
-const ApplyButton = styled(Button)`
+const ApplyButton = Button.extend.attrs<{onClick: any}>({})`
```
damassi: Did this fix all of those typescript errors?

damassi:
LGTM 👍 

danger, merge on green

eessex:
Ugh, looks like more relay errors in Circle -- `Unexpected token: name (RelayServerSSR) [./node_modules/react-relay-network-modern-ssr/lib/server.js:24,0][main.js:217777,6]` 

Headed out but will bump reaction and check this issue out in the am.

src/client/assets/main.coffee
```diff
@@ -3,6 +3,7 @@
 # can break this up into self-initializing app-level asset packages like Force.
 #
 
+require('regenerator-runtime/runtime')
```
damassi: I think regenerator is only needed here -- that or only in `src/index.js`. Can remove in other places.

eessex: Did some experiments moving/removing, but we definitely need it here otherwise it fails to load the bundle and throws errors for missing regeneratorRuntime in the browser console.

damassi:
LGTM 👍 

Danger, merge on green

PR #1721, opened 2018-07-02
@kanaabe => Add circle 2.0 config.yml

eessex:
Adds support for CircleCI 2.0, removes 1.0 config.


kanaabe:
Thanks for being ahead of the curve!

PR #1727, opened 2018-07-11
Update caching strategy

zephraph:
This does two things:

1. Provides a fallback caching strategy for `node_modules`. This means that if the `yarn.lock file` changes it won't have cache but it'll use the last good cache. It'll still have to install/remove some dependencies, but at least it won't have to install all of them
2. Updates the yarn install command to use the `--check-files` flag. This flag forces yarn to look at the actual contents of the `node_modules` directory. It's necessary for the cache fallback strategy. 

Also side note, `yarn` and `yarn install` are equivalent.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,12 +11,13 @@ jobs:
     steps:
       - checkout
       - restore_cache:
-          name: Restore yarn cache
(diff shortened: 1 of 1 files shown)


eessex:
Looks good, thanks for adding!

PR #1730, opened 2018-07-31
[Toolchain] Update to Typescript 3

damassi:
Updates to Typescript 3.x and latest Babel. Additionally brings the app in line with our other JS apps by adding 

- https://github.com/okonet/lint-staged
- Default VSCode settings
- TSLint


eessex:
@damassi the test that is failing here has been fixed, want to pull the latest master and try rebuilding?

damassi:
@eessex - will do first thing tomorrow 👍

eessex:
Hm, maybe that test is not fixed -- i think it is ok to x that one problematic test for the moment, would like to get these changes in, can keep investigating that separately.

tslint.json
```diff
@@ -0,0 +1,20 @@
...
+    "member-ordering": false,
+    "no-unused-expression": false,
+    "no-var-requires": false,
+    "object-literal-sort-keys": false,
+    "ordered-imports": true,
+    "no-namespace": [true, "allow-declarations"],
+    "trailing-comma": false,
+    "variable-name": [true, "ban-keywords", "allow-leading-underscore"]
+  }
+}
```
damassi: Copied from Reaction

package.json
```diff
@@ -6,39 +6,48 @@
...
-    "@babel/preset-react": "7.0.0-beta.42",
-    "@babel/preset-stage-3": "7.0.0-beta.42",
-    "@babel/preset-typescript": "7.0.0-beta.42",
-    "@babel/register": "7.0.0-beta.42",
-    "@babel/runtime": "7.0.0-beta.42",
+    "@babel/cli": "7.0.0-beta.54",
+    "@babel/core": "7.0.0-beta.54",
+    "@babel/node": "7.0.0-beta.54",
+    "@babel/plugin-proposal-class-properties": "7.0.0-beta.54",
+    "@babel/plugin-proposal-decorators": "7.0.0-beta.42",
```
eessex: Why is plugin-proposal-decorators not bumped as well?  have a version running locally that seems ok on beta.54

damassi: I _think_ this was deliberate because tests would fail but let me bump and try again.

eessex: This should work if you add `["@babel/plugin-proposal-decorators", {"legacy": true}],` to `.babelrc`

eessex: I think it also has to come first in the plugins array, or maybe just before `plugin-proposal-class-properties`

PR #1732, opened 2018-08-16
@xtina-starr => Move PlainText to components/draft

eessex:
We've been storing our draft components in a folder called `rich_text`, but we are also including other kinds of fields like `PlainText`.  As part of text component work, this PR starts a new component directory `components/draft` as a new home for any `draft-js` components.

- Moves existing `PlainText` component from `rich_text` to `draft` directory
- Updates imports for `PlainText` component
- Minor cleanup / please linter in touched files


eessex:
Yeah have no idea why this is failing -- looks like it is hanging before starting jest, which is something i've experienced in circle before -- runs green locally.

PR #1733, opened 2018-08-16
Paragraph utils

eessex:
The current utils under `rich_text` are mostly oriented towards the article's `Text` component, adding a lot of complexity that makes them hard to parse.  For the new `draft-js` paragraph I'm removing the article context and focusing instead on strict rules around what html blocks and styles are allowed. To keep the actual components as short as possible, much of the logic is stored in utils files.

This PR breaks out logic for the paragraph component's helper functions into a series of files by use-case:
- convert (html <=> draftjs)
- links (logic related to creating/removing links)
- decorators (logic to setup editor for allowing/rendering links)
- utils (kind of random, some setup functions and key commands)

Tests have been broken out to be more explicit -- especially `convert.test.js`, which now tests each block/style type individually rather than passing in a big chunk of article html.

The actual component will follow after this is merged to keep PRs shorter for review.


eessex:
I believe the test failure has to do with Circle config rather than the code here, this is ready to review -- will explore 🔴

l2succes:
@eessex Does that mean there's a sibling PR that removes a lot of this complex logic from the `Text` component?

eessex:
The immediate work deals specifically with the `Paragraph` component, but the `Text` component will be reworked as well in this fashion.  Moved to the new `draft` components folder to avoid breaking anything in `Text` until that component is addressed.

src/client/components/draft/paragraph/utils/convert.js
```diff
@@ -0,0 +1,152 @@
...
+  'CODE',
+  'ITALIC',
+  'STRIKETHROUGH',
+  'UNDERLINE'
+]
+
+export const convertHtmlToDraft = (html, hasLinks, allowedStyles) => {
+  /**
+   * Convert HTML to Draft ContentState
+   */
```
damassi: For these doc-blocks to be parsed correctly by tooling they need to idiomatically exist on the outside of the function:

```js
/**
 * Convert HTML to Draft ContentState
 */
export const convertHtmlToDraft = (html, hasLinks, allowedStyles) => {
```

eessex: They seem to be working fine locally, what kind of parsing errors would you expect? Is it possible this is why circle is failing?

eessex: still interested to know why this errors, don't see any obvious flags from a quick google

damassi: It's not an error in the sense that the code will break; it's that editor tooling follows idiomatic patterns of use based on a spec. Since docblocking is top down (documentation, then code) and thus meant to be placed on the outside of the function, when placed on the inside the assigned scope becomes invalid (in this case appearing for `let cleanedHtml = stripGoogleStyles(html)`) 

You can test this by placing your cursor above the function and typing `/**` and hitting enter. It will fill in placeholders for variables and params. If on the inside the parser no longer knows how to capture context, and tooling that reads dockblocks will link to an incorrect line number.

l2succes: Good point @damassi. Yeah @eessex we might as well follow the JSDoc format for documenting functions. That way it gets picked up by Typescript and consequently VSCode.

See: http://usejsdoc.org/about-getting-started.html
https://github.com/artsy/reaction/blob/master/src/Components/Authentication/Types.ts#L35

damassi: Visually I do like the inner Python-style comment-blocks as it looks a bit cleaner but that format isn't really seen in JSland.

One last thought is that all of these files could be `.ts` since positron now supports TypeScript. Not suggesting that everything needs to be properly typed out with interfaces, but basic out-of-the-box checks are still safer than `.js`.

src/client/components/draft/paragraph/utils/__tests__/utils.test.js
```diff
@@ -8,9 +8,9 @@ import {
   styleNamesFromMap,
   styleNodesFromMap
 } from '../utils'
-const SelectionUtils = require('../../../../rich_text/utils/text_selection.js')
+const SelectionUtils = require('../../../../rich_text/utils/text_selection')
```
damassi: Can use absolute paths: https://github.com/artsy/positron/blob/master/.babelrc#L20

src/client/components/draft/paragraph/utils/convert.tsx
```diff
@@ -1,7 +1,14 @@
...
-import { stripGoogleStyles } from 'client/components/rich_text/utils/text_stripping'
+import {
+  ContentState,
+  DraftEntityMutability,
+  RawDraftEntity,
+  RawDraftContentBlock
+} from 'draft-js'
+import { stripGoogleStyles } from '../../../rich_text/utils/text_stripping'
 import { styleNamesFromMap, styleNodesFromMap } from './utils'
+import { StyleMap, StyleMapNames, StyleName } from './typings'
```
damassi: 👍

src/client/components/draft/paragraph/utils/__tests__/utils.test.js
eessex: The absolute paths seem to be broken in typescript

PR #1734, opened 2018-08-23
@damassi => Install prettier, update vscode settings

eessex:
Noticed that prettier wasn't working -- but then saw it wasn't installed!  Adds the library and fix on save for typescript.


damassi:
👍

PR #1735, opened 2018-08-23
@damassi => API tests can use Jest

eessex:
Some setup for removing `babel-rewire` plugin:
- Renames `/api` mocha tests to `*.spec.js` extension
- Update mocha calls to look for `*.spec.js` but not `*.test.js`
- Sets up Jest to look for `*.test.js` from root (don't exclude API)
- Renames `/__tests__` and `/tests` directories to `/test`
- Converts `src/api/apps/search/test/routes.test.js` to jest and remove `__Rewire__`
- Fix a few `prettier` complaints

This fixes one out of the three files using `babel-rewire`, have confirmed that the same number of tests are being checked by jest.


eessex:
Yeah I would definitely like to rename everything to `__test__` -- however that will be a very busy PR, so while I agree with your point, I think we should address that in a separate PR where we can search and replace all instances.

damassi:
Sounds good 👍

PR #1739, opened 2018-08-31
[HOLD] @l2succes => Bump reaction

eessex:
HOLD for https://github.com/artsy/reaction/pull/1209

- Updates `reaction` to most recent
- Updates `palette` to most recent
- Bumps `react` and `react-dom` to 16.3 (for context api changes)
- Updates article headers to use edit props rather than children (introduced in `reaction` 2.9.0)
- Removes dependency on Publishing index (to prevent importing files that use relay/context)


eessex:
Need to roll back this one -- it is waiting for artsy/reaction#1209 hence the [HOLD]

eessex:
Or if you're able to review the reaction PR can do another pr to bump again

PR #1742, opened 2018-09-12
Improve the initial setup experience

orta:
* Adds the OSS xapp creds
* Adds an explanation about the error when those creds are empty on launch
* Lets prettier do it's thing on that file


xtina-starr:
LGTM 👍

PR #1744, opened 2018-09-20
Bump reaction / Header QA

eessex:
- Bumps reaction up to most recent
- Minor QA to hide image upload from 'basic' header (has no image option)


eessex:
Merging since trivial

PR #1760, opened 2018-09-27
@xtina-starr => Section container

eessex:
Follows up https://github.com/artsy/reaction/pull/1278

- Updates `eslint` config to match new `prettier` rules
- Removes addition/removal of `section.layout` for `<blockquote>` text in `Text` component
- Updates `SectionContainer` and `SectionControls` to use new Reaction `getSectionWidth` to determine width
- Converts some css from above components to `styled` component

TODO: Backfill to remove existing `section.layout: "blockquote"`

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -15,6 +15,8 @@
   },
   "rules": {
     "camelcase": 0,
+    "comma-dangle": ["error", "never"],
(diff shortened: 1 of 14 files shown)


xtina-starr:
LGTM!

PR #1763, opened 2018-10-09
Typescript text controls

eessex:
- Converts `TextNav` and `TextUrlInput` to typescript
- Updates `TextNav` to determine allowed blocks based on draft js `blockmap`
- Moves popup position calculations to `TextNav` and `TextUrlInput` (out of `Paragraph`)
- Updates `Paragraph` component to use new `TextNav` and `TextUrlInput`
- Converts `Paragraph` tests to typescript
- Updates typings to accommodate full text component


src/client/components/draft/components/text_input_url.tsx
```diff
@@ -0,0 +1,190 @@
...
+          >
+            Apply
+          </Button>
+        </TextInputUrlContainer>
+      </>
+    )
+  }
+}
+
+const TextInputUrlContainer = styled.div.attrs<{ top: number; left: number }>(
```
l2succes: we can do this now
```typescript
styled.div<{ top: number; left: number }>``
````

https://github.com/styled-components/styled-components/issues/630#issuecomment-393638610

eessex: Are we using this in Reaction already?  I'm getting a weird error, `ReferenceError: number is not defined`.  Tried updating `typescript-styled-plugin` and typescript versions are matched to Reaction.

eessex: This requires a few toolchain updates, going to address in a separate PR

PR #1766, opened 2018-10-16
Setup /edit2 route, new actions for text

eessex:
First of a few PRs for new Text component

- Sets up access to route `edit2` for editorial team (currently same as `/edit`)
- Adds redux actions related section changes triggered from new text component

### DIFF
--- a/src/client/actions/edit/sectionActions.js
+++ b/src/client/actions/edit/sectionActions.js
@@ -73,13 +73,11 @@ export const onChangeSection = (key, value) => {
   }
 }
 
-export const onSplitTextSection = (sectionIndex, existingSectionBody, newSectionBody) => {
+export const onSplitTextSection = (existingSectionBody, newSectionBody) => {
   return (dispatch, getState) => {
-    const {
-      edit: { article },
-    } = getState()
-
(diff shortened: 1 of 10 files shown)


src/client/actions/edit/sectionActions.js
```diff
@@ -73,6 +73,67 @@ export const onChangeSection = (key, value) => {
...
+export const maybeMergeTextSections = () => {
+  return (dispatch, getState) => {
+    const {
+      edit: {
+        article: { sections },
+        section,
+        sectionIndex
+      },
+    } = getState()
+    const sectionBefore = sections[sectionIndex - 1]
```
l2succes: any chances `sections.length == 0`?

eessex: There is a check for this in the component where it is called from, but I think you're right that it should be called here, especially since its name includes "maybe"

eessex: Or a check for if sectionIndex is 0, rather -- because this is called from inside of a section, there will always be at least one item in the array.

l2succes:
merge on green

PR #1767, opened 2018-10-17
/Edit2 text utils updates

eessex:
- Adds ImmutableJs
- QA for /edit2 route
- Moves decorators, links, some utils, getSelectionDetails to`draft/shared`
- Updates typing to handle richText component
- Adds utils for RichText
- Sets up some tests as TODOs


package.json
```diff
@@ -81,6 +81,7 @@
     "glossary": "^0.1.1",
     "google-spreadsheet": "^2.0.3",
     "graphql": "^0.7.2",
+    "immutable": "^3.8.2",
```
yuki24: It doesn't seem like we benefit from Immutable enough to justify the dependency. I'd consider removing it if that wasn't a huge investment.

eessex: We need to use it for the `draft-js` library, in order to pass custom blocks to the editor -- there are a few more places it comes up in a follow-up to this pr.

src/client/components/draft/components/text_nav.tsx
```diff
@@ -228,8 +230,12 @@ export class TextNav extends Component<Props, State> {
...
 }
 
-const TextNavContainer = styled.div.attrs<{ top: number; left: number }>({})`
-  max-width: 250px;
+export const TextNavContainer = styled.div.attrs<{
+  top: number
+  left: number
+  width: number
+}>({})`
+  max-width: ${props => props.width}px;
```
xtina-starr: 👍

PR #1770, opened 2018-10-23
Setup Hokusai

izakp:
Steps to setup Positron:
- Ran `hokusai setup --template-remote git@github.com:artsy/artsy-hokusai-templates.git --template-dir nodejs`
- Got `hokusai build` working.  Pinned nodejs to version `8.12.0` as running on Heroku production - yarn to version `1.9.4` as built by CircleCI
- Got `hokusai dev start` running - not mapping a dynamic volume to `/app` at the moment, but we could test live reload in this configuration, as long as we don't sync `/node_modules`
- Got `hokusai test` working - removed `yarn assets` from `scripts/test.sh` as it's run during the Docker build
- Ran `hokusai staging env create` then populated env from Heroku config
- Ran `hokusai staging create` to launch the stack


eessex:
@izakp does the docker build update the `ASSET_MANIFEST` env var as well as pushing to s3?

izakp:
@eessex re:

> does the docker build update the ASSET_MANIFEST env var as well as pushing to s3?

I want to try out setting the `COMMIT_HASH` env var and loading the asset manifest from the CDN rather than `cat`ing it into the `ASSET_MANIFEST` env var... see https://github.com/artsy/bucket-assets#bucket-assets

Updated the CircleCi config to more or less preserve the existing workflow

hokusai/development.yml
```diff
@@ -0,0 +1,28 @@
...
+services:
+  positron:
+    command: ["yarn", "dev"]
+    extends:
+      file: build.yml
+      service: positron
+    ports:
+    - 3005:3005
+    environment:
+    - ELASTICSEARCH_URL=http://positron-elasticsearch:9200
```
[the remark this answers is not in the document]
izakp: Following up - increased list indentation here and opened https://github.com/artsy/artsy-hokusai-templates/pull/1

izakp:
Setting up staging on k8s...
- Pushed initial image at `bf73483dc9865aa0ff68d5626dd066df700f1efd` via `hokusai registry push`
- Set up staging environment vars with `hokusai staging env create`
- Copied over config from heroku `heroku config --app positron-staging` and recreated in k8s staging via `hokusai staging config set ...`
- Ran `hokusai staging env unset ASSET_MANIFEST COMMIT_HASH` to prep staging environment to test setting the COMMIT_HASH via CI and loading via the CDN on app startup

izakp:
@eessex staging is up and running at `abdfc4cfde76311e893cb1211419c762-1090982097.us-east-1.elb.amazonaws.com` (but redirects to `stagingwriter.artsy.net`)

Seeing some strange logs but the Kubernetes-probe check to `/health` appears to be working...
```
GET /health 302 2.443ms ::ffff:10.0.0.218 "kube-probe/1.10"
GET /unsupported 200 0.666ms ::ffff:10.0.0.218 "kube-probe/1.10"
GET /health 302 0.898ms ::ffff:10.0.0.218 "kube-probe/1.10"
GET /unsupported 200 0.735ms ::ffff:10.0.0.218 "kube-probe/1.10"
```

Also on bootstrap the app prints out its entire environment...
```
yarn run v1.9.4
$ sh scripts/start.sh
-e  WARNING: Missing .env file.
+ [ ! -f ./.env ]
+ echo -e \033[1;31m WARNING: Missing .env file. \033[0m
+ cat .env
+ grep NODE_ENV
+ xargs
cat: .env: No such file or directory
+ export
export ACCELERATION_WEB_PORT='tcp://100.66.112.202:443'
export ACCELERATION_WEB_PORT_443_TCP='tcp://100.66.112.202:443'
export ACCELERATION_WEB_PORT_443_TCP_ADDR='100.66.112.202'
export ACCELERATION_WEB_PORT_443_TCP_PORT='443'
```

Would prefer that it didn't do this as sensitive information will end up in Papertrail...  Is this is the `dotenv` module?

eessex:
@izakp yeah I'm guessing this comes from dotenv -- usually it would only print the system vars, are you definitely seeing all the keys?

izakp:
@eessex yes I am seeing all the keys in `hokusai staging config get` plus all the env vars Kubernetes automatically injects into each pod.  I think I'll modify `./scripts/start.sh` as we don't need forever or the dotenv config on Kubernetes (as it sets up the environment and restarts any crashed process)

izakp:
Left `./scripts/start.sh` as-is and overriding the default container args in the Kubernetes yaml spec w/o loading the dotenv module.

izakp:
Okay @eessex all looks good on the new staging deployment.  Is it okay to change DNS for `stagingwriter.artsy.net` to resolve to the new deployment at `abdfc4cfde76311e893cb1211419c762-1090982097.us-east-1.elb.amazonaws.com`?

It may be simpler to get this merged and the app migrated first, leave the existing Mongo deployments running on Heroku (just scale down the dynos) and then do the database migration after we also confirm the new CI build is working as expected.

izakp:
Ready to go @eessex - staging and production Kubernetes deployments are set up.  To roll out we can 1) merge this into master and let CI deploy staging, 2) make a release PR and let CI deploy production, 3) update DNS records to point to new deployments 

A few notes about the new deployment:
- Had to put Nginx in front of Positron to handle websocket connections - running as a sidecar container using the `artsy/docker-nginx` image - the ELB talks to it on ports 8080 / 8443 and it connections to Positron via port 3005 (confirmed websockets is working on staging)
- Baked the commit hash into the image in `COMMIT_HASH.txt`, which is then read by `./scripts/production.sh` and `./scripts/publish-assets.sh` meaning we no longer need to update this env var during deployment.  Similarly, `./scripts/production.sh` uses `curl` to fetch the assets manifest from the CDN and set the `ASSET_MANIFEST` environment variable - it's a workaround as the bucket-assets module _should_ do this but is broken AFAICT
- Refactored the asset pipeline and added the yarn command `publish-assets` that's run as a Hokusai `pre-deploy` hook (launches a new container with the new image and runs the command before deploying) so no need to inject aws credentials etc into the CI environment as it runs in the staging / production environment (also means that deploying / promoting locally will run this hook and always publish assets)

izakp:
Also went back to injecting the `COMMIT_HASH` as a required build argument like we do in Force, as when copying the `.git` repo into the image, even after running `rm -rf .git` the files are still persisted in an underlying layer within the image build, and have to be cleaned up with something like https://github.com/jwilder/docker-squash

To run anything that triggers a build, i.e. `hokusai [build|test|dev start|registry push] you have to inject COMMIT_HASH as is done in CI, i.e. `COMMIT_HASH=$(git rev-parse --short HEAD) hokusai test`

Note to self to consider adding pre-build and post-build hooks in Hokusai so we can put this extra step in hokusai's config.yml and not have to remember to do this all the time...

Dockerfile
```diff
@@ -1,6 +1,40 @@
-FROM node:10.13
-RUN mkdir -p /usr/src/app
-WORKDIR /usr/src/app
-COPY package.json yarn.lock ./
-RUN yarn --pure-lockfile
-COPY . .
+FROM node:10.13-alpine
+ARG COMMIT_HASH
+RUN test -n "$COMMIT_HASH"
```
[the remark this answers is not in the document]
izakp: > Does this mean that this arg will always be necessary

Yes

>  I'm wondering if it's possible to determine the commit hash automatically from here rather than expect the argument.

This was basically my attempt to copy in the entire `.git` repository, then remove it during the build, only to find that this necessitates squashing afterwards

So for now, we can't have it both ways.

However, made a simple POC to add a `pre-build` and `post-build` hooks to hokusai config though so these could be baked into `./hokusai/config.yml` like and run inline with the build like...
```
pre-build: export COMMIT_HASH=$(git rev-parse --short HEAD)
post-build: docker save hokusai_positron:latest | sudo docker-squash -t hokusai_positron | docker load
```

Then we could have it both ways!

.circleci/config.yml
```diff
@@ -1,84 +1,48 @@
...
+          command: hokusai configure --kubectl-version 1.10.7 --s3-bucket artsy-citadel --s3-key k8s/config --platform linux
+      - run:
+          name: Push
+          command: COMMIT_HASH=$(git rev-parse --short HEAD) hokusai registry push --tag $CIRCLE_SHA1 --force --overwrite
       - run:
           name: Deploy to production
-          command: git push --force https://heroku:$HEROKU_TOKEN@git.heroku.com/positron-production.git master
-
+          command: hokusai production deploy $CIRCLE_SHA1
 workflows:
```
eessex: Love how short this gets 👍

dleve123: @izakp When do you typically decide between a `hokusai pipeline promote` versus a direct deploy via `hokusai env deploy`? Are most CI-based deployments direct deploy?

Dockerfile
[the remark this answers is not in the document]
dleve123: I'd recommend documenting the dependency on the Git commit hash for hokusai based commands in the Positron's README

Dockerfile
```diff
@@ -1,6 +1,40 @@
...
+
+RUN apk add curl git
+
+# Set up deploy user and working directory
+RUN adduser -D -g '' deploy
+RUN mkdir -p /app
+RUN chown deploy:deploy /app
+
+# Set up dumb-init
```
dleve123: @izakp If I recall a previous conversation we had correctly, we'll want all/most Dockerfiles to contain this installation for dumb-init. I notice that we don't have a base node image in https://github.com/artsy/docker-images -- could it make sense to extract out the common-components of this image into a base image? Maybe could make future node imagines a bit more consistent.

.circleci/config.yml
izakp: @dleve123 it's pretty much up to the project maintainers whether they prefer to deploy production via a `hokusai pipeline promote` or a PR from say, `staging` -> `release` here then have CI deploy from `release`.  Promoting will ensure the exact same image is deployed to production, while the release-branch strategy creates an additional image build via a merge commit.  @joeyAghion and I tried to decide on on good / recommended solution but ran out of ideas when we realized we couldn't set up PRs that always rebase rather than merge, which would have provided a best-of-both-worlds solution.

izakp: There was also the additional concern that promotion could encounter a race condition if `master` was (unknown to the promoting developer) in the process of being deployed while they ran the `hokusai pipeline promote` command.

izakp:
Going to merge and run the new pipeline before updating DNS

PR #1771, opened 2018-10-24
Section tool QA

eessex:
Resolves https://artsyproduct.atlassian.net/browse/GROW-969

Bug surfaced during text QA -- redux's `setSection` action was being called every time a section was clicked on, expected behaviour is to call the method only when not editing. It was also being called when removing a section, causing a bug where the edit tool would not open after sections were deleted.

- Converts `SectionTool` to `styled-components`,  removes related unused icons
- In `SectionContainer`, moves `setSection` trigger from hover controls and show only when not editing


eessex:
bumping @xtina-starr @l2succes

PR #1772, opened 2018-10-29
@l2succes => Toolchain updates to match Reaction

eessex:
- Pins node to 8.12.0 
- Package updates to bring Babel up to 7.0.0
- Updates Reaction to latest


package.json
```diff
@@ -30,26 +30,15 @@
...
-    "@babel/core": "7.0.0-rc.2",
-    "@babel/node": "7.0.0-rc.2",
-    "@babel/plugin-proposal-class-properties": "7.0.0-rc.2",
-    "@babel/plugin-proposal-decorators": "7.0.0-rc.2",
-    "@babel/plugin-transform-runtime": "7.0.0-rc.2",
-    "@babel/preset-env": "7.0.0-rc.2",
-    "@babel/preset-react": "7.0.0-rc.2",
-    "@babel/preset-typescript": "7.0.0-rc.2",
-    "@babel/register": "7.0.0-rc.2",
-    "@babel/runtime": "7.0.0-rc.2",
```
damassi: If we're not fully compiling out our server-side CoffeeScript and other ES6 / TypeScript before sending to prod then moving these modules to `devDeps` will fail during deploy. (Things like `coffee-script/register` and `@babel/register` mean that code is transformed during runtime (and then cached).

.babelrc
```diff
@@ -3,7 +3,8 @@
     ["@babel/env", {
       "targets": {
         "browsers": ["last 2 versions"]
-      }
+      },
+      "useBuiltIns": "usage"
```
damassi: Can you do some investigating around `useBuiltIns: "usage"` and check to see if `@babel/plugin-transform-runtime` is still required? I _believe_ this supersedes it.

eessex: Yeah looks like it runs fine without `@babel/plugin-transform-runtime`

package.json
eessex: Are we not fully compiling before we deploy? I thought that was what the `yarn assets` step is for.

damassi: Thats only for the client-side

.babelrc
```diff
@@ -3,16 +3,17 @@
...
     }],
     "@babel/react",
     "@babel/preset-typescript"
   ],
   "plugins": [
     ["@babel/plugin-proposal-decorators", { "legacy": true }],
     "@babel/plugin-proposal-class-properties",
-    "@babel/plugin-transform-runtime",
     "inline-react-svg",
+    "lodash",
```
l2succes: I think it's safe to remove lodash here

package.json
eessex: Sounds good, put these back in `dependencies`

.babelrc
damassi: Even though `underscore` is the official utility lib it's being used all over (same in force).

damassi: I wonder if there's some kind of codemod we could run to swap it out

eessex: You mean remove underscore? I'd been encouraged to switch to `lodash` wherever possible.

damassi: Yup remaining underscore usage and replace with lodash using the codemod, would automate all the work away.

damassi: Thinking further though, that codemod won't work with coffeescript /shrug

PR #1788, opened 2018-11-19
Video credits allow linebreaks

eessex:
Updates `Paragraph` component to optionally support empty lines, and sets prop to true on `media.credits` field.

Addresses: https://artsyproduct.atlassian.net/browse/GROW-961

### DIFF
--- a/src/client/components/draft/paragraph/test/paragraph.test.tsx
+++ b/src/client/components/draft/paragraph/test/paragraph.test.tsx
@@ -158,6 +158,17 @@ describe("Paragraph", () => {
       expect(component.state().html).toBe("<p>A paragraph</p>")
     })
 
+    it("Preserves empty blocks if props.allowEmptyLines", () => {
+      props.allowEmptyLines = true
+      props.html = "<p>A paragraph</p><p></p><h1></h1><h2></h2>"
+      const component = getWrapper(props)
+      const instance = component.instance() as Paragraph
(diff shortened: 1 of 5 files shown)


src/client/components/draft/paragraph/test/paragraph.test.tsx
```diff
@@ -158,6 +158,17 @@ describe("Paragraph", () => {
...
 
+    it("Preserves empty blocks if props.allowEmptyLines", () => {
+      props.allowEmptyLines = true
+      props.html = "<p>A paragraph</p><p></p><h1></h1><h2></h2>"
+      const component = getWrapper(props)
+      const instance = component.instance() as Paragraph
+      const editorState = instance.editorStateFromHTML(component.props().html)
+      instance.onChange(editorState)
+
+      expect(component.state().html).toBe("<p>A paragraph</p><p><br /></p>")
```
xtina-starr: This PR looks good. For personal knowledge, I wanted to know where the `<p>` tag gets converted to `<br/>`?

eessex: The space is inserted here - https://github.com/artsy/positron/pull/1788/files#diff-ecf862e9da0cad0967c6864aad7a38cbR211

This is necessary because an empty paragraph will not have any height, but adding an element makes it assume the regular line-height.

PR #1793, opened 2018-11-21
add new has_gradient_overlay property

xtina-starr:
Addresses: https://artsyproduct.atlassian.net/browse/GROW-1009

This adds a new `has_gradient_overlay` property to the canvas model which will allow admins to control whether or not there is an overlay.

screen shot 2018-11-21 at 3 10 38 pm

To Do:
- [ ] Conditionally display an overlay in reaction: https://github.com/artsy/reaction/pull/1576

### DIFF
--- a/src/api/apps/graphql/schemas/display.js
+++ b/src/api/apps/graphql/schemas/display.js
@@ -16,7 +16,7 @@ const unitSchema = Joi.object()
     cover_img_url: Joi.string(),
     disclaimer: Joi.string(),
     headline: Joi.string(),
-    has_gradient_overlay: Joi.boolean(),
+    has_cover_overlay: Joi.boolean(),
     layout: Joi.string(),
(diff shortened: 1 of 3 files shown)


eessex:
Heads up i think we don't need to worry about backfilling the old content, since these are only running 5 at a time -- should be ok to give a heads up to content sales that they should update the running campaigns.

eessex:
Possible to post a screenshot of the UI?

xtina-starr:
I'm not sure what's causing that failing spec. Any ideas @eessex? I tried to see if I could re-run the test but I didn't have access. Could you try that?

I have to run but will see if it fails locally when I get home.

yuki24:
There seems to be some sort of testing pollution. We should probably re-visit but I'll go ahead and merge this with a patch. Merge on green

yuki24:
Build is 🍏

PR #1798, opened 2018-11-27
[FEATURE Video] Add a field that allows the video duration to be edited

yuki24:
finishes https://artsyproduct.atlassian.net/browse/GROW-959

This adds a duration input to the article section.

 * I went with two inputs rather than using a single input that holds a duration in the `mm:ss` format so we don't have to parse/format the duration value and the browser will take care of validating the user input.
 * I used `min` and `sec` to indicate which input refers to what unit as a starter, but I'm also open to other styles as this looks a lot different from what's in Zeplin (`mm:ss`)

## Screenshot

video-duration

### DIFF
--- a/src/client/apps/edit/components/admin/components/article/article_video_release_date.tsx
+++ b/src/client/apps/edit/components/admin/components/article/article_video_release_date.tsx
@@ -1,0 +1,97 @@
+import { Flex } from "@artsy/palette"
+import colors from "@artsy/reaction/dist/Assets/Colors"
+import { clone } from "lodash"
+import moment from "moment"
+import React, { Component } from "react"
+import { ArticleData } from "reaction/Components/Publishing/Typings"
+import styled from "styled-components"
+import { ButtonMedium } from "./button_medium"
+
+interface ArticleReleaseDateProps {
+  article: ArticleData
+  onChangeArticleAction: any
+}
+
+interface ArticleReleaseDateState {
+  hasChanged: boolean
+  releaseDate?: string
+}
+
+export class ArticleVideoReleaseDate extends Component<
+  ArticleReleaseDateProps,
+  ArticleReleaseDateState
+> {
+  constructor(props) {
+    super(props)
+
+    const { article } = props
+
+    this.state = {
+      hasChanged: false,
+      releaseDate:
+        article && article.media
+          ? moment(article.media.release_date).toISOString()
+          : undefined,
+    }
+  }
+
+  onMediaChange = (key, value) => {
+    const { article, onChangeArticleAction } = this.props
+    const media = clone(article.media) || {}
+
+    media[key] = value
+    onChangeArticleAction("media", media)
+  }
+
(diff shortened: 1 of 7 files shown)


eessex:
Looks good from here -- on style we could do sometime like the publish date, which styles two inputs to read as one.

yuki24:
just updated the inputs. Let me know if this still needs to be updated!

yuki24:
I've also moved the Release date and published fields to the Article section.

@eessex @owendodd The current video release date only allows for specifying the date so I didn't add the time field, but this is something we should add, please let me know.

video-fields

src/client/apps/edit/components/admin/components/article/article_video_release_date.tsx
```diff
@@ -0,0 +1,97 @@
...
+const Container = styled(Flex)`
+  margin-top: 5px;
+
+  button {
+    flex: 1;
+  }
+`
+
+const Input = styled.input`
+  width: 326px !important;
```
yuki24: I couldn't figure out a smarter way to get the same width as the publish date field, but I'm open to suggestions.

yuki24:
The CI build is green, but I've seen [the same failure](https://github.com/artsy/positron/pull/1793#issuecomment-440802752) we've seen before as https://github.com/artsy/positron/pull/1793#issuecomment-440802752. We should probably look into this soon.

owendodd:
If it is possible, I think it would be more logical for the video duration to be styled as MM:SS without the 'min' / 'sec' copy. But if that is complicated I think this latest update works!

src/client/apps/edit/components/admin/components/article/button_medium.tsx
```diff
@@ -0,0 +1,20 @@
+import colors from "@artsy/reaction/dist/Assets/Colors"
+import { avantgarde } from "@artsy/reaction/dist/Assets/Fonts"
+import styled from "styled-components"
+
+export const ButtonMedium = styled.button.attrs<{ background?: string }>({})`
```
eessex: Is it possible to use a palette button here?

yuki24: I've tried overriding the `Button` component using the `styled(Button)` trick, but this doesn't seem to override the predefined CSS:

```tsx
// src/client/apps/edit/components/admin/components/article/button_medium.tsx
export const ButtonMedium = styled(Button).attrs<{ hasChanged?: boolean }>({})`
  background: ${({ hasChanged }) => hasChanged && colors.redMedium} // this does not override the background color
`

// src/client/apps/edit/components/admin/components/article/article_publish_date.jsx
<ButtonMedium
  width="100%"
  onClick={this.onMediaReleaseChange}
  hasChanged={hasChanged}
  disabled={hasChanged}
>
  Update
</ButtonMedium>
```

The palette `Button` definitely looks simpler except we can't override the background color, but I'm curious as to what's causing the behaviour above.

PR #1801, opened 2018-11-29
Update README.md

lilyfromseattle:
Updated node version in README


eessex:
Don't forget to add a description!  Peril is our tool to keep PR's in a consistent format.

README.md
```diff
@@ -52,6 +46,12 @@ nvm alias default 8
 git clone git@github.com:craigspaeth/positron.git && cd positron
 ```
 
+- Copy `.env.example` to `.env` in the root of the project and edit all `REPLACE` values with sensitive configuration obtained from `positron-staging`. This should help.
```
eessex: Might be useful if we add something along the lines of 'Use the below script to copy...' and remove 'this should help'

PR #1813, opened 2018-12-08
PLATFORM-1080: Configure data dog tracing for positron

dleve123:
https://artsyproduct.atlassian.net/browse/PLATFORM-1080

* Style: add missing newline at EOF
* Explicitly configure services in tracer.ts
* Initialize DD tracing in boot.js
* Parameterize the hostname and the service name.
* Document process for local development against Data Dog in .env.example

---

Deploy Steps

- [ ] `hokusai staging update`
- [ ] Merge PR
- [ ] `hokusai production update`
- [ ] Deploy to production


dleve123:
@izakp @eessex @saolsen this isn't a WIP anymore (forgot to remove the label)

src/tracer.ts
```diff
@@ -0,0 +1,30 @@
+import tracer from "dd-trace"
+
+export function init() {
+  const ddServicePrefix = "positron"
+
+  tracer.init({
+    // Setting the service name to  `.undetected` so that we detect services that
```
izakp: Nice - how did you come across this pattern?

dleve123: Just via local development. Once I was able to connect to Data Dog for local development, it was relatively painless to rapidly prototype on the configuration layer.

.env.example
```diff
@@ -12,6 +12,32 @@ ELASTICSEARCH_PORT=9200
...
+#
+#```
+# docker run --name datadog-agent \
+#  --network hokusai_default \
+#  -v /var/run/docker.sock:/var/run/docker.sock:ro
+#  -v /proc/:/host/proc/:ro \
+#  -v /sys/fs/cgroup/:/host/sys/fs/cgroup:ro \
+#  -e DD_API_KEY=[GET-FROM-DD-UI] \
+#  -e DD_APM_ENABLED=true
+#  -e DD_HOSTNAME=daniel-phoenix
```
izakp: `- e DD_HOSTNAME=$(whoami)-phoenix` ?

dleve123: Yeah, or even a `"$(whoami)-$(hostname -s)"` - i can update.

dleve123: Maybe the hostname is a bit of overkill but /shrug, it was helpful for me to ensure that metrics were emanating from my computer

izakp:
@alloy remembering the work you did around spans in GraphQL - was all that work merged back upstream?  If so @dleve123 perhaps following up we should enable the GraphQL integration.

dleve123:
> @alloy remembering the work you did around spans in GraphQL - was all that work merged back upstream? If so @dleve123 perhaps following up we should enable the GraphQL integration.

So I initially included the graphql plugin. However, I believe that that plugin captures payloads that _involve requests to external GraphQL endpoints_, not _traffic to a local GraphQL API_. You can see in the [dashboard for the express service](https://app.datadoghq.com/apm/service/positron/express.request?end=1544541490507&env=none&paused=false&start=1541949490507) that it includes traces to the the GraphQL endpoints that Positron hosts. Definitely could be wrong here, but that's what I experienced.

alloy:
> @alloy remembering the work you did around spans in GraphQL - was all that work merged back upstream? 

Yeah all my work on the GraphQL plugin has been merged 👍 

> I believe that that plugin captures payloads that involve requests to external GraphQL endpoints, not traffic to a local GraphQL API.

The GraphQL plugin instruments a local schema's performance. I can't easily figure out what you were trying to link to, can you show me or include a screenshot?

dleve123:
> The GraphQL plugin instruments a local schema's performance. I can't easily figure out what you were trying to link to, can you show me or include a screenshot?

Here's an example trace for a GraphQL endpoint in being captured in the `postiron` service: https://app.datadoghq.com/apm/resource/positron/express.request/4f975c638bcf1350?end=1544547318514&env=none&paused=false&start=1541955318514

Does the MP DD configuration work as expected for GraphQL? Could you maybe link me to an example of the desired behavior in DD for MP?

src/tracer.ts
```diff
@@ -0,0 +1,30 @@
...
+    // are running as part of the application, but aren't explicitly configured
+    // with `tracer.use`. We want to to explicitly configure services so that we
+    // can enforce our `.`-deliminited Service naming convention.
+    service: `${ddServicePrefix}.undetected`,
+    hostname: process.env.DATADOG_AGENT_HOSTNAME,
```
izakp: A note about rolling this out...  Since the code will go out via CI running `hokusai deploy {Git-SHA}` and the config change adding the env var `DATADOG_AGENT_HOSTNAME` needs to be manually applied via `hokusai [staging|production] update`, perhaps default here to `datadog` as trace agent service is available in-cluster by this DNS name (tho it's not guaranteed to resolve to the IP of the container agent running on the same machine as the application's container - hence the use of the [downward api](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#the-downward-api)

dleve123: Don't totally following you here. If I patched the staging and production environments before rolling out the changes (via `hokusai [staging|production] update`), then the ENV vars would exist on the containers that will be created by the CI deploy, correct? Should that be the order of operations?

dleve123: @izakp and I synced IRL. We discussed the fact that a `hokusai staging update` depends on the local check-out of the Hokusai-relevant files. We concurred that setting the env first is a good approach though.

izakp: Ah yes - that would be even simpler if you set up the env var before the deploy rolls out.  Just be sure to run `hokusai [staging|production| update` while checking out this commit.  This loose coupling still manages to be a pain point in the process.  Want to work on it in https://github.com/artsy/hokusai/issues/61

src/tracer.ts
```diff
@@ -0,0 +1,30 @@
+import tracer from "dd-trace"
+
+export function init() {
+  const ddServicePrefix = "positron"
+
+  tracer.init({
```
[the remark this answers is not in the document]
dleve123: The tracer does still attempt to post messages to a local endpoint if the the hostname isn't set, but the app still functions as expected. With that said, we don't need to do anything DD if we don't have the hostname set, so I'll update the code. Good catch.

dleve123:
See https://artsy.slack.com/archives/CE4ENV8QZ/p1544561211003300 - deploying this first to staging, and then patching the staging environment

alloy:
It sure does :rocket: https://app.datadoghq.com/apm/resource/metaphysics.graphql/graphql.query/c1dc8c86a146ee60?end=1544621689067&paused=false&start=1544618089067&env=production

alloy:
I wonder if you’re initialising the tracer too late. As per the docs, it should be initialised _before_ loading any other modules. From very naive skimming of the data I don’t see anything but incoming Express requests, should there eg be requests to upstream services? Also, a single graphql request may not get instrumented.

dleve123:
@alloy 

 We're initializing the tracer in [`src/boot.js`](https://github.com/artsy/positron/blob/master/src/boot.js#L26-L28) which gets called in [`src/index.js`](https://github.com/artsy/positron/blob/master/src/index.js), which seems to require very generic dependencies.

I want to get this out there in prod so that we can better access the configuration once we have more meaningful traffic coming into it.

alloy:
Yeah, it should really be initialized _before_ any of these https://github.com/artsy/positron/blob/4be3eb5788b55356736845ce7ae4b1995973a4e3/src/boot.js#L5-L16.

See [these docs](https://docs.datadoghq.com/tracing/setup/nodejs/), specifically:

> This library MUST be imported and initialized before any instrumented module. When using a transpiler, you MUST import and initialize the tracer library in an external file and then import that file as a whole when building your application. This prevents hoisting and ensures that the tracer library gets imported and initialized before importing any other instrumented module.

The safest bet would be to load it from the `src/index.js` file, but I’m pretty sure that doing it above any of the imports will work too (you can verify this by looking at the output of the babel compiler)

PR #1815, opened 2018-12-10
Migrate cronjobs to K8s

ansor4:
First pass at trying to get the cronjobs currently running in heroku into kubernetes.

I have no idea how to test this, so any input on that front is highly appreciated!

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -105,3 +105,69 @@ spec:
     component: web
   sessionAffinity: None
   type: LoadBalancer
+
+---
+apiVersion: v1
+kind: CronJob
+metadata:
+  name: positron-scheduled-posts-cron
+spec:
+  schedule: "*/10 * * * *"
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+            - name: positron-scheduled-posts-cron
+              image: 585031190124.dkr.ecr.us-east-1.amazonaws.com/positron:production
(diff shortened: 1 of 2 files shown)


hokusai/staging.yml
```diff
@@ -105,3 +105,69 @@ spec:
...
+                    values:
+                    - background
+
+---
+apiVersion: v1
+kind: CronJob
+metadata:
+  name: positron-unqueue-cron
+spec:
+  schedule: "0 15 * * *"
```
izakp: @eessex good to run this at 15:00 UTC every day?

hokusai/staging.yml
```diff
@@ -105,3 +105,69 @@ spec:
...
+            nodeAffinity:
+              preferredDuringSchedulingIgnoredDuringExecution:
+              - weight: 1
+                preference:
+                  matchExpressions:
+                  - key: tier
+                    operator: In
+                    values:
+                    - background
+
```
dleve123: total nitpick - but I'd recommend removing the training whitespace here.

hokusai/production.yml
```diff
@@ -107,7 +107,7 @@ spec:
   type: LoadBalancer
 
 ---
-apiVersion: v1
```
izakp: Ah yes @ansor4 - a bit confusing as these deployment objects go through different API versions - but you can look up the spec for our k8s version (1.10.4) here https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#cronjob-v1beta1-batch

PR #1819, opened 2018-12-12
Run tests for positron before production deploy

dleve123:
The tests are cheap to run, and, from a gut feeling, it provides
developer confidence to see a passing test suite before a production
deploy.

More tactically, in the event that code is pushed to `release` (think
hot-fix scenario), then running a test build before a merge to `release`
will ensure that we catch any possible regressions from the integration
between `release` with `master.


yuki24:
LGTM!

PR #1824, opened 2018-12-17
Commit script to add indices to positron's db

ansor4:
After looking into how to add indices in code (a la mongoid in gravity), I think I'd rather go ahead and get eyes on the actual indices themselves first. If anything, we should first decide on the indices and add them via mongo shell first, as opposed to waiting to commit them into code.

This first draft is a list of recommended indices that mongodb's performance advisor recommends.


joeyAghion:
We created these indices this afternoon.

You can tell when: 

I don't recommend merging this as-is, but we could consider starting to track database migrations more formally in a location like `migrations/2018-12-17_add_article_indices.js`.

We'll discuss follow-up efforts (including user data and excessive+repetitive slugs) on [the ticket](https://artsyproduct.atlassian.net/browse/PLATFORM-1092).

For future reference, this helped us eyeball long `slugs` entries from a mongodb console:

```javascript
var slugs = []; db.articles.find({"slugs.10": {"$exists": true}}).forEach(function(d){ slugs.push(d.slugs); }); slugs
```

eessex:
@ansor4 what's the status of this PR?  Wasn't sure if there is outstanding work from comments.

ansor4:
@eessex sorry just saw this today, the only outstanding work needed was renaming the filename to reflect typical rails migration filenames. Just committed that change, so i'll close this out.

PR #1825, opened 2018-12-18
allow social embeds on feature and standard articles

xtina-starr:
Addresses: https://artsyproduct.atlassian.net/browse/GROW-793

This is the first part bit of work that adds social embeds to standard and feature articles. There is a subsequent PR in reaction that adds the social link to the UI. Ideally we'd want to bucket the two types of embeds into a togglable link but seeing that this is timely and needs to be used for EOY work we can take this approach and refactor to remove duplicated icons later.

screen shot 2018-12-17 at 7 43 38 pm

screen shot 2018-12-17 at 7 43 48 pm

### DIFF
--- a/src/client/apps/edit/components/content/section_tool/index.jsx
+++ b/src/client/apps/edit/components/content/section_tool/index.jsx
@@ -86,38 +86,51 @@ export class SectionTool extends Component {
             Text
           </SectionToolMenuItem>
 
-          <SectionToolMenuItem onClick={() => this.newSection("image_collection")}>
+          <SectionToolMenuItem
+            onClick={() => this.newSection("image_collection")}
+          >
             <IconEditImages />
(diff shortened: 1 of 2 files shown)


src/client/apps/edit/components/content/section_tool/index.jsx
```diff
@@ -225,10 +243,15 @@ export const SectionToolIcon = styled.div`
...
     z-index: 2;
   }
-  ${props => (props.isVisible || props.isOpen) && `
+  ${props =>
+    (props.isVisible || props.isOpen) &&
+    `
     opacity: 1;
-  `}
-  ${props => (props.isVisible && !props.isOpen && !props.isHero) && `
+  `} ${props =>
```
eessex: not blocking, but could use a semicolon here

PR #1833, opened 2019-01-04
Slug cleanup

lilyfromseattle:
Fixed "Invalid Date" and duplicate slug issues. Changed functionality so the slug is moved to the end of the slugs array if it already exists somewhere in the array, instead of appending the date to the slug and adding that to the array. Added more tests for slug creation.

There is one tiny edgecase that might cause an issue in theory but seems really unlikely. If there are two articles with the same title the second will append the publish date to the slug. If a third article is created with that same title and has a publish date on the same day as the publish date of the second article, then the third article will have the same slug as the second. I could fix this by appending a date format with hours/minutes/seconds to the slug if the slug with the same publish date exists but this would require another call to the db to check or completely changing the existing call, or I could change the date format to include hours/minutes/seconds for all appended dates. Neither seems worth it since it seems like an unlikely edgecase, but I can fix it if necessary.

### DIFF
--- a/src/api/apps/articles/test/model/save.test.coffee
+++ b/src/api/apps/articles/test/model/save.test.coffee
@@ -17,7 +17,7 @@ describe 'Save', ->
     app.use '/__gravity', gravity
     @server = app.listen 5000, ->
       done()
-    sandbox.useFakeTimers(new Date(2019,0,1,0,0,0))
+    sandbox.useFakeTimers(new Date(Date.UTC(2019, 0, 1, 0, 0, 0)))
 
   after ->
     @server.close()
@@ -112,7 +112,7 @@ describe 'Save', ->
           author: name: 'Molly'
           published_at: '2017-07-dsdfdf26T17:37:03.065Zsdfdf'
         }, (err, article) =>
-          article.slugs[0].should.equal 'molly-clockwork-01-01-19'
+          article.slugs[0].should.equal 'molly-clockwork-12-31-18'
           done()
       )(null, {
         slugs: ['molly-clockwork']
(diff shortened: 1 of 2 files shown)


src/api/apps/articles/test/model/save.test.coffee
```diff
@@ -85,6 +85,99 @@ describe 'Save', ->
...
+
+    it 'appends the current date to the slug if the slug exists in the database already but there is no publish date', (done) ->
+      todays_date = moment().format('MM-DD-YY')
+      Save.sanitizeAndSave( =>
+        Save.generateSlugs {
+          thumbnail_title: 'Clockwork'
+          published: true
+          author: name: 'Molly'
+        }, (err, article) =>
+          article.slugs[0].should.equal 'molly-clockwork' + '-' + todays_date
```
joeyAghion: My personal preference is for assertions like these to use explicit values rather than variables and interpolation. E.g., `...should.equal 'molly-clockwork-07-01-18'`. (This protects against the test passing while asserting against empty or buggy values.)

Not a big deal though.

lilyfromseattle: Can you think of a way to do this when testing that it appends today's date to the slug?

eessex:
This looks good when running green, and +1 to Joeys' suggestion of stubbing date libraries.
To @joeyAghion's q about precedents for scripts, yeah we have done backfills of data in the past -- usually writing a script for the transform, then running `yarn task scripts/backfill.js`

lilyfromseattle:
@joeyAghion Is backfilling the existing data necessary/part of the scope of this story?

joeyAghion:
Fixing existing data doesn't need to block the PR or deploy, but should be part of the story (because creating the necessary index will be possible with the bad data in place). Up to you if you want to task it out separately.

PR #1839, opened 2019-01-11
Add git-remote for Hokusai

dleve123:
Ensures that we're tracking deploys in artsy/positron's Github releases
for ease of deployment tracking.

### DIFF
--- a/hokusai/config.yml
+++ b/hokusai/config.yml
@@ -1,3 +1,4 @@
 ---
 project-name: positron
 pre-deploy: yarn publish-assets
+git-remote: git@github.com:artsy/positron.git


izakp:
I didn't add this in the original migration as there already exists a staging branch, which gets updated when deploying to staging via CI and didn't want to create a tag with the same name as release PRs get created from `staging` -> `release`.

I think the workflow could be preserved, however, if we drop the staging branch as well as the force-push to staging here https://github.com/artsy/positron/blob/master/.circleci/config.yml#L31 the PR can be made from the staging tag -> release branch and it should be equivalent.

@eessex wdyt?

dleve123:
This PR has been sitting for a while, and it seems like it isn't that big of a priority ATM. Gonna close to clean up the PR queue -- can always re-open later.

PR #1847, opened 2019-01-15
Article admin uses ts, palette componenets

eessex:
Refresh to the `/edit/admin` app ...
- Removes `react-styled-flexboxgrid` library
- Converts all components/tests to `.tsx` (except image upload, should be ticketed separately)
- Updates components to use `@artsy/palette` components
- Removes all `.styl` files

screen shot 2019-01-15 at 2 00 14 pm


src/client/apps/edit/components/admin/components/appearances.tsx
```diff
@@ -1,50 +1,50 @@
...
-import AutocompleteListMetaphysics from "client/components/autocomplete2/list_metaphysics"
+import AutocompleteListMetaphysics from "../../../../../components/autocomplete2/list_metaphysics"
 
-export const AdminAppearances = props => {
+export const AdminAppearances = () => {
   return (
-    <div>
-      <Row>
-        <Col xs={6}>
+    <React.Fragment>
```
yuki24: Could we use the `<>...</>` syntax or we aren't there yet?

src/client/apps/edit/components/admin/components/article/article_authors.tsx
```diff
@@ -45,7 +39,7 @@ export class ArticleAuthors extends Component {
         .query({ query: AuthorsQuery(idsToFetch) })
         .end((err, res) => {
           if (err) {
-            console.error(err)
+            new Error(err)
```
yuki24: I wonder if we ever want to raise an exception. We also may want to disable the TSLint inspection for `console.log`. @damassi Any thoughts?

src/client/apps/edit/components/admin/components/article/article_publish_date.tsx
```diff
@@ -1,17 +1,19 @@
...
-    onChange: PropTypes.func,
-  }
+interface ArticlePublishDateProps {
+  article: ArticleData
+  onChange: (key: string, value: any) => void
+}
+
+export class ArticlePublishDate extends Component<ArticlePublishDateProps> {
+  public date
+  public time
```
yuki24: Are there any objects that actually reference one of these members? Otherwise they could be private. Also specifying a type would be helpful since the names `date` and `time` imply they are plain value objects:

```ts
private date: HTMLInputElement
private time: HTMLInputElement
```

src/client/apps/edit/components/admin/components/featuring/featuring_mentioned.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export const FeaturingMentioned: React.SFC<{ model: string }> = props => {
+  const { model } = props
+
+  const field =
+    model === "artist" ? "primary_featured_artist_ids" : "featured_artwork_ids"
+
+  return (
+    <div>
+      <AutocompleteListMetaphysics field={field} model={`${model}s`} />
```
yuki24: Possibly this `s` is redundant? Could be `featured_artwork_idss`.

src/client/apps/edit/components/admin/components/sponsor.tsx
```diff
@@ -0,0 +1,101 @@
...
+})
+
+const mapDispatchToProps = {
+  onChangeArticleAction: onChangeArticle,
+}
+
+export default connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(AdminSponsor)
```
yuki24: We should probably look into using decorators for simple connected components like this but that could be a separate chore (though testing might be harder):

```tsx
@connect(state => ({ article: state.edit.article }), { onChangeArticleAction: onChangeArticle })
class AdminSponsor extends ... {
  ...
}
```

src/client/apps/edit/components/admin/components/article/article_publish_date.tsx
```diff
@@ -49,8 +51,8 @@ export class ArticlePublishDate extends Component {
 
   onUnschedule = () => {
     const { onChange } = this.props
-    let publish_date = null
-    let publish_time = null
+    const publish_date = null
+    const publish_time = null
```
yuki24: Setting `null` to a const feels a little odd. Perhaps these assignments could be inlined?

src/client/apps/edit/components/admin/components/appearances.tsx
damassi: tsx files support it

src/client/apps/edit/components/admin/components/article/article_authors.tsx
```diff
@@ -1,24 +1,18 @@
...
-  static propTypes = {
-    article: PropTypes.object,
-    apiURL: PropTypes.string,
-    isEditorial: PropTypes.bool,
-    onChangeArticleAction: PropTypes.func,
-    user: PropTypes.object,
-  }
+import { onChangeArticle } from "../../../../../../actions/edit/articleActions"
+import { AutocompleteList } from "../../../../../../components/autocomplete2/list"
+import AutocompleteListMetaphysics from "../../../../../../components/autocomplete2/list_metaphysics"
```
damassi: @eessex - we shouldn't be using anything but absolute paths everywhere; relative paths like this are a nightmare to refactor. Are you running into `tsconfig.json` issues?

src/client/apps/edit/components/admin/components/article/article_authors.tsx
damassi: In clientside code, unless it's totally necessary to stop execution and break the UX we should avoid raising exceptions like this. `console.error` (and perhaps a log to sentry) is more ideal.

src/client/apps/edit/components/admin/components/appearances.tsx
```diff
@@ -1,50 +1,50 @@
...
+      <Flex flexDirection={["column", "row"]}>
+        <Box width={["100%", "50%"]} pr={[0, 20]}>
```
damassi: Should be using [design-system `space` values](https://github.com/artsy/palette/blob/master/src/Theme.tsx#L125). 20px -> 2. This applies to all following code; 40 -> 4, etc. If a value is missing, it's best to find the next closest item so that our designs can remain consistent.

src/client/apps/edit/components/admin/components/article/article_authors.tsx
damassi: For a simple Sentry / logger example, see https://github.com/artsy/prediction/blob/master/src/utils/logger.js

src/client/apps/edit/components/admin/components/article/article_authors.tsx
```diff
@@ -61,32 +55,31 @@ export class ArticleAuthors extends Component {
...
-            <label>Primary Author</label>
-            <input
-              className="bordered-input"
-              defaultValue={name}
-              onChange={e => this.onChangeAuthor(e.target.value)}
-            />
-          </div>
-        </Col>
+      <Flex flexDirection={["column", "row"]}>
+        <Box width={["100%", "50%"]} pb={40} pr={[0, 20]}>
```
damassi:
```
<Box width={["100%", "50%"]} pb={4} pr={[0, 2]}>
```

src/client/apps/edit/components/admin/components/article/article_publish_date.tsx
eessex: This is for testing

src/client/apps/edit/components/admin/components/article/article_authors.tsx
damassi: The way `styled-system` works is it will 
1. take the value passed into the prop and see if it corresponds to a key in the theme file (corresponding to prop type), and then return the value
1. if it cant find the key, it will fall back to pixels

src/client/apps/edit/components/admin/components/featuring/featuring_mentioned.tsx
eessex: I can type this so is clearer, model is "artist" or "artwork"

src/client/apps/edit/components/admin/components/article/article_publish_date.tsx
damassi: React components obscure these props by default, so no need to add `private` (or `public`). These public / private designators do work well for normal classes however.

src/client/apps/edit/components/admin/components/article/article_authors.tsx
eessex: None of the tsx files are resolving correctly in tests, added a ticket to address here- https://artsyproduct.atlassian.net/browse/GROW-1086

src/client/apps/edit/components/admin/components/appearances.tsx
eessex: Maybe we don't have the theme provider here, these are not resolving correctly for me.

src/client/apps/edit/components/admin/components/article/article_video_release_date.tsx
```diff
@@ -1,11 +1,10 @@
-import { Flex } from "@artsy/palette"
-import colors from "@artsy/reaction/dist/Assets/Colors"
+import { Box, Button, Flex } from "@artsy/palette"
 import { clone } from "lodash"
 import moment from "moment"
 import React, { Component } from "react"
 import { ArticleData } from "reaction/Components/Publishing/Typings"
 import styled from "styled-components"
-import { ButtonMedium } from "./button_medium"
+import { FormLabel } from "../../../../../../components/form_label"
```
damassi: I mentioned it up above, but please avoid `../../../` 😄

src/client/apps/edit/components/admin/components/article/article_authors.tsx
eessex: We have an error boundary to catch client side errors and display them nicely, but yes, `console.error` is not approved by tslint

src/client/apps/edit/components/admin/components/featuring/featuring_mentioned.tsx
yuki24: Sorry for the confusion - please ignore this. I was looking at the wrong field when I saw the trailing `s`.

src/client/apps/edit/components/admin/components/sponsor.tsx
damassi: I'm a big fan of this decorator pattern as it makes things much more declarative, but best to do in a separate PR because the TS typings here are hard to define. Recently just got done battling this in Reaction :/

src/client/apps/edit/components/admin/components/super_article.tsx
```diff
@@ -45,7 +48,7 @@ export class AdminSuperArticle extends Component {
         .query({ query: SubArticleQuery(idsToFetch) })
         .end((err, res) => {
           if (err) {
-            console.error(err)
+            new Error(err)
```
damassi: maybe log is better?

src/client/apps/edit/components/admin/components/super_article.tsx
```diff
@@ -76,140 +79,140 @@ export class AdminSuperArticle extends Component {
...
-          </Col>
-        </Row>
-
-        <Row>
-          <Col xs={4}>
-            <div className="field-group">
-              <label>Partner Link Title</label>
-              <input
-                className="bordered-input"
+        <Box pb={40}>
```
damassi: `pb={4}`

src/client/apps/edit/components/admin/components/article/article_authors.tsx
damassi: Do they resolve correctly in source code?

src/client/apps/edit/components/admin/components/appearances.tsx
damassi: Ah yeah, should add 
```js
import { Theme } from  '@artsy/palette'
```

and put it at the root of your react tree.

src/client/apps/edit/components/admin/components/article/article_authors.tsx
eessex: Yeah they are fine in the component files

damassi: I added a test file and everything seemed good: https://github.com/artsy/positron/pull/1849

src/client/apps/edit/components/admin/components/article/article_authors.tsx
damassi: Ah cool, if things are being routed to the ErrorBoundary that seems ideal. 

If you did want to turn it off could do 

```
"no-console": [true, "log"]
```
That way it only guards against log, but allows "warn" and "error". 

We don't yet have a unified linting setup across artsy so in the meantime feel free to adjust based on app preferences.

src/client/apps/edit/components/admin/components/article/article_authors.tsx
damassi: Lol, well that blew up... on sec

src/client/apps/edit/components/admin/components/article/article_publish_date.tsx
eessex: Is there an example of a ref that doesn't use these I can look at?  I made them public so I could alter the values in tests.

eessex:
Ok to merge?

PR #1854, opened 2019-01-16
[WIP] Add Code Coverage Reporting

dleve123:
This PR modifies positron's build tooling so that we capture test coverage statistics across its complex test suite.

Notably, we're using codecov (free for open source) to capture and track the service's test coverage. It includes some nice visualizations and coveralls's node wrapper just didn't work for us (@kierangillen and I).

screen shot 2019-01-16 at 5 03 43 pm
(https://codecov.io/gh/artsy/positron)


package.json
```diff
@@ -14,13 +14,19 @@
...
     "publish-assets": "sh scripts/publish-assets.sh",
     "task": "node -r coffeescript/register -r @babel/register -r dotenv/config",
     "test:watch": "yarn jest -- --watch --runInBand",
     "test": "sh scripts/test.sh",
-    "type-check": "tsc --pretty --noEmit"
+    "type-check": "tsc --pretty --noEmit",
+    "install-codecov": "curl -s https://codecov.io/bash > .codecov && chmod +x ./.codecov",
+    "normalize-jest-coverage": "node scripts/normalize-jest-coverage.js",
+    "create-merged-coverage": "nyc report --reporter=text-lcov > coverage.lcov",
+    "report-coverage": "./.codecov -C $(cat COMMIT_HASH.txt) -p app/ -f coverage.lcov",
```
eessex: Do we want these files to be checked in, or is this something we'd want in the `.gitignore`?  Not sure if this is something we want available to all, or if it is more of a dev tool.

dleve123: which files? The COMMIT_HASH.txt have needed to be present for a bit to publish assets, but I could commit the codecov executable instead of `curl`-ing it down. I'll add coverage.lcov to the gitignore though incase someone runs this yarn script locally.

eessex: I guess whichever files are generated by these commands -- looks like the `.lcov` files

scripts/normalize-jest-coverage.js
```diff
@@ -0,0 +1,29 @@
+const libCoverage = require('istanbul-lib-coverage');
```
dleve123: hopefully we can remove this and use codecov's native report merging -- will see.

dleve123: Update: attempted to get codecov to natively detect and merge multiple test coverage files without success. Punting on further progress on this front for now to prioritize other initiatives.

.circleci/config.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       - setup_remote_docker
       - run:
           name: Test
-          command: COMMIT_HASH=$(git rev-parse --short HEAD) hokusai test --build
+          command: COMMIT_HASH=$(git rev-parse HEAD) CODECOV_TOKEN=$CODECOV_TOKEN hokusai test --build
```
izakp: It would be good to have these dynamic build args pulled in via Hokusai, and specified in a project's yml config - perhaps in a preprocessing step...

scripts/jest.sh
```diff
@@ -0,0 +1,6 @@
+set -e -x
+
+trap "exit" INT
+
+jest --maxWorkers=2 --detectOpenHandles --forceExit \
```
izakp: +1 for `--detectOpenHandles` - I think this has caused some strangeness in the past running the test suite in Docker...

scripts/jest.sh
```diff
@@ -0,0 +1,6 @@
+set -e -x
```
izakp: Add a shabang line plz  - `#!/bin/sh`

scripts/jest.sh
dleve123: FWIW, I didn't add this flag, it was already in the project before this change.

.circleci/config.yml
dleve123: I think, ideally, we wouldn't need to "thread" this Git-state information down to further Docker-layers. Maybe once we introduce this approach to another codebase, we refactor this to be a bit less coupled and a bit flatter.

I think we should re-visit this when we deem it worthwhile to measure code coverage across many projects though, and opting not to change now

PR #1859, opened 2019-01-17
[Related Articles] exclude related articles with unpublished media

xtina-starr:
Addresses: https://artsyproduct.atlassian.net/browse/GROW-940

This adjust the related articles query to exclude related articles with unpublished media.


xtina-starr:
@eessex I created a couple of articles to manually test this that are 404'ing in force. Do you have any idea why that might be?

xtina-starr:
I was also thinking to add an additional test to the resolver. It looks like I would have to update global fixture data to do that but that may affect other tests. Is there anyway around that?

eessex:
@xtina-starr for viewing articles in force, you may not be pointed to the correct db -- we recently added a new db for staging as part of the data-sync project.  Try `hokusai env get` for the new MONGOHQ_URL.

eessex:
A new test to resolvers sounds great to me, let me see if I can find a snippet to add new articles to the fixtures return value.

src/api/apps/articles/model/retrieve.js
```diff
@@ -0,0 +1,201 @@
+const _ = require("underscore")
+const { ObjectId } = require("mongojs")
+
+export const toQuery = function(input, callback) {
```
eessex: Can use the es6 syntax here, 
`export const toQuery = (input, callback) => {`

Also this should be ok to use `.ts` rather than `.js`

src/api/apps/articles/model/retrieve.js
```diff
@@ -0,0 +1,201 @@
+const _ = require("underscore")
```
eessex: We are moving towards using `lodash` over `underscore` -- I believe most of the functions used here are available in in `lodash`, worth trying.

eessex: Additionally for the imports you can use `import x from ...` rather than const.

src/api/apps/articles/test/model/retrieve.spec.js
```diff
@@ -0,0 +1,158 @@
...
+        ObjectId("5086df098523e60002000017")
+      )
+    })
+
+    it("aggregates the query for vertical", function() {
+      const { query } = toQuery({
+        vertical: "55356a9deca560a0137bb4a7",
+        published: true,
+      })
+      query["vertical.id"].should.containEql(
```
eessex: To switch to `jest` syntax use expect rather than should, 
`expect(query["vertical.id"]).toBe(ObjectId("55356a9deca560a0137bb4a7"))`
https://jestjs.io/docs/en/expect

src/api/apps/articles/model/index.js
```diff
@@ -7,7 +7,7 @@ import async from "async"
 import { cloneDeep } from "lodash"
 import schema from "./schema.coffee"
 import Joi from "../../../lib/joi.coffee"
-import retrieve from "./retrieve.coffee"
+import { toQuery } from "./retrieve.js"
```
eessex: Do we need an extension here? Also I'd assume that for coffee files we need to use a const like on line 13

eessex:
One more comment -- we should hold on merging this until tomorrow -- there are some problems with the staging branch not deploying correctly, hoping to resolve in the morning.

src/api/apps/articles/model/index.js
xtina-starr: Yeah we could probably get rid of the `.js` extension. 

When you are referring to the const are you meaning the import on line 9 needs to be change because it's a coffeescript file?

eessex: Yeah, we need to use `require` on all `.coffee` files, `import` works only with js and ts

eessex:
Looks good, will merge this as soon as our staging branch errors are resolved.

xtina-starr:
👍

PR #1862, opened 2019-01-17
Updates to display panel

eessex:
Quick refresh to display panel and CharacterLimit component:
- Convert to tsx
- Remove instances of `react-styled-flexboxgrid`
- Use palette components
- Add required indicators to required fields
- Remove `.styl` files (except for preview components, will save for another time)

screen shot 2019-01-17 at 3 10 55 pm
screen shot 2019-01-17 at 3 05 30 pm


src/client/apps/edit/components/display/components/partner.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { DisplayProps } from "../index"
+import { FillTitlePrompt } from "./magazine"
+const ImageUpload = require("client/apps/edit/components/admin/components/image_upload.coffee")
+
+export const DisplayPartner: React.SFC<DisplayProps> = props => {
+  const { article, onChangeArticleAction } = props
+
+  return (
+    <div>
+      <Intro pt={50}>
```
damassi: I keep thinking we need to add a `5` entry in the Theme file, though it seems like they deliberately didn't include.

damassi:
So solid 👍

PR #1866, opened 2019-01-18
Add code coverage reporting

dleve123:
### Code Coverage Info in CodeCov

screen shot 2019-01-22 at 6 26 37 pm

Anyone within the Artsy org should be able to sign into Codecov via Github-auth to get more granular coverage statistics: https://codecov.io/gh/artsy/positron

### Code Coverage Into in Console Output

screen shot 2019-01-23 at 1 02 46 pm

---

Supersedes https://github.com/artsy/positron/pull/1854 (wanted to rename branch) 

https://artsyproduct.atlassian.net/browse/PLATFORM-1136

Problem:

One part of the product health matrix story [spike][spike] is code
coverage for each service. However, there is no pattern at Artsy for
measuring code coverage for Javascript applications.

While there are existing tools out there for tracking and reporting code
coverage, such as [istanbul.js][istanbul-js], [coveralls][coveralls],
and [codecov][codecov], the complexities:

1. A multi-runner test suite (jest and mocha), and
2. A Circle CI -> Docker -> Docker test runtime

have resulted in a tough problem to solution around.

Solution:

After spending a lot of time trying to get Coveralls instrumentation to
work, we pivoted to using Codecov. At a high level, this commit does the
following:

1. Updates our calls to `jest` and `mocha` to write code coverage
information to a `.nyc_output` directory (default for nyc/istanbul)

2. Manipulates the `jest` coverage information to be compatible with
`mocha` coverage information via a JS script. See [motivating GH
issue][gh-issue].

3. Merges the `jest` and `mocha` coverage information together using the
`nyc` CLI

4. Introduces a bash executable provided by codecov to instrument the
merged information to codecov

Co-authored-by: Kieran Gillen <kgillen@gmail.com>

[spike]:https://artsyproduct.atlassian.net/browse/PLATFORM-1098
[istanbul-js]:https://github.com/istanbuljs/nyc
[coveralls]:https://coveralls.io/
[codecov]:https://codecov.io/gh/artsy/positron/
[gh-issue]:https://github.com/facebook/jest/issues/2418

### DIFF
--- a/scripts/normalize-jest-coverage.js
+++ b/scripts/normalize-jest-coverage.js
@@ -1,28 +1,27 @@
-const libCoverage = require('istanbul-lib-coverage');
-const { createReporter } = require('istanbul-api');
-const fs = require('fs')
-const path = require('path')
+const libCoverage = require("istanbul-lib-coverage")
+const { createReporter } = require("istanbul-api")
+const fs = require("fs")
+const path = require("path")
 
-const normalizeJestCoverage = (result) => {
-
+const normalizeJestCoverage = result => {
   Object.entries(result).forEach(([k, v]) => {
-    if (v.data) result[k] = v.data;
-  });
+    if (v.data) result[k] = v.data
+  })
 
-  return result;
-};
+  return result
+}
 
-const map = libCoverage.createCoverageMap();
-const jestCoverage = require('../.nyc_output/coverage-final.json');
+const map = libCoverage.createCoverageMap()
+const jestCoverage = require("../.nyc_output/coverage-final.json")
 
-map.merge(normalizeJestCoverage(jestCoverage));
+map.merge(normalizeJestCoverage(jestCoverage))
 
-const reporter = createReporter();
-reporter.addAll(['json']);
-reporter.write(map);
+const reporter = createReporter()
+reporter.addAll(["json"])
+reporter.write(map)
 
 fs.copyFile(
-  './coverage/coverage-final.json',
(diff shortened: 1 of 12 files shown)


scripts/normalize-jest-coverage.js
```diff
@@ -0,0 +1,29 @@
...
+map.merge(normalizeJestCoverage(jestCoverage));
+
+const reporter = createReporter();
+reporter.addAll(['json']);
+reporter.write(map);
+
+fs.copyFile(
+  './coverage/coverage-final.json',
+  './.nyc_output/coverage-final.json',
+  (error) => { if (error) console.log(error) }
```
damassi: This looks super readable (not suggesting a change) but could also do

```js
error => error && console.log(error) 
```

Also, by the way this is formatted I can tell your editor is missing a [Prettier integration] :) which we use for all of our JS repos based on configuration [located here](https://github.com/artsy/positron/blob/master/package.json#L206-L211)

dleve123:
@damassi Is Prettier compliance enforced during CI? I think style checking would be a bit more robust if we asserted upon a certain style during CI, instead of assuming that everyone use an editor that is integrated with Prettier.

damassi:
On all of our JS repos we typically run things like prettier during a pre-commit hook which is [defined here](https://github.com/artsy/positron/blob/master/package.json#L246) -- not too sure why this didn't run?

dleve123:
It's possible that I ran a `--no-verify` when committing at times -- that probably explains it :)

PR #1867, opened 2019-01-19
Update Display Panel components

eessex:
- Removes `react-styled-flexboxgrid`
- Converts to TSX
- Minor updates to use palette components


src/client/apps/settings/client/curations/display/index.tsx
```diff
@@ -30,8 +50,8 @@ export default class DisplayAdmin extends React.Component {
       {},
       {
         success: () => this.setState({ saveStatus: "Saved" }),
-        error: error => {
-          console.error(error)
+        error: _error => {
+          // console.error(error)
```
damassi: Not sure if this is intentional

PR #1878, opened 2019-01-30
Refactor Drag & Drop component, change authors order

eessex:
Related to https://artsyproduct.atlassian.net/browse/GROW-638
- Converts Drag & Drop component to `.tsx`
- Refactors D&D components to use ReactDOM instead of jQuery
- Adds D&D option to `AutocompleteList` enabled via `isDraggable` prop
- Updates image editing to use new D&D components

TODO in separate PR:
- Update sections list to use new D&D component
- Remove old D&D code

drag-drop-author
drag-drop-image


src/client/components/autocomplete2/list.tsx
```diff
@@ -57,55 +61,93 @@ export class AutocompleteList extends Component<
...
+  }
+
+  renderListItems = () => {
+    const { formatListItem, isDraggable } = this.props
+    const { items } = this.state
+
+    return items.map((item, i) => {
+      const title = item ? item.title || item.name : ""
+      return (
+        <ListItem key={i} isDraggable={isDraggable}>
```
iskounen: Since `isDraggable` is an optional prop, will `{isDraggable}` result in a false value if one is not provided?

iskounen:
This looks good to me. I have one question about optional props, but otherwise, these changes look good to go 👍.

src/client/components/autocomplete2/list.tsx
eessex: Yep exactly, we assume that the component will not be drag and drop unless explicitly specified.

src/client/components/autocomplete2/list.tsx
```diff
@@ -57,55 +61,93 @@ export class AutocompleteList extends Component<
...
         )}
         <Autocomplete {...this.props} />
-      </div>
+      </AutocompleteListContainer>
     )
   }
 }
 
-export const ListItem = styled(Flex)`
+export const ListItem = styled(Flex).attrs<{ isDraggable?: boolean }>({})`
```
l2succes: you can do the following to define Props for styled-components now
```typescript
styled(Flex)<{ isDraggable?: boolean }>`
  ...
`
```

src/client/apps/edit/components/content/sections/images/index.tsx
```diff
@@ -232,32 +237,38 @@ const SectionImagesContainer = styled.section.attrs<{
...
-const SectionImagesList = styled(Flex)`
+const SectionImagesList = styled(Flex).attrs<{
+  isEditing?: boolean
+  isWrapping?: boolean
+}>({})`
   position: relative;
-  z-index: -1;
+  z-index: ${props => (props.isEditing ? "2" : "-1")};
+
+  ${DragContainer} {
```
l2succes: Why not set the styles on `DragContainer` itself, in general it's good practice to encapsulate all styling/logic for a given component within it's definition & it's easier to reason about them.

src/client/components/autocomplete2/list.tsx
eessex: Actually we can't do this yet, need to bump up some tooling in Positron first.

l2succes: Ah ok, probably need to upgrade `styled-components` first

eessex: Yeah exactly, I'll make a separate PR for that.

PR #1887, opened 2019-02-04
Use newer styled-components syntax for typing attrs

eessex:
Was pretty certain this syntax wasn't working in a few weeks back, but at this moment it looks like we no longer need to use `attrs` when declaring styled component props.  This does a quick remove of the old syntax.


damassi:
👍

PR #1888, opened 2019-02-04
Use palette components for Image Generator

eessex:
Missed one field when the `Admin` panel was updated to `ts` & palette components


damassi:
👍

PR #1894, opened 2019-02-05
[Text QA] Remove empty text sections while editing

eessex:
Related to https://artsyproduct.atlassian.net/browse/GROW-977
Not sure at which point this functionality was (assuming unintentionally) removed, but Writer should not allow empty Text sections to be created (we already remove all other types of empty sections). 

- adds `maybeRemoveEmptyText` when user stops editing a section
- allows sections with an empty `<h1>` block to remain (used for section breaks)
- converts touched files to `.ts` and `.tsx`


xtina-starr:
LGTM!

PR #1911, opened 2019-02-13
Build out RichText tests, handle empty blocks

eessex:
- Restores rules for stripping empty blocks (unless h1) on html import/export and paste
- Adds tests for `RichText` component
- Finishes tests for rich test utils file

PR #1915, opened 2019-02-14
Move empty block stripping into rich_text

eessex:
Introduced a bug in https://github.com/artsy/positron/pull/1911

Issue:
Empty text blocks should not be allowed in `RichText`, however we accidentally removed this restriction when introducing optional linebreaks to the `Paragraph` component for this ticket: https://artsyproduct.atlassian.net/browse/GROW-961

This PR moves linebreak stripping out of `stripGoogleStyles` and into `convertToHtml` for `RichText`


xtina-starr:
This looks good to me.

PR #1921, opened 2019-02-25
Update styled components, @artsy packages

eessex:
- Updates `styled-components` to 4.1.3
- Updates reaction to v14
- Updates palette to v3
- Updates related libraries for typings and tests
- Fixes any tests affected by lib changes


package.json
```diff
@@ -34,12 +34,13 @@
     "relay-runtime": "https://github.com/alloy/relay/releases/download/v1.5.0-artsy.5/relay-runtime-1.5.0-artsy.5.tgz",
     "react": "^16.8.2",
     "react-relay": "https://github.com/alloy/relay/releases/download/v1.5.0-artsy.5/react-relay-1.5.0-artsy.5.tgz",
-    "react-test-renderer": "^16.8.2"
+    "react-test-renderer": "^16.8.2",
+    "styled-components": "4.1.3"
```
damassi: 👍

src/client/apps/settings/client/curations/gucci/test/metadata.test.tsx
```diff
@@ -155,7 +155,7 @@ describe("Metadata", () => {
         value: "Email Tags",
       },
     } as unknown) as React.FormEvent<HTMLInputElement>
-    input.onChange(event)
+    input.onChange && input.onChange(event)
```
damassi: Only q so far is why the need for these guards?

eessex: They raise a type error than input is possibly undefined. This is the quickest fix, but it is on my TODO's to change these components to palette's input, at which time will go back to these tests.

PR #1933, opened 2019-03-14
Update Hokusai configuration

eessex:
These changes have not yet been applied to production (though they are active in staging now). This PR addresses all items flagged in https://artsyproduct.atlassian.net/browse/PLATFORM-1151, also tracked via https://artsyproduct.atlassian.net/browse/GROW-1157

- Update hokusai required version
- Add pre-build definition for COMMIT_HASH
- Use autoscaling for replicas
- Update heath check endpoint delay/period/timeout seconds
- Fixes yml indentation

### DIFF
--- a/hokusai/config.yml
+++ b/hokusai/config.yml
@@ -1,3 +1,3 @@
----
 project-name: positron
 pre-deploy: yarn publish-assets
+hokusai-required-version: ">=0.5.3"
--- a/hokusai/config.yml
+++ b/hokusai/config.yml
@@ -1,3 +1,4 @@
 project-name: positron
+pre-build: export COMMIT_HASH=$(git rev-parse --short HEAD)
 pre-deploy: yarn publish-assets
 hokusai-required-version: ">=0.5.3"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,6 +21,10 @@ jobs:
           name: Configure
           command: hokusai configure --kubectl-version 1.10.7 --s3-bucket artsy-citadel --s3-key k8s/config --platform linux
       - run:
+          name: Validate Kubernetes Yaml
+          command: |
+            kubectl --context staging apply -f ./hokusai/staging.yml --dry-run
+      - run:
           name: Push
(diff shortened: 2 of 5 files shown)


eessex:
Hokusai changes have been applied to production.

hokusai/config.yml
```diff
@@ -1,3 +1,4 @@
----
 project-name: positron
+pre-build: export COMMIT_HASH=$(git rev-parse --short HEAD)
```
izakp: Can do this in Force too @saolsen @alloy @damassi

izakp:
Thanks for pairing on this @eessex !

Want to note that we can do a sanity-check of these changes with, (for instance, staging):
```
kubectl --context staging apply -f ./hokusai/staging.yml --dry-run
```

or production...
```
kubectl --context production apply -f ./hokusai/production.yml --dry-run
```

And can work that into CI or Hokusai as we see fit.

hokusai/config.yml
izakp: Ah @eessex with this we can get rid of other hardcoded `COMMIT_HASH=$(git rev-parse HEAD)` scattered about such as [here](https://github.com/artsy/positron/blob/master/.circleci/config.yml#L12) as this hook will ensure it's set before the build phase

izakp:
LGTM!  @eessex one minor comment but not a blocker, just some DRY-ing up of the references to `COMMIT_HASH`

hokusai/config.yml
damassi: In force we no longer need this because we removed bucket-assets in favor of strictly Webpack hashing. That work should also eventually be applied over here, but is less urgent since we don't (necessarily) need to worry about long-term caching due to Positron being an internal app.

damassi: But yah this looks way cleaner than how it was setup in force!

eessex:
Note -- COMMIT_HASH was removed from build steps, but a bug in Hokusai is preventing us from removing in the test step.  @izakp will follow up on a ticket for the Hokusai side and will update the test step once merged.

izakp:
Related Hokusai issue: https://github.com/artsy/hokusai/issues/145

hokusai/config.yml
izakp: Ah that's right!  Thanks @damassi

PR #1934, opened 2019-03-18
Bump @artsy libs

eessex:
Renovate PRs have been failing this week due to error writing to yarn lock. Investigating this further, but for now manually updating these libs.


zephraph:
Oh, interesting. I didn't see this.

PR #1938, opened 2019-03-19
Mock debounce in Draft js tests

eessex:
Lodash's `debounce`, has an unusual implementation of timers that doesn't play nicely with jest's timers.  For the Draft Js components we used `setTImeout` liberally to wait for updates before assertions, but this had a few problems -- tests are hard to read, they run slowly, and they are very unreliable in CI. 

This PR mocks `debounce` so that debounced functions in components are called immediately, removing the need to timeout. A handful of additional tests were also added.


eessex:
Draft tests are still relatively slow likely due to their size, but on this run the `RichText` component (110 tests) runtime is down to 14.496s from 42.554s

PR #1941, opened 2019-03-20
Convert Redux actions to TypeScript

eessex:
- Converts all redux actions to ts and adds typings where necessary
- Converts all related tests to ts and updates some to be more explicit/obvious to what they assert
- Adds additional tests for untested functions
- Adds comments for all actions to explain how they are used


damassi:
At some point we'll want to start getting proper intellisense in for Redux TS code, e.g., 

intellisense

(ignore the different redux project in ^, just for an example)

But that would be best for a different PR and some experimentation. Lots of docs around that here: https://redux.js.org/recipes/usage-with-typescript. For vanilla Redux the setup is a bit verbose :/

eessex:
@damassi yeah definitely interesting for the future, but not in scope here -- am focused on including documentation via comments how actions are used.

src/client/actions/test/edit/articleActions.test.ts
```diff
@@ -36,44 +38,63 @@ describe("articleActions", () => {
...
+        type: "CHANGE_ARTICLE",
+        payload: { data: { title: "Title" } },
+      })
     })
 
     it("Can accept a key/value pair as args", () => {
       changeArticleData("title", "Title")(dispatch, getState)
-      expect(dispatch.mock.calls[0][0].payload.data.title).toBe("Title")
+
+      expect(dispatch).toBeCalledWith({
```
xtina-starr: I'm not sure how this assertion is different from the one above on line 42.

src/client/actions/test/edit/articleActions.test.ts
```diff
@@ -111,52 +134,69 @@ describe("articleActions", () => {
...
-      }, 550)
+      expect(Backbone.sync.mock.calls[0][0]).toBe("update")
     })
   })
 
   describe("#publishArticle", () => {
     let getState
     let dispatch
-    let setArticleSpy = jest.spyOn(Article.prototype, "set")
+    const setArticleSpy = jest.spyOn(Article.prototype, "set")
```
xtina-starr: What's the different in using `spyOn()` versus `fn()` here?

eessex: `spyOn` lets you track calls to the function without overwriting/effecting its implementation

src/client/actions/test/edit/articleActions.test.ts
eessex: ahh you are totally correct -- updating now

eessex: This is actually a duplicate test, so going to remove

src/client/actions/test/edit/articleActions.test.ts
xtina-starr: Gotcha

xtina-starr:
@eessex feel free to merge on 🍏

PR #1945, opened 2019-03-25
Additional reducer tests

eessex:
Adds additional tests for redux reducers.

TODO: `articlesReducer` is not yet tested -- will address separately

### DIFF
--- a/src/client/reducers/appReducer.ts
+++ b/src/client/reducers/appReducer.ts
@@ -1,9 +1,19 @@
 import { data as sd } from "sharify"
 
-// TODO: confirm attrs
-interface UserState {
+interface ChannelState {
   name?: string
   id: string
+  type?: string
+}
+
+interface UserState {
+  access_token?: string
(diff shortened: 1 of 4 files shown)


src/client/reducers/appReducer.ts
```diff
@@ -25,19 +35,21 @@ export interface AppState {
...
-  channel: sd.CURRENT_CHANNEL,
-  forceURL: sd.FORCE_URL,
-  isAdmin: sd.USER.type === "Admin",
-  isEditorial: sd.CURRENT_CHANNEL.type === "editorial",
-  isPartnerChannel: sd.CURRENT_CHANNEL.type === "partner",
-  metaphysicsURL: sd.GRAPHQL_ENDPOINT,
-  user: sd.USER,
-}
+export const getInitialState = () =>
+  ({
```
eessex: Changed this to a function to help with tests -- this ensures that sd values are always fresh when `appReducer` is called

PR #1953, opened 2019-03-27
RichText2 QA/tests

eessex:
- Moves text-related redux actions from `sectionActions` to new `textSectionActions`
- Adds tests for `sections/text/index2`
- Fixes issue with container not focusing on click
- Fixes issue with calling `setSection` when focusing with tab-key
- Fixes issue z-index issue of text overlapping header menu
- Removes restriction on blockquotes for partner channels


src/client/actions/edit/textSectionActions.ts
```diff
@@ -0,0 +1,192 @@
...
+}
+
+/**
+ * Extract a blockquote from html and return any html blocks
+ * occurring before or after the blockquote
+ */
+export const extractBlockQuote = (html: string) => {
+  return (_dispatch, _getState) => {
+    let blockquote = html
+    const beforeHtml = _s(html).strLeft("<blockquote>")._wrapped
```
xtina-starr: What is `_wrapped` doing here?

PR #1956, opened 2019-03-28
Graphql can return authors on RelatedArticles

eessex:
- Updates API to support querying `authors` on the `relatedArticles` resolver
- Adds authors query to display authors in series editing `relatedArticles` display


xtina-starr:
This looks good!

PR #1968, opened 2019-04-03
Use ts for ArticlesList app and flesh out tests

eessex:
The ArticlesList was the least tested app in the repo, this converts the app fully to `ts` and brings test coverage up from 48% to 97%.

This also gives us an example of a client app that is fully ts/jest.


src/client/apps/articles_list/client.tsx
```diff
@@ -14,7 +14,9 @@ initWebsocket(store, sd.APP_URL)
 export const init = () =>
   render(
     <Provider store={store}>
-      <ArticlesList articles={sd.ARTICLES} published={sd.HAS_PUBLISHED} />
+      <Theme>
+        <ArticlesList articles={sd.ARTICLES} published={sd.HAS_PUBLISHED} />
```
xtina-starr: What does wrapping this in `<Theme>` do?

src/client/apps/articles_list/components/articles_list_header.tsx
```diff
@@ -0,0 +1,73 @@
...
+  top: 0;
+  left: 110px;
+  right: 0;
+  background-color: white;
+  z-index: 10;
+  border-bottom: 1px solid ${color("black10")};
+`
+
+const HeaderInner = styled(Flex)`
+  ${Flex} {
```
xtina-starr: These are nested Flex elements?

src/client/apps/articles_list/client.tsx
eessex: This wraps the app in Palette's theme, so it will receive the proper sizes for `m={1}` etc. and breakpoints.

src/client/apps/articles_list/components/articles_list_header.tsx
eessex: Yeah, I'm trying to target the wrapper for the `Tab` component from Palette which is not exported.

xtina-starr: It may be worth adding that as a comment.

src/client/apps/articles_list/client.tsx
xtina-starr: 👍

PR #1972, opened 2019-04-04
Use circle context hokusai

eessex:
Updates circle config to use context: hokusai. 

Follow up step: remove aws keys for Positron's circle env settings.


eessex:
Self merging this to test on staging deploy.

izakp:
Thanks @eessex !

PR #1982, opened 2019-04-09
Codecov updates

eessex:
- Adds extension flag for `nyc` to read coffee mocha tests
- Condenses mocha rules to search entire `/src` directory
- Moves coverage settings for jest to `package.json` jest rules
- Adds `FullScreenProvider` to tests where it was causing a large error when missing

### DIFF
--- a/package.json
+++ b/package.json
@@ -245,6 +245,12 @@
     ],
     "roots": [
       "<rootDir>/src"
+    ],
+    "coverageDirectory": "./.nyc_output/",
+    "collectCoverage": true,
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -245,6 +245,12 @@
     ],
     "roots": [
       "<rootDir>/src"
+    ],
+    "coverageDirectory": "./.nyc_output/",
+    "collectCoverage": true,
+    "coverageReporters": [
+      "text",
+      "json"
```
damassi: 👍

PR #1990, opened 2019-04-22
Don't hide click-off background for text

eessex:
Fixes https://artsyproduct.atlassian.net/browse/GROW-1190
We need this for `/edit2`, but not in production yet.


xtina-starr:
👍

PR #1991, opened 2019-04-22
Use codecov orb over local script

eessex:
Some cleanup to coverage reporting, this pr demonstrates a model for uploading coverage reports from multiple test runners.

- Use the `codecov` orb instead of saving this package's contents locally in `/scripts/codecov`
- Let codecov handle merging reports on their site and remove transformation logic
- Move `nyc` config to `package.json`
- Use `hokusai/test` step from orb over `yarn test` in ci


package.json
```diff
@@ -18,14 +18,12 @@
     "start": "sh scripts/start.sh",
     "production": "sh scripts/production.sh",
     "publish-assets": "sh scripts/publish-assets.sh",
+    "report-coverage": "./node_modules/.bin/codecov -c $COMMIT_HASH_FULL -b $BRANCH_NAME",
```
damassi: 👌

damassi:
@eessex - Can this be merged?

eessex:
Something is not connecting quite right, will follow up on this soon.

Dockerfile
```diff
@@ -10,7 +10,7 @@ RUN apk add --no-cache --quiet \
   bash \
   curl \
   dumb-init \
-  git && \
+  git mercurial && \
```
damassi: What is mercurial?

eessex: Mercurial provides the lib `hg` which is used by codecov's upload script-- however we did not need it in the end.

PR #2033, opened 2019-05-31
add VPN and Google DNS resolvers to development.yml

izakp:
Adding the VPC DNS server `10.0.0.2` as well as a fallback to Google's DNS servers `8.8.8.8` allows Positron development docker container to use VPN connections, if enabled.

### DIFF
--- a/hokusai/development.yml
+++ b/hokusai/development.yml
@@ -15,6 +15,9 @@ services:
     depends_on:
       - positron-mongodb
       - positron-elasticsearch
+    dns:
+      - 10.0.0.2
+      - 8.8.8.8
   positron-elasticsearch:
     image: artsy/elasticsearch:2.1
     ports:


izakp:
Huh actually we may only need to add `10.0.0.2` as the documentation states that "If the container cannot reach any of the IP addresses you specify, Google’s public DNS server 8.8.8.8 is added, so that your container can resolve internet domains." https://docs.docker.com/config/containers/container-networking/#dns-services

Or we may be able to configure the Docker daemon with DNS settings and not have to add this setting on a per-project basis... https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-configuration-file

izakp:
Just tested locally by setting DNS nameservers in the Damon config...

Screen Shot 2019-06-05 at 3 42 42 PM

I containers automatically use the staging VPN connection when connected, so I think it's best not to have to update project configs

```
$ docker run -ti nicolaka/netshoot bash
bash-5.0# host internal-ad4d5b35d878a11e9be6d123b2480b7a-1807827252.us-east-1.elb.amazonaws.com
internal-ad4d5b35d878a11e9be6d123b2480b7a-1807827252.us-east-1.elb.amazonaws.com has address 10.0.10.98
internal-ad4d5b35d878a11e9be6d123b2480b7a-1807827252.us-east-1.elb.amazonaws.com has address 10.0.11.234
internal-ad4d5b35d878a11e9be6d123b2480b7a-1807827252.us-east-1.elb.amazonaws.com has address 10.0.12.211
bash-5.0# curl http://internal-ad4d5b35d878a11e9be6d123b2480b7a-1807827252.us-east-1.elb.amazonaws.com:5601
<script>var hashRoute = '/app/kibana';
var defaultRoute = '/app/kibana';

var hash = window.location.hash;
if (hash.length) {
  window.location = hashRoute + hash;
} else {
  window.location = defaultRoute;
}</script>
```

PR #2045, opened 2019-06-04
Upgrade jQuery to v3

eessex:
Upgrade from v2 which has security vulnerabilities


eessex:
There was previously a renovate pr for this, but it seems to have been closed. Tried this locally and everything seems to work as expected.

zephraph:
🚢'in it

PR #2052, opened 2019-06-04
Remove unused html-janitor lib

eessex:
This package was flagged for severe security warnings, it does not appear to be used so removing.


zephraph:
Good catch!

PR #2061, opened 2019-06-06
Swap bucket-assets for @artsy/bucket-assets

eessex:
Updates to @artsy-namespaced `bucket-assets` lib, which has been updated (caused deprecation conflicts when updating `superagent` due to `knox` and `mime` versions).

### DIFF
--- a/scripts/publish-assets.sh
+++ b/scripts/publish-assets.sh
@@ -10,4 +10,4 @@ fi
 
 export COMMIT_HASH=`cat COMMIT_HASH.txt`
 
-yarn assets && yarn bucket-assets -b $S3_BUCKET
+yarn assets && ./node_modules/.bin/bucket-assets -b $S3_BUCKET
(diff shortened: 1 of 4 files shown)


scripts/publish-assets.sh
```diff
@@ -10,4 +10,4 @@ fi
 
 export COMMIT_HASH=`cat COMMIT_HASH.txt`
 
-yarn assets && yarn bucket-assets -b $S3_BUCKET
+yarn assets && ./node_modules/.bin/bucket-assets -b $S3_BUCKET
```
damassi: Why is this change necessary?

eessex: I was worried that the name change might affect this command, but might not be necessary -- will remove and see if it works as expected on staging deploy.

damassi: Nope, the `@foo` is a npm namespace scope: https://docs.npmjs.com/about-scopes for grouping packages. 
 What matters is the `node_modules/.bin/file-name` reference as those are what's made available to npm or yarn via `yarn x` commands.

PR #2065, opened 2019-06-06
Use nginx sidecar

eessex:
Deletes nginx rules from app and uses `default-websockets` rules instead.

Migration:
- [ ] `hokusai staging update`
- [ ] `hokusai production update`
- [ ] production deploy


hokusai/production.yml
```diff
@@ -61,6 +61,27 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
+        - name: positron-nginx
+          image: artsy/docker-nginx:1.14.2
+          ports:
+            - containerPort: 80
+            - containerPort: 8443
```
eessex: Not 100% sure on these ports

izakp: Those look good - will need to update the service spec to match - `80` here https://github.com/artsy/positron/blob/master/hokusai/staging.yml#L116 and `443` here https://github.com/artsy/positron/blob/master/hokusai/staging.yml#L120  (same in `production.yml`)

izakp: Also to use the `default-websockets` config, Positron itself should be configured to listen on port `8080` rather than `3005` so either set it with `hokusai staging env set PORT=3005` before running `hokusai staging update` or if you want to hardcode it into the container spec you can add this above the `envFrom` directive [here](https://github.com/artsy/positron/blob/master/hokusai/staging.yml#L30) (anything hardcoded will override any values set by `hokusai staging env` in the `positron-environment` ConfigMap
```
env:
  - name: PORT
     value: 8080
```

izakp:
One more thing!  Drop starting Nginx here https://github.com/artsy/positron/blob/master/scripts/production.sh#L14

hokusai/build.yml
```diff
@@ -5,3 +5,5 @@ services:
       context: ../
       args:
         - COMMIT_HASH
+    environment:
```
izakp: Not necessary to set that here - can leave it to the default `3005` locally (environment variables are set only when starting a container - values set here won't persist)

hokusai/production.yml
```diff
@@ -44,7 +43,7 @@ spec:
               memory: 1Gi
           readinessProbe:
             httpGet:
-              port: 3005
+              port: 8080
```
izakp: 👍

izakp:
Opened https://github.com/artsy/positron/pull/2079 - if we merge this first and use the named ports, Kubernetes should roll out port changes here with zero-downtime

izakp:
Merged changes from https://github.com/artsy/positron/pull/2079 and resolved conflicts

izakp:
This should be good to go with the named ports and deploying with hokusai version 0.5.6 and the `--update-config` flag...

`hokusai staging deploy TAG --update-config`

PR #2082, opened 2019-06-11
use artsy/hokusai@dev:0.2.0.43 orb

izakp:
Let's give it a go!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  hokusai: artsy/hokusai@0.1.0
+  hokusai: artsy/hokusai@dev:0.2.0.43
```
eessex: Why is this not `0.2`?

PR #2087, opened 2019-06-11
Docker cleanup

eessex:
Some minor updates to the Dockerfile based on recent changes by @dblandin in Force, is a bit more brief and declarative than the current implementation.
- get `dumb-init` from apk
- simplify yarn setup
- simplify deploy user setup


eessex:
I am seeing some errors related to permissions when running `hokusai dev start` related to the `watch` command and webpack. I'm assuming this will not affect production, where `watch` is not used, but need to find a workaround for development.

When running `hokusai dev start`:
`Error: EACCES: permission denied, watch '/app/src/client/public'`

Dockerfile
```diff
@@ -1,40 +1,39 @@
...
+
+WORKDIR /app
+
+# Ensure COMMIT_HASH is present
 ARG COMMIT_HASH
 RUN test -n "$COMMIT_HASH"
 
-RUN apk add curl git bash
+# Install system dependencies
+RUN apk add \
```
dblandin: Might want to add `--no-cache --silent` to this command.

Dockerfile
```diff
@@ -2,36 +2,33 @@ FROM node:10.13-alpine
 
 WORKDIR /app
 
+ENV PORT 3005
+EXPOSE 3005
+
 # Ensure COMMIT_HASH is present
 ARG COMMIT_HASH
 RUN test -n "$COMMIT_HASH"
+RUN echo $COMMIT_HASH > COMMIT_HASH.txt
```
dblandin: Since this value of `$COMMIT_HASH` is likely to change often, I think we should move it down so it doesn't interfere with the caching of other layers. The downside of moving these lines further down would be that you spend time building the docker image only to find out later that there's a required argument, but I think that's a reasonable trade off to ensure that we best utilize the docker layer caching mechanism.

Another tip is to collapse any neighboring `RUN` instructions to limit the total amount of docker layers defined.

For example:

```dockerfile
RUN test -n "$COMMIT_HASH" && \
    echo $COMMIT_HASH > COMMIT_HASH.txt
```

Dockerfile
```diff
@@ -2,36 +2,33 @@ FROM node:10.13-alpine
...
 
-# Add deploy user
-RUN adduser -D -g '' deploy
-
 # Install the packages
 COPY package.json yarn.lock ./
 RUN yarn install
 
+# Add deploy user
+RUN adduser -D -g '' deploy
```
dblandin: I think this can move up as well as there's nothing changing here build to build. Perhaps joined with the `apk` RUN instruction?

eessex: I'd rather leave things on separate lines so that it is clearer to those reading, is there any advantage to putting them into one command besides being shorter?

dblandin: Each separate docker instruction results in a separate docker layer, which adds complexity and time to docker pushes and pulls.

One option would be to retain the separate comment lines but still collapse to a single RUN instruction:

```dockerfile
# Install system dependencies
RUN apk add --no-cache --quiet \
      bash \
      curl \
      dumb-init \
      git && \
  # Add deploy user \
  adduser -D -g '' deploy
```

Dockerfile
```diff
@@ -1,40 +1,39 @@
...
-ADD https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_amd64 /usr/local/bin/dumb-init
-RUN chown deploy:deploy /usr/local/bin/dumb-init
-RUN chmod +x /usr/local/bin/dumb-init
+# Install the packages
+COPY package.json yarn.lock ./
+RUN yarn install
 
-RUN npm install -g yarn@1.15.2
+# Update file/directory permissions
+RUN chown -R deploy:deploy ./
```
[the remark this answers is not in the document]
izakp: 👍 indeed - would move after the `COPY` statement

izakp:
> I am seeing some errors related to permissions when running hokusai dev start related to the watch command and webpack. I'm assuming this will not affect production, where watch is not used, but need to find a workaround for development.

@eessex if it's simpler to use `yarn start` in development, I think that's fine - from what I can tell most devs don't use `hokusai dev start` when doing significant work on a local project - it's useful when you want to quickly bootstrap a project without installing all dependencies (and without syncing the app directory on each change watch won't matter much)

eessex:
Great news, moving that copy line fixed the watch issues -- all good to go!

PR #2143, opened 2019-07-09
Split Webpack into dev and production configs

eessex:
Related ticket: https://artsyproduct.atlassian.net/browse/PLATFORM-1508

This PR handles updates to the webpack config, but does not yet touch the asset pipeline. Will confirm that this works as expected in production, than switch to the s3 library (over bucket-assets) in a follow up PR. 

- Moves webpack config to subdirectory and breaks out dev/production config
- Updates webpack and related libs
- Chunks js by app


package.json
```diff
@@ -133,6 +134,7 @@
     "stylus": "0.54.5",
     "superagent": "3.7.0",
     "superagent-bluebird-promise": "3.0.2",
+    "terser-webpack-plugin": "^1.3.0",
```
zephraph: Terser is the default minifier that webpack uses. You very likely don't need to include it as a dependency.

webpack/environments/devConfig.js
```diff
@@ -0,0 +1,72 @@
+const chalk = require("chalk")
+// @ts-check
```
zephraph: Was this comment being on line 2 intentional?

webpack/environments/prodConfig.js
```diff
@@ -0,0 +1,33 @@
...
+    new HashedModuleIdsPlugin(),
+    new WebpackManifestPlugin({
+      fileName: path.resolve(__dirname, "../../manifest.json"),
+      basePath: "/assets/",
+      seed: getCSSManifest(),
+    }),
+  ],
+  optimization: {
+    minimizer: [
+      new TerserPlugin({
```
zephraph: Again, it _may_ be sufficient to just fallback to the defaults. I'm not 100% sure what the default configuration is though, so that might be worth double checking.

zephraph:
So far so good!

package.json
eessex: Looks like we do indeed need this.

webpack/environments/devConfig.js
eessex: Yes, I think I would need to install `@types/chalk` to get it to stop erroring -- doesn't seem worth it for the small amount we use it at the moment.

webpack/environments/baseConfig.js
```diff
@@ -0,0 +1,117 @@
...
+  basePath,
+  isCI,
+  isDeploy,
+} = require("../../src/lib/environment")
+const { getEntrypoints } = require("../utils/getEntrypoints")
+
+const baseConfig = {
+  devtool: "cheap-module-eval-source-map",
+  entry: {
+    webpack: [
```
zephraph: We should actually probably move this to the devConfig. I don't think you want to generate this for the production build.

zephraph: If you _do_ move this, it'd be good to run the build in both modes just to be sure the bundles are output like you expect.

src/client/assets/main.coffee
```diff
@@ -31,25 +31,60 @@ class Router extends Backbone.Router
     'settings/authors': 'authorsEdit'
 
   articleEdit: ->
-    require('../apps/edit/client').init()
+    require.ensure(
```
damassi: 🎉 

Side note: if this was JS/TS we could use `import()` here

eessex: Would like to get this router moved out of backbone, but it seems a chore in itself -- can look into tackling this and converting in a separate pr.

PR #2145, opened 2019-07-10
Use ts for related articles

eessex:
Quick update to the `relatedArticles` component (used for series sub-articles) to use typescript and styled components. Also limits articles to those from the same channel as the parent article.


src/client/apps/edit/components/content/sections/related_articles/index.tsx
```diff
@@ -146,7 +160,17 @@ export class RelatedArticles extends Component {
...
           onChange={this.onAddArticle}
         />
-      </div>
+      </Flex>
     )
   }
 }
+
+const RelatedArticlePreview = styled.div`
+  div[class^="ArticleCard__ImageContainer"] {
```
ashleyjelks: Is it always img content in this container? I think you're following BEM here for css naming conventions, in which case its not clear if ImageContainer correctly describes the Element

src/client/apps/edit/components/content/sections/related_articles/index.tsx
```diff
@@ -146,7 +160,17 @@ export class RelatedArticles extends Component {
...
+    min-height: 300px;
```
ashleyjelks: Why add this min-height here? Were these CSS props (background and min-height) defined in a stylesheet under the `RelatedArticles__preview` class previously?

src/client/apps/edit/components/content/sections/related_articles/index.tsx
eessex: Good point, would be smarter to export from reaction here and target `ImageContainer` -- I can do this as a follow up, for now it is able to target correctly.

src/client/apps/edit/components/content/sections/related_articles/index.tsx
eessex: This is a placeholder box for when an image is not present -- the empty list shows an empty article that demo's what the related article would look like.

src/client/components/autocomplete2/index.tsx
```diff
@@ -79,6 +80,12 @@ export class Autocomplete extends Component<
       remote: {
         url,
         filter: filter || returnItems,
```
ashleyjelks: This is just an additional safeguard to ensure the header is being set with validated user tokens?

eessex: Yeah, we need to manually set the access token for Bloodhound, before we were running un-authenticated requests. (the end game is showing a mix of published/unpublished content on the autocomplete, but that is not yet implemented here).

ashleyjelks: Nice!

src/client/apps/edit/components/content/sections/related_articles/index.tsx
```diff
@@ -41,9 +51,13 @@ export class RelatedArticles extends Component {
         .query({ query: RelatedArticleQuery(idsToFetch) })
         .end((err, res) => {
           if (err) {
-            console.error(err)
+            new Error(err)
+          }
+          const articles =
+            res && res.body && res.body.data && res.body.data.articles
```
xtina-starr: Do we use lodash in this repo? We could probably use [get](https://lodash.com/docs/4.17.14#get) here if so.

```js
const article = get(res, "res.body.data.articles")
```

But that's minor and not at all a blocker.

PR #2177, opened 2019-07-25
[HOLD] Update assets script

eessex:
Uses the `yarn webpack` command in assets rather than pointing to node modules, which ensures we pass all expected options (including config) to webpack.

### DIFF
--- a/scripts/assets.sh
+++ b/scripts/assets.sh
@@ -6,7 +6,7 @@ set -e -x
 
 yarn clean
 
-NODE_ENV=production node_modules/.bin/webpack
+NODE_ENV=production yarn webpack
 stylus \
   $(find src/client/assets -name '*.styl') \
   --compress \
--- a/src/lib/assetMiddleware.js
+++ b/src/lib/assetMiddleware.js
@@ -1,0 +1,37 @@
+import chalk from "chalk"
+import fs from "fs"
+import { isProduction } from "./environment"
+import path from "path"
+
+const { CDN_URL } = process.env
+
+export const assetMiddleware = () => {
+  if (isProduction) {
+    const manifestPath = path.resolve(process.cwd(), "manifest.json")
+
+    let manifest = {}
+    try {
+      manifest = JSON.parse(fs.readFileSync(manifestPath, { encoding: "utf8" }))
+    } catch (error) {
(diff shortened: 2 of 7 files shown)


scripts/assets.sh
```diff
@@ -6,7 +6,7 @@ set -e -x
 
 yarn clean
 
-NODE_ENV=production node_modules/.bin/webpack
+NODE_ENV=production yarn webpack
 stylus \
```
eessex: @damassi wanted to check -- now that we are using the assets compiled by webpack, is the stylus compression here necessary?

webpack/environments/prodConfig.js
```diff
@@ -24,8 +24,6 @@ module.exports.prodConfig = {
   optimization: {
     minimizer: [
       new TerserPlugin({
-        cache: false,
```
eessex: not sure why, but this change did not seem to make it into the last pr

scripts/assets.sh
damassi: The way it works in force is that when we started using the asset manifest plugin (which was a PR _after_ the webpack refactor) we build and [seed the CSS here](https://github.com/artsy/force/blob/master/webpack/envs/productionConfig.js#L29). 

I think removing this is for a later PR, once bucket-assets is ready to be removed completely along with its asset hashing.

src/lib/assetMiddleware.js
```diff
@@ -0,0 +1,37 @@
+import chalk from "chalk"
+import fs from "fs"
+import { isProduction } from "./environment"
+import path from "path"
+
+const { CDN_URL } = process.env
+
+export const assetMiddleware = () => {
```
damassi: Before adding the asset middleware we'll want to make sure the current webpack stuff in master is working 100%. In force I did things in two phases; 1) make webpack setup a little friendlier to changes over time; 2) remove bucket assets. 

These were the stages (with maybe a few intermediate prs inbetween that fixed issues): 
1) https://github.com/artsy/force/pull/3434
2) https://github.com/artsy/force/pull/3458
3) https://github.com/artsy/force/pull/3630

I would reference the Arena PR linked in the ticket, however. It's straight forward and to the point.

PR #2185, opened 2019-07-29
Revert to Webpack 3

eessex:
Made an error reverting the last work and missed the actual lib updates -- including a major version webpack update. This goes back to the original libs.


damassi:
One request is to be sure to test PRs before submitting for review / merge. I presume https://github.com/artsy/positron/pull/2179 would have failed without this?

PR #2237, opened 2019-08-13
Sorting articles by requested order

mbilokonsky:
I was initially going to try to use the `sort` input value on the query, but that expects to take a string of field names like "date_modified,-date_created" and then it generates a sorting rule based on how it parses that value.

But we want the output to be sorted into the same order as the input - and as far as I can tell, mongo requires us to do that in-memory at request time. So in this PR, we're just looking at the place where related articles are requested and we're ensuring that the returned set of articles is sorted based on the sort order of the original IDs returned.

This code has not been run or tested because I don't have positron fully operational yet.

### DIFF
--- a/src/api/apps/graphql/test/resolvers.spec.js
+++ b/src/api/apps/graphql/test/resolvers.spec.js
@@ -197,13 +197,11 @@ describe("resolvers", () => {
       results.length.should.equal(1)
     })
 
-    it("preserves sort order of related articles", async () => {
+    it.only("preserves sort order of related articles", async () => {
       const root = {
         ...SeriesArticle,
-        related_article_ids: SeriesArticle.related_articles,
       }
 
-      delete root.related_articles
       const related1 = {
         ...SeriesArticle,
         _id: "594a7e2254c37f00177c0ea9",
@@ -217,6 +215,11 @@ describe("resolvers", () => {
         id: "597b9f652d35b80017a2a6a7",
         related_article_ids: [],
       }
+
+      console.log("data:")
+      console.log(root)
+      console.log(related1)
+      console.log(related2)
 
       promisedMongoFetch.onFirstCall().resolves({
         total: 20,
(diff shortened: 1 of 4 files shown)


src/api/apps/graphql/test/resolvers.spec.js
```diff
@@ -196,6 +197,46 @@ describe("resolvers", () => {
...
+      }
+
+      const related2 = {
+        ...SeriesArticle,
+        _id: "597b9f652d35b80017a2a6a7",
+        id: "597b9f652d35b80017a2a6a7",
+        related_article_ids: [],
+      }
+
+      console.log("data:")
```
eessex: extra logs?

eessex:
Fixtures in reaction have been updated, try pulling the latest and hopefully this will go green.

yarn.lock
```diff
@@ -57,10 +57,10 @@
   resolved "https://registry.yarnpkg.com/@artsy/react-responsive-media/-/react-responsive-media-2.0.0-beta.5.tgz#d61e1a9f215d38d90ec2bb97dc1ff7409d0346fa"
   integrity sha512-XRYA77v/j3mVInwy5EYb8NtmMkkAD1/XDxDi45BO0oduv4hsljJfjtjK7jN0ziJ6JUm2QTx/t6QwuaT9E7vTuA==
 
-"@artsy/reaction@16.24.12":
-  version "16.24.12"
-  resolved "https://registry.yarnpkg.com/@artsy/reaction/-/reaction-16.24.12.tgz#c9ea1f2fff8259493205eae28d1bde6f27879747"
-  integrity sha512-NZaJOLgt4jeQQCCVh+UCTdr8tUtPbkjuypR35ukdYpHxRRadr6DzBddRQe2VRGiqk1dhwHtQEgeI55zN4wa6lw==
+"@artsy/reaction@16.24.8":
```
eessex: I think you'll need to run `yarn install` again, this is for an older version.

mbilokonsky: nice catch, thanks - updating now.

PR #2251, opened 2019-08-20
Related article can return their parent series

eessex:
Updates `relatedArticles` resolver to allow returning the `seriesArticle` that is its parent. Related to routing for Vanguard project.

### DIFF
--- a/src/api/apps/graphql/resolvers.js
+++ b/src/api/apps/graphql/resolvers.js
@@ -271,7 +271,6 @@ export const seriesArticle = root => {
   return new Promise(async (resolve, reject) => {
     const seriesArticles = await promisedMongoFetch({
       layout: "series",
-      published: true,
     }).catch(e => reject(e))
 
     seriesArticles.results.map(article => {
--- a/src/api/apps/graphql/test/resolvers.spec.js
+++ b/src/api/apps/graphql/test/resolvers.spec.js
@@ -197,7 +197,7 @@ describe("resolvers", () => {
       results.length.should.equal(1)
     })
 
-    it.only("preserves sort order of related articles", async () => {
+    it("preserves sort order of related articles", async () => {
       const root = {
         ...SeriesArticle,
       }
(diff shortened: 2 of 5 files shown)


src/api/apps/graphql/test/resolvers.spec.js
```diff
@@ -197,7 +197,7 @@ describe("resolvers", () => {
       results.length.should.equal(1)
     })
 
-    it.only("preserves sort order of related articles", async () => {
```
mbilokonsky: whoops, thanks!

src/api/apps/graphql/resolvers.js
```diff
@@ -271,7 +271,6 @@ export const seriesArticle = root => {
   return new Promise(async (resolve, reject) => {
     const seriesArticles = await promisedMongoFetch({
       layout: "series",
-      published: true,
```
mbilokonsky: Help me understand what's happening here - we had initially been saying that `published` was automatically true, but that was a bug? Or this is being removed and appended elsewhere, but I'm not seeing it (it only shows up in the integration mocks?)

Is this change being captured in any tests anywhere? Should we add a third mock to the integration test that isn't published and ensure that it comes through that way, or?

ashleyjelks: I think this is b/c we need to return both published and unpublished articles so the bool check is removed. We need to fetch unpublished articles b/c editorial is adding content to the series every day, as such we need to be able to fetch the articles that are still in draft mode in order to render them at the `/series/some-vanguard-slug` route. But @eessex please correct me if I'm wrong!

ashleyjelks:
@eessex lgtm! Merge when you're ready!

src/api/apps/graphql/resolvers.js
eessex: The articles api does not return unpublished content from users who do not have channel access, so this is extra -- it is hiding unpublished content from admins, making them unable to properly spec drafts.

PR #2319, opened 2019-09-24
Fix admin button active state

eessex:
Noticed while working on Vanguard, changes in palette's `Button` caused the active-state color to stop working. Switched to variant so these buttons are functional again.

**AFTER:**
Screen Shot 2019-09-24 at 4 19 47 PM

**BEFORE:**
Screen Shot 2019-09-24 at 4 20 08 PM


xtina-starr:
LGTM!

PR #2472, opened 2020-01-06
Fix series sub-article edit positioning

eessex:
Minor but noticed while doing QA on staging, buttons to edit and remove articles from a series were incorrectly positioned.

Before:
Screen Shot 2020-01-06 at 4 07 15 PM

After:
Screen Shot 2020-01-06 at 4 04 26 PM


eessex:
Going to merge this since a trivial change.

PR #2515, opened 2020-02-04
Introduce cypress tests

pepopowitz:
This PR introduces Cypress to the positron codebase. It needs some work before I'd consider merging it.

positron-e2e

### Why?

Faced with accruing dependabot PRs, I wanted to be able to verify that when I merged a dependency update, it didn't break things. I didn't want to run a ton of tests, just a few to verify that things weren't totally busted.

I included three tests: [I can edit an article and the changes show up on the Artsy site](https://github.com/artsy/positron/pull/2515/files#diff-b2e5ece24f4480f0970238921be94dbaR2), [SEO validation is hooked up when I'm viewing an article](https://github.com/artsy/positron/pull/2515/files#diff-b2e5ece24f4480f0970238921be94dbaR46), and [I can add a new tag](https://github.com/artsy/positron/pull/2515/files#diff-dcd940b22da2088457fcb30fca2541d0R2). The editing of an article is most important; the others were just some easy things I knew I could do without messing with the data too much.

### Why not use the existing testing infrastructure in positron?

To be honest, I don't know if the existing integration tests work. I couldn't get any meaningful feedback from running the existing tests. I am not sure if tests are passing or failing locally - I just can't get enough info from the output to tell. I poked around with this for a while, but decided it'd be interesting to start exploring Cypress.

### Room for improvement

The method of authentication is maybe not great. I initially attempted to let the browser redirect me to stagingapi.artsy.net and enter a user/password there, but I couldn't get it to function with Cypress. I might just need to take another look at it, but to get myself quickly to passing tests, I instead chose to [set the cookies](https://github.com/artsy/positron/pull/2515/files#diff-6adfff58252b1c35d133c5a20efaf83aR31) that positron needs to recognize a user. They're configured in my `cypress.env.json`, which is not included in the PR, but an example of that file is included.

#### 1. [`testing-library`](https://github.com/testing-library/cypress-testing-library) is really nice for finding elements the way a user would.

I've used `testing-library` in Jest tests, and really like it. The Cypress version is just as nice, and allows us to write element selectors that are more meaningful and declarative.

#### 4. Working with temporary variables is strange but doable

I'd read that [you can't really use temporary variables in Cypress tests](https://docs.cypress.io/guides/core-concepts/variables-and-aliases.html) due to the fact that each line in a test is effectively a single statement queued up for later playback, but I didn't think much of it because I didn't know when I'd run into it. I ran into it in this PR! I didn't want to hardcode the article ID, but grab it from the URL when editing an article instead. This was a bit harder than I thought, and it took me a while to find the right combination of Cypress commands. To store the value, I had to [extract the ID from the URL with a combination of `.location(pathname)` to get the full path, `.invoke()` to call the `.split()` method on the resulting string, `.its(2)` to access a property on the resulting array, and then `.as()` it to store it for future access](https://github.com/artsy/positron/pull/2515/files#diff-b2e5ece24f4480f0970238921be94dbaR78). Later in the test, I [use `cy.get()` to access the stored value, and make a `cy.request()` to a URL constructed from the ID](https://github.com/artsy/positron/pull/2515/files#diff-b2e5ece24f4480f0970238921be94dbaR78).


eessex:
Regarding auth, I’m ok with using an env for setting cookies as this will prevent us from depending on gravity staging to run the suite. Is it possible to set these in the test env, or CircleCI? I’d rather not expose these values. 

For the issue of seeing the changes published in Force/browser tabs, can we add a test that hits positron’s rest api directly to confirm that the db has saved the changes? Cypress should be able to read the json output at, for example, https://stagingwriter.artsy.net/api/articles

On repeating tests, this is a *bit* hacky, but we could choose a favorite article, and edit it in the before block for the article list test.
Another option would be to count that 10 articles are present, and ignore confirming the actual titles.

pepopowitz:
> Regarding auth, I’m ok with using an env for setting cookies as this will prevent us from depending on gravity staging to run the suite. Is it possible to set these in the test env, or CircleCI? I’d rather not expose these values.

The values of the cookies are defined in the `cypress.env.json` ([see `cypress.env.json.example`](https://github.com/artsy/positron/pull/2515/files#diff-c3beea501f1db8181b570a4a652c4206R2)). I am pretty sure the circleCI values would be used if they were configured, just like for integrity. I only meant that they came from staging in the sense that that's where I originally pulled them from, to put in my local `cypress.env.json`. 

---

> For the issue of seeing the changes published in Force/browser tabs, can we add a test that hits positron’s rest api directly to confirm that the db has saved the changes? Cypress should be able to read the json output at, for example, https://stagingwriter.artsy.net/api/articles

Good idea! I'll investigate.

---

> On repeating tests, this is a bit hacky, but we could choose a favorite article, and edit it in the before block for the article list test.
Another option would be to count that 10 articles are present, and ignore confirming the actual titles.

Yeah, great suggestions. I was locked into thinking I needed to verify the page had loaded by checking for a recent article. I could totally verify the page has loaded just by making sure _any_ articles are loaded. And honestly, the test already does a search for a specific article, so it might not even need to wait for the articles to load. I'll look at changing this, too!

pepopowitz:
@eessex I made the changes you suggested. Take a look again when you have a chance, and let me know if you think this is a worthwhile PR, or if there's more work I can do.

package.json
```diff
@@ -22,6 +22,7 @@
     "task": "node -r coffeescript/register -r @babel/register -r dotenv/config",
     "test:watch": "yarn jest -- --watch --runInBand",
     "test": "sh scripts/test.sh",
+    "test:e2e": "cypress open",
```
eessex: Lets go with `acceptance` to match convention in Force

pepopowitz:
Status update: I renamed the command to align with Force (force uses `cypress` and `cypress:run` commands for cypress tests), and attempted to resolve the merge conflict on `yarn.lock` by merging master in. I think something downstream got updated in the lockfile, as the original tests are failing with a bunch of `TypeError: jsdom.env is not a function` errors.

I'll dig into those errors when I pick up this work again.

pepopowitz:
Woohoo! Updating `benv` resolved the test failures. I'm going to merge this and work on connecting to CI as a follow-up.

PR #2526, opened 2020-02-10
Fix position bug for removing users from channels

eessex:
Minor change to make legacy autocomplete components contain their remove button, so we can remove former employees from editorial channel.

After:
Screen Shot 2020-02-10 at 11 09 32 AM

Before:
Screen Shot 2020-02-10 at 11 09 44 AM


eessex:
Merging since relatively trivial

PR #2641, opened 2020-04-17
Update admin authorization to check team role vs Admin type

dblandin:
Whenever we check for admin authorization, check that the currently authenticated user has the `team` role vs the `Admin` user type.

I chose to only change the underlying check logic and not the names of functions or spec descriptions to keep the diff focused. Happy to follow-up with those changes in a subsequent pull request.

### DIFF
--- a/src/api/apps/articles/routes.coffee
+++ b/src/api/apps/articles/routes.coffee
@@ -59,7 +59,9 @@ User = require '../users/model.coffee'
 
 # Don't let non-admins feature
 @restrictFeature = (req, res, next) ->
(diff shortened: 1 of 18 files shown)


eessex:
Looks like some failing specs related to these changes. I wonder if all staff automatically has the team role?

dblandin:
> Looks like some failing specs related to these changes. I wonder if all staff automatically has the team role?

That might be true. Honestly, I was struggling with these changes. I'd love to pair on this with you sometime!

dblandin:
Going to go ahead and close this for now.

PR #2734, opened 2020-06-08
[WIP] AutocompleteV2 uses metaphysics v2

eessex:
Updates metaphysics autocomplete components to target MP v2. Shows are still fetching from v1.

I ran into an issue where we do not have endpoints to fetch a list of items by `id` for all models. An equivalent to `partner_shows` from v1 does not seem to exist. Partners exist, but labelled as `_unused_gravity_partners`. I've used this for the moment, but hopefully there is a better option.


eessex:
Closing in favor of https://github.com/artsy/positron/pull/2760

PR #2801, opened 2020-08-17
Sanitized links use https

eessex:
Updates `sanitizeLink` to: 
- add https protocol if no protocol present
- use https for all artsy links
- add `www.` to `artsy.net` links
- use typescript
 
Also adds `news_source.url` and `postscript` to fields sanitized on save.

I've run the updated `sanitizeAndSave` as a backfill against staging to confirmed all works as expected. 

https://artsyproduct.atlassian.net/browse/PLATFORM-2715

### DIFF
--- a/src/api/apps/articles/model/save.coffee
+++ b/src/api/apps/articles/model/save.coffee
@@ -192,11 +192,10 @@ sanitize = (article) ->
 sanitizeLink = (urlString) ->
   return unless urlString
   u = url.parse urlString
+  return urlString if u.protocol is "https:"
+
   if u.protocol
-    if u.protocol is "http:"
-      urlString.replace("http:", "https:")
-    else
-      urlString
+    urlString.replace("http:", "https:") if u.hostname is "artsy.net"
   else
     'https://' + u.href
 
--- a/src/api/apps/articles/model/save.coffee
+++ b/src/api/apps/articles/model/save.coffee
@@ -20,11 +20,11 @@ artsyXapp = require('artsy-xapp')
   @generateSlugs article, cb
 
 @onUnpublish = (article, cb) =>
-  @generateSlugs article, (err, article) =>
-    deleteArticleFromSailthru getArticleUrl(article), =>
+  @generateSlugs article, (err, article) ->
+    deleteArticleFromSailthru getArticleUrl(article), ->
       cb null, article
 
-setOnPublishFields = (article) =>
+setOnPublishFields = (article) ->
   article.email_metadata = article.email_metadata or {}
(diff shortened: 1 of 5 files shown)


src/api/apps/articles/model/save.coffee
```diff
@@ -187,8 +190,15 @@ sanitize = (article) ->
...
+  return unless urlString
   u = url.parse urlString
-  if u.protocol then urlString else 'http://' + u.href
+  if u.protocol
+    if u.protocol is "http:"
+      urlString.replace("http:", "https:")
+    else
+      urlString
+  else
+    'https://' + u.href
```
joeyAghion, line 201: I can't tell from this PR where else this method is called. I think we can _only_ transform the protocol for `artsy.net` links. Otherwise, we might be prescribing SSL for an external site that doesn't support it (leading to a broken link).

eessex: Cool I misunderstood- will update this to force https only to artsy.net

eessex: This is now updated to only force https for artsy links or links without a protocol. To your question of where this is run, `sanitize` calls the message on the article body text and on explicitly defined article keys. This PR adds a couple new fields to ensure we hit anywhere that site link would be saved.

src/api/apps/articles/model/save.coffee
```diff
@@ -187,8 +190,14 @@ sanitize = (article) ->
   sanitized
 
 sanitizeLink = (urlString) ->
+  return unless urlString
   u = url.parse urlString
-  if u.protocol then urlString else 'http://' + u.href
+  return urlString if u.protocol is "https:"
+
+  if u.protocol
+    urlString.replace("http:", "https:") if u.hostname is "artsy.net"
```
joeyAghion, line 198: My coffeescript may be rusty, but does this line risk returning `null` when there _is_ a protocol but the hostname isn't artsy.net?

joeyAghion: Also, a hostname of `www.artsy.net` _should_ trigger this treatment. (Ideally even all `*.artsy.net` hostnames, but I know there are one or 2 that were never upgraded for SSL.)

joeyAghion: Sorry if this^ wasn't clear. The most recent commit strips `www` out of links, but it's preferable to keep that.

Basically, all of `http://artsy.net/*`, `http://www.artsy.net/*`, `artsy.net/*` and `www.artsy.net/*` should be replaced by `https://www.artsy.net/*`.

PR #2889, opened 2020-12-10
fix: changes for hokusai dev/test.

artsyjian:
- fix known problem with docker-compose `extends`.
- speed up dev stack start (notes inline).
- make development.yml services use network mode `host`, so that positron container can access mongo/es containers at `localhost`, so those URL's can be taken from `.env` (consistent with native local dev) and no need to redefine them in development.yml.
- re-organize README to delineate steps between manual vs hokusai dev.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,17 +18,19 @@ RUN apk add --no-cache --quiet \
 COPY package.json yarn.lock ./
 RUN yarn install --frozen-lockfile && yarn cache clean
 
-# Copy application code
-COPY . ./
+# Update file/directory permissions
+RUN chown -R deploy:deploy ./
 
-# Ensure COMMIT_HASH is present
-RUN test -n "$COMMIT_HASH" && \
-  echo $COMMIT_HASH > COMMIT_HASH.txt && \
-  # Update file/directory permissions
-  chown -R deploy:deploy ./
+# Copy application code
+COPY --chown=deploy:deploy . ./
 
 # Switch to less-privileged user
 USER deploy
+
+# Ensure COMMIT_HASH is present
+# Run this step as late as possible b/c it busts docker cache.
+RUN test -n "$COMMIT_HASH" && \
+  echo $COMMIT_HASH > COMMIT_HASH.txt
 
 ENTRYPOINT ["/usr/bin/dumb-init", "--"]
 CMD ["yarn", "start"]
--- a/.env.example
+++ b/.env.example
@@ -9,7 +9,7 @@ SESSION_SECRET=p0s1tr0n
 ELASTICSEARCH_PORT=9200
 
 # Local development only for the below; refers to the key in docker-compose.yml
-ELASTICSEARCH_URL=http://elasticsearch:9200
+ELASTICSEARCH_URL=http://localhost:9200
 MONGOHQ_URL=mongodb://localhost:27017/positron
 
 ##
--- a/README.md
+++ b/README.md
@@ -80,9 +80,7 @@ brew services start elasticsearch
 
 #### Using staging database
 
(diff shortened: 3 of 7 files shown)


Dockerfile
```diff
@@ -18,17 +18,19 @@ RUN apk add --no-cache --quiet \
...
-# Copy application code
-COPY . ./
+# Update file/directory permissions
+RUN chown -R deploy:deploy ./
 
-# Ensure COMMIT_HASH is present
-RUN test -n "$COMMIT_HASH" && \
-  echo $COMMIT_HASH > COMMIT_HASH.txt && \
-  # Update file/directory permissions
-  chown -R deploy:deploy ./
```
artsyjian, line 28: Docker re-runs `chown -R...` upon every `hokusai dev start` which I had to do frequently, and it takes ~10min, excruciating. It seems that the `COMMIT_HASH` part busts docker cache. So I am breaking this step up: put `COMMIT_HASH` near the end and after `USER` deploy so the file is owned by deploy user, put `chown -R...` before COPY, so its cache is not busted by local file changes, add `--chown` flag to COPY to take care of files copied-in.

eessex (approved):
❤️ ❤️ ❤️  Thanks for these improvements!

artsyjian:
Looks like an ARG's value is tied with many instructions in docker cache, so when the ARG's value is changed, cache is busted for those instructions. This is a bit of an obstacle. The README says to run hokusai dev with `COMMIT_HASH=$(git rev-parse --short HEAD) hokusai dev start`, and that hash changes after every `git commit`. Let me see if there's a way to decouple the ARG from caches.

```
jian@artsy:~/code/positron$ docker history hokusai_positron
IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
2013bbaab310        48 minutes ago      /bin/sh -c #(nop)  CMD ["yarn" "start"]         0B
c046c12e78aa        48 minutes ago      /bin/sh -c #(nop)  ENTRYPOINT ["/usr/bin/dum…   0B
427fa7d92416        48 minutes ago      |1 COMMIT_HASH=7c7fc703 /bin/sh -c test -n "…   9B
a86d7d9cf99a        48 minutes ago      /bin/sh -c #(nop)  USER deploy                  0B
a6a3f7503f47        48 minutes ago      /bin/sh -c #(nop) COPY --chown=deploy:deploy…   6.08MB
194939b64392        About an hour ago   |1 COMMIT_HASH=7c7fc703 /bin/sh -c chown -R …   397MB
df8f63b7dcdc        About an hour ago   |1 COMMIT_HASH=7c7fc703 /bin/sh -c yarn inst…   862MB
02d59be4cc41        2 hours ago         /bin/sh -c #(nop) COPY multi:5de88c50c172b1a…   630kB
89c8b86be3b3        2 hours ago         |1 COMMIT_HASH=7c7fc703 /bin/sh -c apk add -…   18.6MB
dbc7ac1912ae        42 hours ago        /bin/sh -c #(nop) WORKDIR /app                  0B
...
```

https://stackoverflow.com/questions/55619593/bust-cache-bust-within-dockerfile-without-providing-external-build-args

Dockerfile
```diff
@@ -18,17 +18,19 @@ RUN apk add --no-cache --quiet \
 COPY package.json yarn.lock ./
 RUN yarn install --frozen-lockfile && yarn cache clean
 
-# Copy application code
```
eessex, line 21: @artsyjian I think we needed to keep this copy - staging releases are failing with no templates: https://app.circleci.com/pipelines/github/artsy/positron/4081/workflows/1cfc8032-b2f2-4cf5-a876-97ef30377e3f/jobs/9090

PR #2893, opened 2020-12-14
chore: Show Admin Tab (on Article Edit) based on team role, not admin role.

artsyjian:
https://artsyproduct.atlassian.net/browse/PLATFORM-2320.

Parse user's roles out of JWT per this [discussion](https://artsy.slack.com/archives/CP9P4KR35/p1607710756442900).

The [admin](https://github.com/artsy/positron/tree/master/src/client/apps/edit/components/admin) component [references](https://github.com/artsy/positron/blob/9446e57cbe657308baa11bd47e3b366e82c896a0/src/client/apps/edit/components/admin/components/article/article_authors.tsx#L91) bunch of app api routes. Those routes are open to non-admins for GET's, but many [require](https://github.com/artsy/positron/blob/9446e57cbe657308baa11bd47e3b366e82c896a0/src/api/apps/tags/index.coffee#L9) admin for `put/post/delete`. Those actions don't seem to be performed by the component, so I think no changes are required.

The component also [queries](https://github.com/artsy/positron/blob/9446e57cbe657308baa11bd47e3b366e82c896a0/src/client/apps/edit/components/admin/components/featuring/index.tsx#L11) Gravity/Metaphysics. I wonder if any changes are required there.

[settings](https://github.com/artsy/positron/tree/master/src/client/apps/settings) app is also restricted to `admin` role. The actions performed in it seem too sensitive for `team` role so I am leaving it alone. (But I wonder if [partner_access](https://github.com/artsy/gravity/blob/af499450c88ec8816936ee2178b5a0cf2e2c15d4/config/initializers/_user_roles.rb#L11) role is more appropriate here.)

### DIFF
--- a/src/client/lib/setup/auth.coffee
+++ b/src/client/lib/setup/auth.coffee
@@ -9,6 +9,7 @@ passport = require 'passport'
 OAuth2Strategy = require 'passport-oauth2'
 User = require '../../models/user'
 Channel = require '../../models/channel'
+jwtDecode = require 'jwt-decode'
 
 setupPassport = ->
   passport.use 'artsy', new OAuth2Strategy
@@ -22,6 +23,7 @@ setupPassport = ->
       headers: 'X-Access-Token': accessToken
       error: (m, err) -> done err
       success: (user) ->
+        user.set 'roles', jwtDecode(accessToken).roles
         id = user.get('channel_ids').concat(user.get('partner_ids'))[0]
         id = process.env.DEFAULT_PARTNER_ID if not id and user.get('type') is 'Admin'
         new Channel(id: id).fetchChannelOrPartner
--- a/src/client/apps/edit/components/header/index.tsx
+++ b/src/client/apps/edit/components/header/index.tsx
@@ -20,7 +20,7 @@ interface Props {
   deleteArticleAction: () => void
   edit: Edit
   forceURL: string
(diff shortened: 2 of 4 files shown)


eessex (approved):
Looks great! My guess is that settings can also be a `team` role, this is how we add our team-members to different channels... so for example the editorial team would use to enable access for a new staff member. Could be handled here or as a separate pr.

src/client/lib/setup/auth.coffee
```diff
@@ -22,6 +23,7 @@ setupPassport = ->
       headers: 'X-Access-Token': accessToken
       error: (m, err) -> done err
       success: (user) ->
+        user.set 'roles', jwtDecode(accessToken).roles
```
joeyAghion, line 26: (Sorry for later review.)

Is this operation done on every request?

artsyjian: No worries. Let me check that.

artsyjian: Looks like it's done once per session:
```
positron_1                | GET /login 302 1.106 ms - 0
positron_1                | GET /api/users/me 200 132.749 ms - 164
positron_1                | setting user.roles
positron_1                | GET /api/channels/5fd6d1f9fa338d0e1d8ac143 200 4.184 ms - 101
```

PR #2908, opened 2021-01-12
Sanitize urls to limit redirect chains

ashleyjelks:
We ran this logic against the articles on staging, and want these checks to be included on any future articles that are created or existing ones that are updated, so committing these changes here.

### DIFF
--- a/src/api/apps/articles/model/sanitize.js
+++ b/src/api/apps/articles/model/sanitize.js
@@ -13,7 +13,6 @@ export const sanitizeLink = urlString => {
   } catch (_e) {
     url = new URL(`https://${urlString}`)
   }
-
   if (url.hostname === "artsy.net") {
     url.hostname = "www.artsy.net"
   }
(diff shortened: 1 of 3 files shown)


eessex (approved):
👍 Looks good - thanks for the improvement to backfill logging!

src/api/apps/articles/model/sanitize.js
```diff
@@ -19,6 +20,12 @@ export const sanitizeLink = urlString => {
   if (url.hostname.includes(".artsy.net")) {
     url.protocol = "https:"
   }
+  if (url.href.includes("artsy.net/post/")) {
+    url.href = url.href.replace("/post/", "/article/")
+  }
+  if (url.href.includes("/posts")) {
```
eessex, line 26: This looks like an unsafe replace-- we should use `"artsy.net/posts/"` like above to ensure we don't change non-artsy URLS

eessex (changes requested):
Actually one blocking comment!

src/api/apps/articles/model/sanitize.js
zephraph:
```suggestion
  if (url.href.includes("artsy.net/posts")) {
```

eessex (approved):

PR #2909, opened 2021-01-12
feat: replace admin role restriction with team for SETTINGS

artsyjian:
https://artsyproduct.atlassian.net/browse/PLATFORM-3048
https://artsy.slack.com/archives/CA8SANW3W/p1610051152084300

Access to SETTINGS (SETTINGS button on sidebar) is currently restricted based on `admin` role. Switch that to `team`. However, keep Channel Edit (SETTINGS->Channels->[Channel Name]) on `admin`. We intend for `team` to allow just enough privilege for article writing. Channel editing seems an access beyond that. On CHANNELS page (`SETTINGS->Channels`), say `admin role required` to make it obvious.

Update the access restriction on these routes which are called in SETTINGS:

`frontend` (mostly GET/POST)
/settings, /settings/curations, /settings/channels, /settings/tags, /settings/authors

`backend` (mostly PUT/POST/DELETE)
/api/authors, /api/curations, /api/tags

I made [similar](https://github.com/artsy/positron/pull/2893) change to Admin panel (the `Admin` tab when editing an article) and now see that some fields there query Gravity match endpoints which are restricted to `admin
` role:

`Article` section.
- Contributing Authors

`Featuring` section.
- Partners
- Shows
- Auctions

For now, we add `(admin role required)` to the field names to make it obvious, and disable the fields if user does not have `admin` role.

Also in Admin panel, found that toggling `Magazine Feed` to `yes` requires `admin` role in the [backend](https://github.com/artsy/positron/blob/28ca6c55d7258ed500e1864d7260eec3ad6be734/src/api/apps/articles/routes.coffee#L61). Switch to `team`.

Lastly, do the same for these backend routes. However, I don't know where they are actually called.
- /api/sections
- /api/verticals

### DIFF
--- a/src/client/apps/edit/components/admin/components/featuring/index.tsx
+++ b/src/client/apps/edit/components/admin/components/featuring/index.tsx
@@ -7,7 +7,7 @@ interface AdminFeaturingProps {
   isAdmin: boolean
 }
 
-export const AdminFeaturing = (props: AdminFeaturingProps) => {
+export const AdminFeaturing: React.SFC<AdminFeaturingProps> = props => {
   const { isAdmin } = props
   return (
     <div>
--- a/src/client/apps/edit/components/admin/components/featuring/index.tsx
+++ b/src/client/apps/edit/components/admin/components/featuring/index.tsx
@@ -3,16 +3,22 @@ import AutocompleteListMetaphysics from "client/components/autocomplete2/list_me
 import React from "react"
 import { FeaturingMentioned } from "./featuring_mentioned"
 
-export const AdminFeaturing = () => {
+interface AdminFeaturingProps {
+  isAdmin: boolean
+}
+
+export const AdminFeaturing = (props: AdminFeaturingProps) => {
+  const { isAdmin } = props
   return (
     <div>
       <Flex flexDirection={["column", "row"]}>
         <Box width={["100%", "50%"]} pr={[0, 2]}>
           <AutocompleteListMetaphysics
             field="partner_ids"
-            label="Partners"
+            label="Partners (admin role required)"
             model="partners"
             placeholder="Search by partner name..."
+            disabled={!isAdmin}
           />
         </Box>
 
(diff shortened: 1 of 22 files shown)


artsyjian:
@eessex , @izakp , and I looked over the UI changes during Platform Mob today. In regards to the pages/fields that are hidden from `team` role due to `admin` required on Gravity's side, we think it's better to show them but also show an error message that `admin` role is required. This will help inform Editorial of the restrictions and bring up for discussion of any further changes we should do prior to removing `admin` role from users.

artsyjian:
@damassi - This started out as a straight `admin` -> `team` swap, but looks like some UI changes are inevitable. I am proceeding as mentioned in the last comment. Let me know if you have any suggestions. I will hit you up if I need help!

damassi:
@artsyjian - If you rebase from `master` client-side dev (in more modern sections of the app) has been fixed. Also fixed a few other tool chain issues like linting, etc.

artsyjian:
@damassi - nice! thanks for cleaning it up. looks a lot nicer now!

artsyjian:
@damassi , @joeyAghion - This PR is good for review. Thanks!

src/client/apps/edit/components/admin/components/featuring/index.tsx
```diff
@@ -3,16 +3,22 @@ import AutocompleteListMetaphysics from "client/components/autocomplete2/list_me
 import React from "react"
 import { FeaturingMentioned } from "./featuring_mentioned"
 
-export const AdminFeaturing = () => {
+interface AdminFeaturingProps {
+  isAdmin: boolean
+}
+
+export const AdminFeaturing = (props: AdminFeaturingProps) => {
```
damassi, line 10: We'll want to use react-specific types for this (and elsewhere, if there are changes):

```tsx
export const AdminFeaturing: React.FC<AdminFeaturingProps> = props => { ... }
```

damassi (approved):
One tiny comment; otherwise LGTM 👍

src/client/apps/edit/components/admin/components/featuring/index.tsx
artsyjian: changed. thanks @damassi .

PR #2911, opened 2021-01-14
toolchain: Fix dev env by updating react-hot-loader

damassi:
Noticed that booting Positron in local dev was broken (seemingly since Jan 2020), but turns out the fix was easy -- we just needed to bump `react-hot-loader`'s version, per [these instructions](https://github.com/gaearon/react-hot-loader/issues/1227#issuecomment-482518698) 

cc @artsy/grow-devs


damassi:
cc @pepopowitz, if yr curious

PR #2915, opened 2021-01-15
bugfix: Fix issue where edit fails if sub req fails

damassi:
Addresses https://artsyproduct.atlassian.net/browse/GRO-147 

Ran into an issue where [this specific feature article](https://www.artsy.net/article/artsy-editorial-trends-watch-2021-return-nature) would now accept edits, but others worked just fine. Digging in, it turns out that there was an error around an artist parsed from the body (or hyperlinked) being removed from gravity, with its ID returning 404, which would cause the entire edit to fail. 

Since there's lots of automatic parsing and linking happening in Positron (and potentially many calls to grav for a given article), it would be better to gracefully fail, versus erroring out completely.

### DIFF
--- a/src/api/apps/articles/model/save.coffee
+++ b/src/api/apps/articles/model/save.coffee
@@ -14,6 +14,7 @@ Article = require './index'
 { ARTSY_URL, GEMINI_CLOUDFRONT_URL } = process.env
 artsyXapp = require('artsy-xapp')
 { sanitizeLink } = require "./sanitize.js"
+chalk = require 'chalk'
 
 @onPublish = (article, cb) =>
   unless article.published_at or article.scheduled_publish_at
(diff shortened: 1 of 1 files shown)


src/api/apps/articles/model/save.coffee
```diff
@@ -137,7 +138,17 @@ removeStopWords = (title) ->
             .set('X-Xapp-Token': token)
             .end callback
   async.parallel callbacks, (err, results) =>
-    return cb(err) if err
+    if err
+      console.log chalk.red('[ERROR: articles/model/save @ generateKeywords]'), err.response.error
```
damassi, line 142: Another problem around this, which is the reason it took so long, was because of all the callbacks, errors were getting consumed and passed all around the stack -- and when the error finally landed it contained no valuable info. `error.response.error` fills in details, like req url, error, and stack trace, making debugging much easier.

src/api/apps/articles/model/save.coffee
```diff
@@ -137,7 +138,17 @@ removeStopWords = (title) ->
...
+
+      # TODO: because there was an error saving keywords (eg artwork was
+      # unpublished, or artist deleted, or hyperlink incorrect), we return
+      # article as is, without updated keywords. It would be better if there
+      # was a way to surface this in the UI, either as an error state or as
+      # a validation.
```
damassi, line 148: Addressing this would be significant time-wise, I think. So added a `TODO`.

jonallured:
@damassi thanks for all the investigation that went into figuring this out!! 🙌

PR #2917, opened 2021-01-15
chore: fixes artsy urls with capitalized letters causing redirects

ashleyjelks:
Adds an additional check to the sanitize urls function to clean up redirects caused by capitalized letters in Artsy urls.

cc @artsy/purchase-devs 

e.g.
Before: `https://artsy.net/DutchPavilion`
After: `https://artsy.net/dutchpavilion`

### DIFF
--- a/src/api/apps/articles/model/sanitize.js
+++ b/src/api/apps/articles/model/sanitize.js
@@ -29,6 +29,9 @@ export const sanitizeLink = urlString => {
     const result = str.replace(regex, subst);
     url.href = result
   }
+  if (url.hostname.includes("artsy.net")) {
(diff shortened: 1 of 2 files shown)


src/api/apps/articles/model/sanitize.js
```diff
@@ -29,6 +29,9 @@ export const sanitizeLink = urlString => {
     const result = str.replace(regex, subst);
     url.href = result
   }
+  if (url.hostname.includes("artsy.net")) {
+    url.href = url.href.toLowerCase()
+  }
```
starsirius, line 34: Nice one!

PR #2934, opened 2021-04-13
feat: protect cronjobs from eviction by cluster-autoscaler, and general true-up.

artsyjian:
Collaboration with @mc-jones.

https://artsyproduct.atlassian.net/browse/PLATFORM-3200

- Set [safe-to-evict](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node) to false for crons so they cannot be evicted by cluster-autoscaler.
- Set [backoffLimit](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy) for crons to reduce retries.
- Generally, true up with hokusai [template](https://github.com/artsy/artsy-hokusai-templates/tree/master/nodejs).

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -65,7 +65,7 @@ spec:
           lifecycle:
             preStop:
               exec:
-                command: ["sh", "-c", "sleep 10"]
+                command: ["sh", "-c", "sleep 5"]
       dnsPolicy: ClusterFirst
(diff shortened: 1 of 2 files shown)


damassi:
👍

PR #2972, opened 2021-09-14
feat: turn-key env

ovasdi:
https://artsyproduct.atlassian.net/browse/PLATFORM-3401

This PR introduces _turk-key [development environments](https://github.com/artsy/README/blob/main/playbooks/development-environments.md)_ which:
- consolidated project setup into a script
- introduced `.env.oss` for common env vars:
  - moved most of `.env.example` into `.env.shared` to keep base `.env` minimal
- fix docker development:
  - parts of SALT var needed to be escaped for the stack to work locally
  - revert the `network_mode: host` back to `ports` for the stack to work on macs
- update how configs are loaded for local dev by introducing `loadenv.js` script
- update README

---

Tested by:
- setting up the project via script
- `yarn start`, `yarn dev`, `yarn test` (2 tests fail locally, not in CI)
- in docker (on a Mac) `COMMIT_HASH=$(git rev-parse --short HEAD hokusai dev start` and `hokusai test`


artsyjian (approved):
LGTM. Only thing is `setup.sh` exits with code `3` at `source ~/.nvm/nvm.sh` which I get around by switching script to `#!/bin/bash`.

.gitignore
```diff
@@ -28,3 +27,9 @@ coverage*
 .nyc_output/
 
 cypress.env.json
+Brewfile.lock.json
+
+# Do not ignore
+!.env.example
+!.env.test
+!.env.oss
```
icirellik, line 35: Nice, I love this coupled with `.env*`. ✨

src/lib/loadenv.js
```diff
@@ -0,0 +1,33 @@
...
+const { parse } = require("dotenv")
+
+function loadEnv(env) {
+  return checkFileExistsSync(env) ? parse(readFileSync(env)) : {}
+}
+
+function applyToEnv(config) {
+  if (Object.keys(config).length !== 0) {
+    for (const k in config) {
+      if (!process.env[k]) {
```
artsyjian, line 13: Since `.env.shared` is loaded before `.env`, I expect this conditional to prevent a var `FOO` in `.env` from overriding `FOO` in `.env.shared`, since by the time when `.env` is loaded, `process.env['FOO']` already exists (was loaded from `.env.shared`). But my testing shows that the overriding does work. Javascript magic?

src/lib/loadenv.js
```diff
@@ -0,0 +1,33 @@
+// @ts-check
```
icirellik, line 1: Given how widespread this helper has become we may want to ticket creating an npm module for it!

scripts/setup.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
+
+# This assumes you have general prerequisites installed as by:
+# https://github.com/artsy/potential/blob/master/scripts/setup
+
+# Exit if any subcommand fails
+set -e
+
+GREEN='\033[0;32m'
+NO_COLOR='\033[0m'
```
icirellik, line 10: No need to address this now, it may be worth using `tput` for color manipulation as it degrades nicely and won't output the color codes to unsupported terminals.

scripts/setup.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
```
icirellik, line 1: Was there any particular reason to use `sh` instead of `bash`?

icirellik (approved):
Looks good!

scripts/setup.sh
ovasdi: TBH I thought we preferred `sh` in scripts so I did this for the sake of consistency but after looking around it seems `bash` is used just as much if not more. In potential, force, pulse, aprd setup scripts we use `bash` so I am switching this to bash as its more feature rich and is used in many of our scripts.

scripts/setup.sh
ovasdi: Switched to `tput`. Thank you for bringing this up.

src/lib/loadenv.js
ovasdi: I believe you are exactly right, it is JS magic. Namely a result of how the spread operator works on objects literals.

```
let obj1 = { foo: 'bar', x: 42 };
let obj2 = { foo: 'baz', y: 13 };

let clonedObj = { ...obj1 };
// Object { foo: "bar", x: 42 }

let mergedObj = { ...obj1, ...obj2 };
// Object { foo: "baz", x: 42, y: 13 }
```
[SOURCE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#spread_in_object_literals)

ovasdi: @icirellik please correct me if im wrong :-)

artsyjian: i see. thanks for clarifying!

src/lib/loadenv.js
ovasdi: Sent you a slack about this.

src/lib/loadenv.js
icirellik: @ovasdi that's exactly right. The spread operator functions similarly to `Object.assign` where the right-most object overrides and keys assigned before it.

PR #2998, opened 2021-10-21
chore: allow kubernetes+cloudfare for ingress connections::PLATFORM-1308

sonecabr:
More info here: https://artsyproduct.atlassian.net/browse/PLATFORM-1308
This PR is a dependence of this [one](https://github.com/artsy/substance/pull/263)
As we are trying between services to happens internally, this change is required, as kubernetes IP's might be allowed for incoming connections


artsyjian (approved):
lgtm

PR #3001, opened 2021-10-27
feat: index only visible_to_public articles to algolia

gkartalis:
# Description

Ensure that we index only visible to public articles to algolia


anandaroop (approved):
Thanks for sticking with this 👍🏽

PR #3025, opened 2022-01-04
feat(script): Add `yarn dump-schema`

damassi:
This adds a new `package.json` script to positron, `yarn dump-schema`, allowing us to export (and sync) schema stuff for MP for stitching.

### DIFF
--- a/src/api/apps/graphql/index.js
+++ b/src/api/apps/graphql/index.js
@@ -107,7 +107,7 @@ const metaFields = {
 
 const ArticleSchema = object(Article.inputSchema).concat(object(metaFields))
 
-const schema = joiql({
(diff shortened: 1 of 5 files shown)


src/api/apps/graphql/index.js
```diff
@@ -146,6 +146,8 @@ const schema = joiql({
   },
 })
 
+module.exports.schema = schema
```
damassi, line 149: Weirdly this was needed to execute the script directly from node command line even though `babel-node` is being loaded before everything [here](https://github.com/artsy/positron/blob/91dcaee3c909685c7e3710302e152021c35e315d/package.json#L25)

PR #3047, opened 2022-04-21
feat: use latest version of artsy managed orbs, and track app version in k8s. (PLATFORM-4103)

artsyjian:
feat: use latest version of artsy managed orbs, and track app version in k8s. (PLATFORM-4103)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,9 +2,9 @@ version: 2.1
 
 orbs:
   codecov: codecov/codecov@1.2.5
-  hokusai: artsy/hokusai@0.10.0
-  yarn: artsy/yarn@6.1.0
-  horizon: artsy/release@0.0.1
+  hokusai: artsy/hokusai@volatile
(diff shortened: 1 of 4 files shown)


ovasdi:
@artsyjian this failure seems to be related to the improvements to [git protocol security](https://github.blog/2021-09-01-improving-git-protocol-security-github/) made permanent on March 15th. Specifically, that github is no longer accepting unencrypted Git protocol in this form: `git://`:

> No more unauthenticated Git
> On the Git protocol side, unencrypted git:// offers no integrity or authentication, making it subject to tampering. We expect very few people are still using this protocol, especially given that you can’t push (it’s read-only on GitHub). We’ll be disabling support for this protocol.

This project depends on [passport-oauth2](https://github.com/artsy/positron/blob/main/package.json#L111) a package that contains the no longer supported protocol form in its [deps](https://github.com/craigspaeth/passport-oauth2/blob/master/package.json#L36).

ovasdi (approved):
🦖

PR #3052, opened 2022-04-26
feat: upgrade datadog tracer

artsyjian:
To support [Datadog Deployment Tracking](https://docs.datadoghq.com/tracing/deployment_tracking/) feature. Going with 0.36.2 because it's the highest version in use by our Node apps.

PR #3054, opened 2022-04-27
fix: revert datadog tracer version due to suspected memory leak

artsyjian:
Revert https://github.com/artsy/positron/pull/3052

We suspect that the new version of datadog tracer causes memory leak.

https://artsy.slack.com/archives/C0HP61PUJ/p1651080383034169


ovasdi (approved):
LGTM

PR #3083, opened 2022-12-20
feat(FX-4495): RabbitMQ event when an article is published / un-published

nickskalkin:
- [Jira story](https://artsyproduct.atlassian.net/browse/FX-4482)

Emits a RabbitMQ message when an article is published.

## Deployment

- [ ] Create exchange and queue on staging
- [ ] Create exchange and queue on production
- [ ] Create RabbitMQ user & add env variables (RABBITMQ_URL and ENABLE_PUBLISH_RABBITMQ_EVENTS) on staging
- [ ] Create RabbitMQ user & add env variables (RABBITMQ_URL and ENABLE_PUBLISH_RABBITMQ_EVENTS) on production

### DIFF
--- a/src/api/apps/articles/model/save.coffee
+++ b/src/api/apps/articles/model/save.coffee
@@ -1,7 +1,7 @@
 _ = require 'underscore'
 _s = require 'underscore.string'
 db = require '../../../lib/db'
-ampq = require '../../../lib/amqp'
+{ amqp } = require '../../../lib/amqp'
 stopWords = require '../../../lib/stopwords'
 async = require 'async'
 moment = require 'moment'
@@ -18,17 +18,19 @@ artsyXapp = require('artsy-xapp')
 { sanitizeLink } = require "./sanitize.js"
 chalk = require 'chalk'
 { cloneDeep } = require 'lodash'
+{ ObjectId } = require "mongojs"
 
 @onPublish = (article, cb) =>
   unless article.published_at or article.scheduled_publish_at
     article.published_at = new Date
-  @generateSlugs article, cb
-  @enqueuePublishEvent article
+  @generateSlugs article, (err, article) =>
+    @enqueuePublishEvent article
+    cb null, article
 
 @onUnpublish = (article, cb) =>
   @generateSlugs article, (err, article) =>
+    @enqueueUnublishEvent article
     cb null, article
-  @enqueueUnublishEvent article
 
 setOnPublishFields = (article) =>
   article.email_metadata = article.email_metadata or {}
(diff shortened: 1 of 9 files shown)


src/api/apps/articles/model/save.coffee
```diff
@@ -17,11 +18,21 @@ artsyXapp = require('artsy-xapp')
...
+Channel = require '../../channels/model'
 
 @onPublish = (article, cb) =>
   unless article.published_at or article.scheduled_publish_at
     article.published_at = new Date
-  @generateSlugs article, cb
+
+  self = @
+  @generateSlugs article, (err, article) =>
+    if process.env.ENABLE_PUBLISH_RABBITMQ_EVENTS == 'true'
```
dblandin, line 29: **suggestion**: I feel like it's getting a bit confusing to have this behavior injected here in the middle of the publishing process. As is, this code would run (and the message emitted) after slugs are generated but _before_ the article is saved.

What do you think of moving this logic up [here](https://github.com/artsy/positron/blob/5d313c74b5ff4dcbfd1c32f10ddc0402f8706e4f/src/api/apps/articles/model/index.js#L166) so that we can chain this behavior after the modified article is persisted:

```javascript
if (publishing) {
  return onPublish(modifiedArticle, (err, article) => {
    // at this point, all of the publishing modifications have been set but not persisted
    sanitizeAndSave((err, article) => {
      if (err) {
        // if an error occurs here, we would not want to emit the message
        return callback(err)
      } else {
        // successfully persisted. emit message.
      }
    })(err, article)
  })
}
```

dblandin: We might also be able to leverage the ArticleModel to utilize some of its helper methods:

```javascript
model = new ArticleModel(cloneDeep article)
model.dispatchPublishEvent()
```

```coffeescript
module.exports = class Article extends Backbone.Model

  dispatchPublishEvent: ->
    if @isVisibleToPublic() and ENABLE_PUBLISH_RABBITMQ_EVENTS == 'true'
      payload =
        id: @get('id')
        featured_artist_ids: @get('primary_featured_artist_ids')
        href: @href()
        title: @get('title')

      amqp.publish("editorial", "article.published", payload)
      
```

nickskalkin: nice suggestion, thanks

src/api/apps/articles/model/save.coffee
```diff
@@ -1,6 +1,7 @@
 _ = require 'underscore'
 _s = require 'underscore.string'
 db = require '../../../lib/db'
+{ amqp } = require '../../../lib/amqp'
```
anandaroop, line 4: **question (non-blocking):** Minor — are the imports on L4 and L21 necessary in this file?

(I think there was a refactor since these were added)

anandaroop (approved):

src/api/lib/amqp.js
```diff
@@ -0,0 +1,22 @@
...
+        if (err) throw err
+
+        ch.assertExchange(exchange, "topic")
+        ch.publish(exchange, routingKey, Buffer.from(JSON.stringify(message)))
+      })
+    }
+  )
+}
+
+module.exports = { amqp }
```
dblandin, line 22: **suggestion / non-blocking**: As this is a new file, might be nice to have it in TypeScript.

dblandin (approved):
LGTM! Nice refactoring 👍🏼

src/api/lib/amqp.js
nickskalkin: will do in a separate PR this week!

PR #3110, opened 2023-06-12
Disable Algolia search

nickskalkin:
Removes Algolia integration completely

### DIFF
--- a/src/api/test/models/article.test.coffee
+++ b/src/api/test/models/article.test.coffee
@@ -87,7 +87,7 @@ describe "Article", ->
 
     it 'creates a lower score for an older article', ->
       @article.set 'published', true
-      @article.set 'published_at', new Date(2016, 4, 10)
+      @article.set 'published_at', new Date(2022, 4, 10)
       @article.searchBoost().should.be.below(1000)
       @article.searchBoost().should.be.above(100)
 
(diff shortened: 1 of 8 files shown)


src/api/test/models/article.test.coffee
```diff
@@ -87,7 +87,7 @@ describe "Article", ->
 
     it 'creates a lower score for an older article', ->
       @article.set 'published', true
-      @article.set 'published_at', new Date(2016, 4, 10)
+      @article.set 'published_at', new Date(2022, 4, 10)
```
anandaroop, line 90: The one spec failure was legit, based on this hard-coded date. 

I just bumped the date instead of doing anything more clever, and the test passes now.

anandaroop (approved):
LGTM, with the spec fix below.

I'll merge and try to QA this on staging.

anandaroop:
There is a newer and weirder failure in CI which can _not_ be reproduced locally:

https://app.circleci.com/pipelines/github/artsy/positron/5283/workflows/ecd846cc-7028-47d6-ba49-5feff0ed2060/jobs/11510

I'm inclined to merge this PR anyway and see if there is any negative effect. If so we can revert.

anandaroop:
Going for it…

PR #3120, opened 2023-11-07
fix: uses unica for og:image generation

dzucconi:
Closes [DIA-292](https://artsyproduct.atlassian.net/browse/DIA-292)

Sort of just blindly changing to use Unica here. Not sure what this looks like as I can't get Positron to install/run but this should be _fine_.

[DIA-292]: https://artsyproduct.atlassian.net/browse/DIA-292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jonallured (approved):
LGTM!! 🎉

PR #3126, opened 2024-01-03
feat: replace team role with editorial role 

egdbear:
This pull request removes the 'team' and 'admin' roles and replaces them with the 'editorial' role.

[DIA-230]

[Gravity PR](https://github.com/artsy/gravity/pull/17223) for enabling the Editorial roles

[DIA-230]: https://artsyproduct.atlassian.net/browse/DIA-230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

@artsy/diamond-devs

### DIFF
--- a/src/client/apps/settings/index.coffee
+++ b/src/client/apps/settings/index.coffee
@@ -8,14 +8,14 @@ routes = require './routes'
 app = module.exports = express()
 app.set 'views', __dirname + '/templates'
 app.set 'view engine', 'jade'
-{ teamOnly, adminOnly } = require '../../lib/middleware.coffee'
+{ editorialOnly } = require '../../lib/middleware.coffee'
 
-app.get '/settings', teamOnly, routes.index
-app.get '/settings/curations', teamOnly, routes.curations
-app.get '/settings/curations/:id/edit', teamOnly, routes.editCuration
-app.post '/settings/curations/:id', teamOnly, routes.saveCuration
-app.get '/settings/channels', teamOnly, routes.channels
-app.get '/settings/channels/:id/edit', adminOnly, routes.editChannel
-app.post '/settings/channels/:id', adminOnly, routes.saveChannel
-app.get '/settings/tags', teamOnly, routes.tags
-app.get '/settings/authors', teamOnly, routes.authors
+app.get '/settings', editorialOnly, routes.index
+app.get '/settings/curations', editorialOnly, routes.curations
+app.get '/settings/curations/:id/edit', editorialOnly, routes.editCuration
+app.post '/settings/curations/:id', editorialOnly, routes.saveCuration
+app.get '/settings/channels', editorialOnly, routes.channels
+app.get '/settings/channels/:id/edit', editorialOnly, routes.editChannel
+app.post '/settings/channels/:id', editorialOnly, routes.saveChannel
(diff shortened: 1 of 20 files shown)


joeyAghion (commented):
**Suggested migration:** This will block Positron's existing users from certain actions unless we grant them the new role prior to releasing this. We can query recent log-ins (like [here](https://artsyproduct.atlassian.net/browse/AMBER-281)) to identify the accounts that should be considered for the new role, then confirm with the Editorial folks.

**Potential bug:** Some of these `adminOnly` cases may depend on back-end APIs that _also_ require the `admin` role, like to search users. Does Gravity need changes to support the new role there?

> I couldn't create efficient testing environment...

Let's fix this! At worst, you could copy down Positron's data locally, or just connect to staging's db.

src/client/apps/settings/index.coffee
```diff
@@ -8,14 +8,14 @@ routes = require './routes'
...
 app.set 'view engine', 'jade'
-{ teamOnly, adminOnly } = require '../../lib/middleware.coffee'
+{ editorialOnly } = require '../../lib/middleware.coffee'
 
-app.get '/settings', teamOnly, routes.index
-app.get '/settings/curations', teamOnly, routes.curations
-app.get '/settings/curations/:id/edit', teamOnly, routes.editCuration
-app.post '/settings/curations/:id', teamOnly, routes.saveCuration
-app.get '/settings/channels', teamOnly, routes.channels
-app.get '/settings/channels/:id/edit', adminOnly, routes.editChannel
```
egdbear, line 18: This could be also change to editorial role. @joeyAghion

egdbear:
>Suggested migration: This will block Positron's existing users from certain actions unless we grant them the new role prior to releasing this. We can query recent log-ins (like [here](https://artsyproduct.atlassian.net/browse/AMBER-281)) to identify the accounts that should be considered for the new role, then confirm with the Editorial folks. 

I was wondering where to add those emails. I will add the account emails to the ticket.

> Potential bug: Some of these admin-only cases may depend on back-end APIs that also require the admin role, like to search users. Does Gravity need changes to support the new role there?

`/search` - [under the hood](https://github.com/artsy/positron/blob/bafc8e4dbc1a963c30b0a1a6adb545f5f26230e4/src/api/lib/elasticsearch.coffee) is calling an ES endpoint. I believe we don't have any authorization set for that?.

`/api/v1/me` - is being called when using the `/users` endpoint.

egdbear:
Possibly because M1, asdf and outdated SSL (due to old ruby versions) I couldn't run the app.

To run locally connected to staging I had to update the `scripts/start.sh` 

```
# !/usr/bin/bash

set -e -x

forever -c 'node --max_old_space_size=1024 --no-experimental-fetch --openssl-legacy-provider' ./src/index.js --colors
```

joeyAghion:
I haven't followed the latest changes closely enough to know when this can be merged, so I'll approve and add you to assignees. Let me know if you can use more feedback!

joeyAghion (approved):

mc-jones:
🙏

PR #3128, opened 2024-01-23
chore: remover Cinder references

amonkhouse:
Replacing Cinder references RE the articles sitemap with Ember references.

I'm going to add a ticket to the backlog to add information on how to add/change sitemaps because I don't think we have that anywhere for the new ones, and these docs referenced some Cinder docs.


cavvia (approved):
Marvellous

PR #3131, opened 2024-01-23
chore: prepare for mongodb version 5.0

mc-jones:
This PR:

1. Updates the `test.yml` and `development.yml` yamls to use `mongodb@5.0.23`. 
2. removes the `mongojs` library and replaces it with the native `mongodb` driver. It also makes the necessary changes to CRUD operations.

### Problem
On February 29th (leap year), MongoDB will auto-upgrade our Atlas clusters to `5.0` due to version `4.4` being at its EOL date. Positron is currently using version `4.4`.

### Solution
To avoid unnecessary disruptions, this PR updates `positron` to use version `5` in its `development` and `test` environments. It also replaces the `mongojs` library, which is [unmaintained](https://github.com/mongo-js/mongojs), with the native `mongodb` driver at version `4.10`. 

The `4.10` driver is compatible up to `mongodb@6.0` (inclusive).

### Migration
We will be bumping the `staging` atlas cluster and monitoring affected services for about a week. If there are no issues, we will bump the `production` cluster. 

### Considerations
- Most of the work was updating the CRUD operations to use the `mongodb` operations. However, the operation functions are not 1:1. For example, `mongodb` does not have a `save` function. I've used the tests to guide the work and pick the correct operation, but we should spot-check the work to make sure no incorrect assumptions were made about the business logic.
- I toyed with bumping `node` at the same time, as I don't think it will require too many changes. But I didn't want to do too much at once, so I may follow up with a `node` bump once this is in `production`. 

[PHIRE-544]

[PHIRE-544]: https://artsyproduct.atlassian.net/browse/PHIRE-544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/api/apps/articles/model/save.coffee
```diff
@@ -160,13 +160,11 @@ removeStopWords = (title) ->
...
-    indexForSearch(article, ->) if article.indexable
-    db.articles.save sanitize(article), callback
-  else
-    indexForSearch(article, ->) if article.indexable
-    db.articles.save sanitize(article), callback
+  article = setOnPublishFields article if article.published or article.scheduled_publish_at
+  indexForSearch(article, ->) if article.indexable
+  inputData = _.omit(article, '_id')
+  db.collection('articles').updateOne { _id: article._id }, { $set: sanitize(inputData)}, { upsert: true }, (err, res) ->
+    db.collection('articles').findOne { _id: article._id || res.upsertedId }, callback
```
mc-jones, line 167: From the tests my understanding is this function will need to support document creation and updating, hence the `upsert: true` option. If the passed in `article` has an `_id`, it will update the existing document, otherwise it will create a new one. 

I want to point out that the test documents have both an `_id` and `id` and they are often different (I confirmed this is true in the current main branch, also). I assume that this is a result of our test setup. To not expand the scope of this PR, I haven't addressed this and always assume the `_id` is the correct primary key.

src/api/apps/sessions/model.js
```diff
@@ -68,12 +68,19 @@ export const save = (inputData, callback) => {
       _id: id,
     }
 
-    db.sessions.save(data, callback)
+    db.collection("sessions").updateOne(
+      { _id: data._id },
+      { $set: inputData },
+      { upsert: true },
```
mc-jones, line 74: If someone with better knowledge can confirm that an `upsert` here is correct, it would be appreciated. Otherwise I will consider the passing tests enough validation.

src/api/lib/db.coffee
```diff
@@ -1,26 +1,20 @@
...
-
-mongojs = require 'mongojs'
+{ MongoClient } = require 'mongodb'
 fs = require 'fs'
 path = require 'path'
 { MONGOHQ_URL } = process.env
 debug = require('debug') 'api'
 
-collections = ['articles', 'users', 'sections', 'artists', 'curations', 'channels', 'tags', 'verticals', 'authors', 'sessions']
-db = mongojs MONGOHQ_URL, collections
```
mc-jones, line 13: This `collections` array appears to be necessary for the previous library `mongojs`. It seems fine to remove.

src/api/test/helpers/db.coffee
```diff
@@ -7,20 +7,29 @@ bcrypt = require 'bcryptjs'
...
 
-@fabricate = (collection, data, callback) =>
+@fabricate = (collectionName, data, callback) =>
+  collection = @db.collection(collectionName)
   if _.isArray(data)
-    data = (fixturize collection, obj for obj in data)
+    data = (fixturize collectionName, obj for obj in data)
+    return collection.insertMany data, (err, res) ->
+      collection.find({}).toArray(callback)
+  if collectionName == 'sessions'
```
mc-jones, line 16: The `mongojs` implementation of `insert` seemed to accept a single document and an array of documents. This is not the case for the native `mongojs` driver, which needs us to be more specific and use `insertOne` or `insertMany`. The test setup for `sessions` seems to be the only one that passes in an array of documents as the argument.

hokusai/test.yml
```diff
@@ -23,7 +23,7 @@ services:
     ports:
       - "9200:9200"
   positron-mongodb:
-    image: mongo:4.4
+    image: mongo:5.0.23
```
mc-jones, line 26: Matches the gravity [version](https://github.com/artsy/gravity/pull/17240).

src/client/apps/articles_list/test/routes.test.ts
```diff
@@ -1,3 +1,7 @@
+/**
+ * @jest-environment node
+ */
+
```
mc-jones, line 4: The few files where I've set the `jest` environment to `node` are a result of `js-dom` not having a `TextEncoder` function, which caused `mongodb` to break on a `TypeError`. This was a tricky one to debug but I found a anser that pointed to the root of the cause in a comment from a `jest` maintainer, [here](https://github.com/jsdom/jsdom/issues/2524#issuecomment-902027138). 

Please tell me if you know of any reason this should not be done.

mc-jones:
I'm not sure what to do with the blocking codecove issues as nothing has changed about test coverage.

src/api/apps/sessions/model.js
joeyAghion: Not knowledgeable, but this seems reasonable.

joeyAghion (approved):
Epic work. Over enough time, the preferred high-level language changed from coffee to ts, the mongojs library became unmaintained, the primitives changed from arrays to cursors, the node runtime needs upgrading.... This PR has it all!

mc-jones:
Oh shoot. Forgot the co-authored by @ovasdi!

hokusai/development.yml
```diff
@@ -19,7 +19,7 @@ services:
     ports:
       - 3005:3005
   positron-elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:5.6.16
+    image: docker.elastic.co/elasticsearch/elasticsearch:6.8.23
```
ovasdi, line 22: *thought* : perhaps a good opportunity to switch to using the artsy [multi-arch elasticsearch image](https://hub.docker.com/layers/artsy/elasticsearch/6.8.23/images/sha256-11ad2ab19b7d65b7f5c61dff914abd865749940b437dbff8211674c82a8831c0?context=explore)?

mc-jones: Good call! Switched in [22b08ac](https://github.com/artsy/positron/pull/3131/commits/22b08acd788c39ba0191e66b2dca765566e2c7f5)

amonkhouse:
random comment to say that's a v high quality and descriptive pr description

egdbear (approved):
Looks good!

mc-jones:
@ovasdi - I've taken the draft status off, so merge at will. The codecov failure is not helpful in this instance as there have been no changes to code coverage. Discussed a bit [here](https://artsy.slack.com/archives/C05EEBNEF71/p1706129453715949?thread_ts=1706122829.495369&cid=C05EEBNEF71) 🔒 .

ovasdi:
> @ovasdi - I've taken the draft status off, so merge at will. The codecov failure is not helpful in this instance as there have been no changes to code coverage. Discussed a bit [here](https://artsy.slack.com/archives/C05EEBNEF71/p1706129453715949?thread_ts=1706122829.495369&cid=C05EEBNEF71) 🔒 .

Re-running failed check to see what happens. Mostly curious.

ovasdi:
> > @ovasdi - I've taken the draft status off, so merge at will. The codecov failure is not helpful in this instance as there have been no changes to code coverage. Discussed a bit [here](https://artsy.slack.com/archives/C05EEBNEF71/p1706129453715949?thread_ts=1706122829.495369&cid=C05EEBNEF71) 🔒 .
> 
> Re-running failed check to see what happens. Mostly curious.

Re-running the failed check does not seem to be working.

PR #3134, opened 2024-01-26
fix: articles and sessions save functions

mc-jones:
This PR patches #3131 after testing on staging exposed a regression. 

### Problem
Users were unable to create or edit articles as a bug in the save logic meant that documents were being persisted with a `_id` of `null` when `upserted`.

### Solution
Change the implementation of the `articles` and `sessions`' `save` method to check for an existing `_id` of the object to be persisted, use `updateOne` when it exists, and use `insertOne` (instead of an `upsert` on a `updateOne`) when it doesn't. 

### Considerations
- I**n the existing implementation (i.e. pre #3131), the save method accepts explicitly setting an `_id` value for documents rather than always relying on `mongodb` to generate them upon creation. So, for example, if I called `Article.save` while passing an object with a key-value pair of `_id: 123`, it would respect this and persist an `article` with `123` as the `_id`. It would also except `null` or `undefined` as values, which is what is causing the bug.** 
- With refactors of this kind, I aim to not change test implementations as they can easily result in false positives, however, it felt warranted in this case. My read of the code is that the tests I have modified were relying on this behavior as a test convenience rather than an assertion that it would explicitly set the `_id` field with whatever value is passed in. 

If I am missing something in the code, or using bad assumptions, then please let me know.

### DIFF
--- a/src/api/apps/articles/model/save.coffee
+++ b/src/api/apps/articles/model/save.coffee
@@ -162,9 +162,13 @@ removeStopWords = (title) ->
   return callback err if err
   article = setOnPublishFields article if article.published or article.scheduled_publish_at
   indexForSearch(article, ->) if article.indexable
-  inputData = _.omit(article, '_id')
-  db.collection('articles').updateOne { _id: article._id }, { $set: sanitize(inputData)}, { upsert: true }, (err, res) ->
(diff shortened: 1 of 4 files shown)


ovasdi (approved):
LGTM

PR #3135, opened 2024-01-28
fix: model persistence

mc-jones:
This PR is a follow-up to #3134 refactors the `save` method of the `authors`, `channels`, `curations`, `sections`, `tags`, and `users` models to support `update`, not just `insert`, operations. There were no assertions about updating in the previous test suite, so these have been added.

### DIFF
--- a/src/api/apps/users/model.coffee
+++ b/src/api/apps/users/model.coffee
@@ -52,7 +52,7 @@ jwtDecode = require 'jwt-decode'
 save = (user, accessToken, callback) ->
   async.parallel [
     (cb) ->
-      db.collection('users').findOne {_id: ObjectId(user.id)}, cb
+      db.collection('users').findOne {user_ids: ObjectId(user.id)}, cb
     (cb) ->
       bcrypt.hash accessToken, SALT, cb
   ], (err, results) ->
--- a/src/api/apps/users/model.coffee
+++ b/src/api/apps/users/model.coffee
@@ -61,16 +61,16 @@ save = (user, accessToken, callback) ->
     user.partner_ids = _.map partner_ids, ObjectId
     user.channel_ids = _.pluck results[0], '_id'
     encryptedAccessToken = results[1]
-    db.collection('users').insertOne {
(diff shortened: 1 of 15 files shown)


src/api/apps/users/model.coffee
```diff
@@ -61,16 +61,16 @@ save = (user, accessToken, callback) ->
...
       email: user.email
       type: user.type
       access_token: encryptedAccessToken
       partner_ids: user.partner_ids
       channel_ids: user.channel_ids
-    }, (err, res) ->
-      db.collection('users').findOne {_id: res.insertedId}, callback
+    }
+    db.collection('users').updateOne { _id: ObjectId(user.id) }, { $set: data }, { upsert: true }, (err, res) ->
+      db.collection('users').findOne {_id: res.upsertedId || user._id}, callback
```
mc-jones, line 73: Upsert is correct here because we're guaranteed a unique mongoId from `gravity`.

src/api/apps/users/model.coffee
```diff
@@ -52,7 +52,7 @@ jwtDecode = require 'jwt-decode'
 save = (user, accessToken, callback) ->
   async.parallel [
     (cb) ->
-      db.collection('users').findOne {_id: ObjectId(user.id)}, cb
+      db.collection('users').findOne {user_ids: ObjectId(user.id)}, cb
```
ovasdi, line 55: Could you elaborate a bit about this change? I am unclear about the change from `_id` to `user_ids` for the user lookup by id.

ovasdi (approved):
LGTM
Just one non blocking question.

src/api/apps/users/model.coffee
mc-jones: This was actually the original logic which was mistakenly [changed](https://github.com/artsy/positron/commit/fada33157ff1d80afe36cd307595c9112834bb84#diff-47ba709974806d6b0cb616d465bfb8a8f3e7d5dca78f2afdcfd0c38a92407e77L55) in #3131. This just corrects that change.

PR #3138, opened 2024-01-29
fix: use find instead of findOne for user collections

mc-jones:
Patch #3135 to use the correct operator when setting up the `channel_ids` on the `user`.


ovasdi:
🙌

PR #3140, opened 2024-02-07
fix: use collection.countDocuments instead of cursor.count

mc-jones:
This patches #3131 to use `collection.countDocuments` instead of  `cursor.count` which has a different behavior in the native `mongodb@4.10` driver than the previously used `mongojs` library. Additionally, [support it is deprecated](https://my.papertrailapp.com/groups/3675843/events?q=program%3Apositron+cursor.count).

`cursor.count()` will return the count of the artworks returned from the `find` cursor. However, what our clients are expecting is the total number of documents that will be returned from a passed-in query. This currently breaks pagination, as flagged by the editorial team 🔒 [here](https://artsy.slack.com/archives/C03N12SR0RK/p1707168354368809).

```javascript
const cursor = db
    .collection("articles")
    .find(query)
    .skip(offset || 0)
    .sort(sort)
    .limit(limit)
    
    // Will return the # of documents returned from the above query, factoring in the `limit`.  
    // If the value of `limit` is `15`, it will return `15` (or less if there are not `15` documents matching the query). 
    return cursor.count(cb)
    
    // Will return the total # of documents matching the query in the whole collection 
    return db.collection("articles").countDocuments(query, cb)
   ```

### DIFF
--- a/src/api/apps/articles/model/index.js
+++ b/src/api/apps/articles/model/index.js
@@ -60,7 +60,7 @@ export const mongoFetch = (input, callback) => {
         if (!count) {
           return cb()
         }
-        return cursor.count(cb)
+        return db.collection("articles").countDocuments(query, cb)
       },
     ],
     (err, results) => {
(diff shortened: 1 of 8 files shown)


joeyAghion:
Good catch!

PR #3161, opened 2024-12-09
chore: use vault for secrets

mc-jones:
Load secrets from Vault, per [playbook](https://www.notion.so/artsy/Migration-of-Artsy-Projects-Secrets-07e5ce9e1ed044dba8d4b7c3f0ae1ccf#13fcab0764a080e2be4df5529deb7838).

[PHIRE-831]

[PHIRE-831]: https://artsyproduct.atlassian.net/browse/PHIRE-831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3165, opened 2024-12-11
fix: load secrets in data-sync cron containers

mc-jones:
Yet another patch #3161. We need to make the script available to the data sync cron jobs. This follows the pattern set out in https://github.com/artsy/artsy-unleash/pull/15 to store the script in a configMap and load into a volume mounted on the cronjob.


ovasdi (approved):
LGTM

PR #3167, opened 2025-01-14
refactor: use load-secrets-and-run script from updated docker image

mc-jones:
This PR removes the use of a `configMap` to store and mount the `load-secrets-and-run.sh` script. Once [these updated images](https://github.com/artsy/docker-images/pull/113) are pushed to our docker hub, we can rely on the script included in the containers. 

I'll need to sequence the roll out, so am assigning to myself.

### Related PR(s)
- https://github.com/artsy/docker-images/pull/113

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -305,10 +305,6 @@ spec:
             - name: secrets
               mountPath: /secrets
               readOnly: true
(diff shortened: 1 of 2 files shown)


artsyjian (approved):
lgtm

PR #3171, opened 2025-04-03
fix: configurable elasticsearch index suffix

nickskalkin:
**Issue:**

Gravity is using the `production` Elasticsearch index suffix even on staging (for instance, we use `articles_production`, `artworks_production` and so on, when performing ES operations). It seems that when we switched Gravity from using the `staging` suffix to `production` (happened some months ago), we forgot to make the corresponding change in Positron.

**Impact:**

Positron is still indexing articles into the `articles_staging` index, while search operations (via Gravity) are targeting the `articles_production` index. As a result, newly indexed articles are not found during search. This bug affects staging only.

# Pre-deploy

- [ ] Set `production` suffix on staging and prod


anandaroop (approved):
Thanks 🙏🏽 

I think one more related chore will be to update the .env.shared file on S3 with the new ENV var as well (in case that wasn't on your radar already)

PR #3174, opened 2025-05-09
chore: remove data sync tasks

amonkhouse:
These syncs are now handled in the [MongoDB sync step in the weekly staging data sync DAG in motion](https://github.com/artsy/motion/blob/main/dags/staging_data_sync/pipeline.py#L43-L57) so can be removed here.

The frequency will go down from once a day to once per week, but this app is not being actively developed so should be fine.


cavvia (approved):
Looks gutenberg to me. Weekly sync sounds fine for positron unless we hear vocal objections from our esteemed colleagues.

PR #3176, opened 2025-05-16
chore: Retire Jon

jpotts244:
It comes with a heavy heart but its time to remove Jon from Positron so he isnt slammed with notifications


.github/dependabot.yml
```diff
@@ -7,7 +7,6 @@ updates:
     # Limit to 0 to enable only security updates:
     open-pull-requests-limit: 0
     assignees:
-      - jonallured
```
jpotts244, line 10: Anyone want to volunteer from diamond here? I'll add

egdbear:
😭

PR #3180, opened 2025-06-16
fix: make params that can throw errors optional

iskounen:
In Sentry I found that Positron is throwing 50K errors per week, and it happens because some clients are requesting articles without a "sort" or "featured" parameter.

After looking at how our clients make these requests, I noticed that almost all of them are requesting articles with a `sort` of `PUBLISHED_AT_DESC`, and only some request `featured` articles.

Therefore, I have added a default value to the validation schema for the `sort` param and made `featured` an optional param to stop these errors.

cc: @artsy/diamond-devs


egdbear:
👏

PR #3203, opened 2025-07-18
feat(dia-1401): supports querying articles by inclusion in author_ids

dzucconi:
Writer already supports querying by `author_id` but weirdly that's not the one we use to display the author of the article... which is: `author_ids` 🤷 

cc @artsy/diamond-devs


dzucconi:
We do not need codecov in this repo...

PR #3212, opened 2025-07-24
docs: point people

ovasdi:
This is the only criticality 3 project without assigned point persons. Going by contributions (which I know are not many) you two are at the top over last 24mo. Hope its OK to add you both.


ovasdi:
Thanks both!

PR #3222, opened 2025-08-14
chore: deny claude access to remote commands

joeyAghion:
chore: deny claude access to remote commands


anandaroop:
Not sure if the test failure is flake or not, but it's surely unrelated to this change, so I'll force-merge…

PR #3228, opened 2025-08-19
fix: reindex_articles script

egdbear:
This PR enhances the article reindexing script (scripts/reindex_articles.coffee) by adding logging and improving error handling. The previous version provided minimal feedback during execution, which made it difficult to monitor progress and troubleshoot failures. 

Screenshot 2025-08-19 at 12 33 32

/cc @artsy/diamond-devs


scripts/reindex_articles.coffee
```diff
@@ -5,32 +5,115 @@ path = require 'path'
...
+
+      new Promise (resolve, reject) ->
+        processedCount = 0
+        errorCount = 0
+
+        asyncLib.mapSeries articles, (article, cb) ->
+          processedCount++
+          console.log("[#{processedCount}/#{articles.length}] Indexing article: #{article._id} (#{article.title || 'No title'})")
+
+          indexWorker article, (err) ->
```
mc-jones, line 55: Bit confused, are we duplicating the same logging below as well as the `indexWorker` function? Seems like we will log successes and errors for each run, twice?

scripts/reindex_articles.coffee
```diff
@@ -5,32 +5,115 @@ path = require 'path'
...
-    console.log('Completed indexing ' + results.length + ' articles.')
-  )
+# Log Elasticsearch configuration
+console.log('Elasticsearch Configuration:')
+console.log('  - URL:', process.env.ELASTICSEARCH_URL || process.env.OPENSEARCH_URL || 'NOT CONFIGURED')
+console.log('  - Index Name:', search.index)
+console.log('  - Index Suffix:', process.env.ELASTICSEARCH_INDEX_SUFFIX || 'production')
+console.log('  - Using OpenSearch:', search.isOpenSearch ? 'Yes' : 'No')
+
+main = ->
```
mc-jones, line 28: The multiple `.then` chaining makes this a bit tough to read. Unless it isn't supported for some reason, may be easier to read if we used `async/await`.

egdbear: I’m also curious how this script worked before, since async/await support was only introduced in CoffeeScript 2, and our current version is `"coffeescript": "1.11.1"`?

PR #3237, opened 2025-08-28
feat: upgrade OpenSearch v1.3.20 to 2.17.0

narikazu:
# Summary

The PR upgrades OpenSearch v.1.3.20 to 2.17.0. There is no code change we need to introduce when upgrading OS to 2.17.

# Ticket
https://artsyproduct.atlassian.net/browse/PHIRE-2126


mc-jones:
I believe we still need to make some code changes? For example, looks like we're still are using the elasticsearch [client](https://github.com/artsy/positron/blob/main/package.json#L85)(also [here](https://github.com/artsy/positron/blob/197404bf5152d2835af9a1630ddb46c39922d97b/src/api/lib/elasticsearch.coffee#L1)). Should have caught this on the first upgrade.

narikazu:
> I believe we still need to make some code changes? For example, looks like we're still are using the elasticsearch [client](https://github.com/artsy/positron/blob/main/package.json#L85)(also [here](https://github.com/artsy/positron/blob/197404bf5152d2835af9a1630ddb46c39922d97b/src/api/lib/elasticsearch.coffee#L1)).

As I mentioned [here](https://github.com/artsy/gravity/pull/19305#issuecomment-3306761461)
, I’d recommend keeping the elasticsearch gem for now and switching to the opensearch gem once the OpenSearch clusters are fully consolidated, since the elasticsearch gem works fine with OpenSearch v2.17.0.

PR #3242, opened 2025-10-01
feat(DIAM-81): Update author page URLs to have name rather than author ID

lilyfromseattle:
@artsy/diamond-devs 

This is part of a 3 repo change (there are PRs in metaphysics and force) to support author page urls. There is a backfill script I need to run but wanted to get feedback here first.

DESCRIPTION
    We need to update author page URLs to use the author’s name instead of the ID, so that URLs are more descriptive, user-friendly, and SEO-friendly.

ACCEPTANCE CRITERIA
    All new and existing author pages use URLs in the format: https://www.artsy.net/articles/author/<author-name> instead of https://www.artsy.net/articles/author/<author-id>.

    Redirects from old ID-based URLs to the new name-based URLs are implemented to avoid broken links.

NOTES
    Current example
        https://www.artsy.net/articles/author/651b4c6f0bccda001fd6cc74

    Desired outcome 
        https://www.artsy.net/articles/author/maxwell-rabb

QA STEPS
    Test that old ID-based URLs redirect correctly to the new name-based URLs.
    Confirm that new URLs display correct content and metadata.
    Validate that internal links, sitemaps, and canonical tags point to the new URLs.

### DIFF
--- a/scripts/backfill_author_slugs.js
+++ b/scripts/backfill_author_slugs.js
@@ -88,7 +88,7 @@ const backfillAuthorSlugs = async () => {
 
 backfillAuthorSlugs()
   .then(() => {
-    console.log("\nMigration successful!!")
+    console.log("\nMigration successful!")
     process.exit(0)
   })
   .catch(error => {


src/api/apps/authors/test/model.test.coffee
```diff
@@ -45,10 +45,11 @@ describe 'Author', ->
           author._id.toString().should.equal '5086df098523e60002000018'
           done()
 
-    it 'finds the author by name', (done) ->
-      fabricate 'authors', { name: 'Kana Abe' }, ->
-        Author.find 'Kana Abe', (err, author) ->
```
iskounen, line 50: **issue:** I think we should keep this test and add a new one for `finds the author by slug'`.

src/api/apps/authors/model.coffee
```diff
@@ -76,14 +78,48 @@ Joi = require '../../lib/joi'
...
-    data = extend omit(input, 'id'),
-      _id: input.id
-    if data._id
-      db.collection("authors").updateOne { _id: data._id }, { $set: data }, (err, res) ->
-        db.collection("authors").findOne { _id: data._id }, callback
-    else
-      db.collection("authors").insertOne data, (err, res) ->
-        db.collection("authors").findOne { _id: res.insertedId }, callback
+
+    generateSlug = (name, existingSlug, cb) =>
```
iskounen, line 82: **question:** I'm not familiar with CoffeeScript, but it looks like we're declaring a function named `generateSlug` every time the `@save` handler runs, and then calling it on the new object. Is that correct?

If so, we should declare this outside of the handler.

src/api/apps/authors/model.coffee
```diff
@@ -76,14 +78,48 @@ Joi = require '../../lib/joi'
...
+          counter = 1
+          checkUniqueSlug = (slugWithCounter) ->
+            uniqueQuery = { slug: slugWithCounter }
+            uniqueQuery._id = { $ne: input.id } if input.id
+
+            db.collection("authors").findOne uniqueQuery, (err, found) ->
+              return cb(err) if err
+              if found
+                counter++
+                checkUniqueSlug("#{baseSlug}-#{counter}")
```
iskounen, line 102: **issue:** It looks like this is recursively `calling checkUniqueSlug` (and making a new database call each time) if a match is found. Would it be possible to do this in one go by searching for authors where the slug is `LIKE baseSlug%` and finding the last one in the list?

scripts/backfill_author_slugs.js
```diff
@@ -0,0 +1,132 @@
+/**
+ * Migration script to populate slugs for all existing authors
```
iskounen, line 2: **suggestion:** Could we just call `@save` on all existing authors and have that method handle generating a new slug?

PR #3243, opened 2025-10-07
refactor: generalize search naming

mc-jones:


mc-jones:
I will likely ignore the codecov failure. We may want to get rid of this.

PR #3296, opened 2026-03-23
fix(dx): update webpack config, add some docs

anandaroop:
[DI-59](https://www.notion.so/artsy/positron-Fix-webpack-src-targeting-32ccab0764a0802fb731f9cae716f352) and [DI-61](https://www.notion.so/artsy/positron-Document-or-even-default-to-staging-env-setup-for-local-development-32ccab0764a08052bb84c165b484d52c)

Fixes the webpack config for users (like me) who keep their Artsy repos in a path like `~/src/artsy`. 

This fixes a false match on that `src` string which was causing Webpack to overeagerly try to compile _everything_ in the positron directory — including <code>\~/src/artsy/positron/node_modules</code> — instead of just the stuff in <code>\~/src/artsy/positron/<b>src</b></code> i.e. that last subdirectory.

Also adds some instructions on how to properly connect to the staging env in order to be able to work with realistic data.

## Before

before

## After (local env)

after

## After (staging env)

after staging

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,15 @@ yarn start
 
 #### Using staging database
 
-In order to write articles, you will need to be a member of a channel. If you are an Artsy dev, you can point MONGOHQ_URL env to the staging database. Connecting to staging database requires VPN, please see details on [setting up a VPN connection here](https://github.com/artsy/infrastructure/blob/master/README.md#vpn).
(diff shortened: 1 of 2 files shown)


anandaroop:
cc @dblandin I just merged this, in case it helps with your issues booting up Positron.

PR #3306, opened 2026-05-13
fix(DI-283): detect user channel_ids drift and refresh them

iskounen:
This PR adds a check for drift between a user's (cached) channel IDs and the (source of truth) user IDs registered to each channel, and updates the former to match the latter if necessary.

Stale cached channel IDs on user records were causing an issue where some users would login to Writer and be presented with an error message like this one:

Screenshot 2026-05-13 at 5 09 31 PM

This (correctly) happens when a user does not have the `editorial` role and no channel IDs. It also (incorrectly) happens after they have been assigned channel IDs since the cached channel IDs on their user account is still empty. This check fixes that second scenario by actively checking for drift between the cached and actual channel IDs on each request that calls `fromAccessToken` via the `setUser` middleware.

- [DI-283](https://www.notion.so/artsy/Positron-channel-user-management-broken-user-adds-permissions-are-incorrect-35fcab0764a080fb9e34cd1522926a7a?v=310cab0764a08156aa73000cda8f267f&source=copy_link)
- cc: @artsy/diamond-devs

### DIFF
--- a/src/api/apps/users/model.coffee
+++ b/src/api/apps/users/model.coffee
@@ -29,7 +29,18 @@ jwtDecode = require 'jwt-decode'
         # Compare the user's partner access from the database vs JWT token to see if it needs a refresh
         savedPartnerIds = user.partner_ids.map((a) => a.toString())
         latestPartnerIds = jwtDecode(accessToken)?.partner_ids or []
-        return callback null, user if _.isEqual(savedPartnerIds, latestPartnerIds)
+        if _.isEqual(savedPartnerIds, latestPartnerIds)
+          # Also re-check channel membership in case access was granted since last login
+          return db.collection('channels').find({user_ids: new ObjectId(user._id)}).toArray (err, channels) ->
+            return callback err if err
+            latestChannelIds = _.pluck channels, '_id'
+            savedChannelIdsSorted = user.channel_ids.map((a) -> a.toString()).sort()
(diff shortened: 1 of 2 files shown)


src/api/apps/users/model.coffee
```diff
@@ -29,7 +29,18 @@ jwtDecode = require 'jwt-decode'
...
-        return callback null, user if _.isEqual(savedPartnerIds, latestPartnerIds)
+        if _.isEqual(savedPartnerIds, latestPartnerIds)
+          # Also re-check channel membership in case access was granted since last login
+          return db.collection('channels').find({user_ids: new ObjectId(user._id)}).toArray (err, channels) ->
+            return callback err if err
+            latestChannelIds = _.pluck channels, '_id'
+            savedChannelIdsSorted = user.channel_ids.map((a) -> a.toString()).sort()
+            latestChannelIdsSorted = latestChannelIds.map((a) -> a.toString()).sort()
+            return callback null, user if _.isEqual(savedChannelIdsSorted, latestChannelIdsSorted)
+            db.collection('users').updateOne { _id: user._id }, { $set: { channel_ids: latestChannelIds } }, (err) ->
```
iskounen, line 40: I decided to keep this instead of falling through to the `refresh` because it does significantly less work.

iskounen:
I am going to squash and merge this despite the failing codecov check because I think I did a good job with test coverage.

joeyAghion:
Thanks!

PR #3313, opened 2026-07-06
fix: reject non-http(s) protocols in article link sanitization

narikazu:
## What & Why

`sanitizeLink` (run on every article save) rewrote `artsy.net` URLs but accepted **any** URL protocol, so `javascript:`, `data:`, and `vbscript:` URLs were stored verbatim on article media sections and body links. When rendered (embed/video iframe `src`, or `<a href>`), these can execute script — a stored XSS.

Reported via the security bounty program as **GF658**.

## Change

Add a protocol allowlist to `sanitizeLink`: only `http:`, `https:`, `mailto:`, `tel:` are permitted; anything else is neutralized (returns `undefined`, so the field is stored empty).

- `http`/`https` — the normal case for embeds, media, and links
- `mailto`/`tel` — legitimately used in article **body** anchor links; kept to avoid a regression
- `javascript`/`data`/`vbscript`/etc. — rejected

Applies to all four `sanitizeLink` call sites in `save.coffee`: `video` / `social_embed` / `embed` section URLs, slideshow video items, `news_source.url`, and `<a href>`s inside text bodies.

## Testing

- Unit tests (`sanitize.test.ts`): reject `javascript:` (incl. mixed-case), `data:`, `vbscript:`; still allow `mailto:`/`tel:` and all existing cases.
- Save-path spec (`save.spec.ts`): a malicious `embed`/`video` section URL is stripped end-to-end through `sanitizeAndSave`; a valid YouTube URL is preserved.

### API testing

```bash
POSITRON="https://stagingwriter.artsy.net" # or localhost:3005
TOKEN="<fresh x-access-token>"          # from a logged-in stagingwriter session
CHANNEL="5759e4f3b5989e6f98f77998"      # a channel you have access to

# 1) Create a throwaway article with a malicious embed + a valid control
RESP=$(curl -s -X POST "$POSITRON/api/articles" \
  -H "x-access-token: $TOKEN" -H 'Content-Type: application/json' \
  -d '{
    "channel_id": "'"$CHANNEL"'",
    "title": "XSS sanitize test - DELETE ME",
    "sections": [
      { "type": "embed", "url": "javascript:alert(1)" },
      { "type": "embed", "url": "data:text/html,<script>alert(1)</script>" },
      { "type": "embed", "url": "https://www.youtube.com/watch?v=QWtsV50_-p4" }
    ]
  }')
echo "$RESP" | python3 -m json.tool
ID=$(echo "$RESP" | python3 -c "import sys,json;print(json.load(sys.stdin)['id'])")

# 2) Re-read what was stored
curl -s "$POSITRON/api/articles/$ID" -H "x-access-token: $TOKEN" | python3 -m json.tool

# 3) Clean up
curl -s -X DELETE "$POSITRON/api/articles/$ID" -H "x-access-token: $TOKEN"
```

## Migration / existing data

The fix guards **new** saves only. I ran a read-only audit against staging: **0 existing records use a disallowed protocol**, so no stored article is broken or newly-stripped by this change.

## Ticket

https://artsyproduct.atlassian.net/browse/PHIRE-3245

Assisted-by: Claude:Opus-4.8 [claude-code]

### DIFF
--- a/src/api/apps/articles/model/sanitize.js
+++ b/src/api/apps/articles/model/sanitize.js
@@ -44,10 +44,13 @@ export const sanitizeLink = urlString => {
   return url.href
 }
 
-const EMBED_ID_PATTERNS = {
-  vimeo: /^[0-9]+$/,
-  youtube: /^[A-Za-z0-9_-]+$/,
-}
+// A video id is interpolated (unescaped) into an iframe src="..." by metaphysics
+// extractEmbed on render. These characters would let a crafted id break out of
+// that attribute and inject markup/handlers; real youtube/vimeo ids never
+// contain them. We reject only on these, so legitimate but non-standard video
+// URLs (youtu.be links with ?t=/?si=, vimeo user/showcase/private urls, trailing
+// slashes) are preserved rather than destroyed. See security bounty GF658.
+const UNSAFE_EMBED_ID = /["'<>`\s]/
 
 const detectEmbedProvider = hostname => {
   if (hostname.includes("vimeo.com")) return "vimeo"
@@ -76,7 +79,7 @@ export const sanitizeEmbedUrl = urlString => {
     return sanitized
   }
   const id = detectEmbedId(url, provider)
-  if (typeof id !== "string" || !EMBED_ID_PATTERNS[provider].test(id)) {
+  if (typeof id === "string" && UNSAFE_EMBED_ID.test(id)) {
     return
   }
   return sanitized
--- a/src/api/apps/articles/model/sanitize.js
+++ b/src/api/apps/articles/model/sanitize.js
@@ -1,6 +1,11 @@
 // @ts-check
 import { URL } from "url"
 
+// Protocols permitted in article links. Anything else (e.g. javascript:, data:,
+// vbscript:) is rejected so it cannot be rendered as an iframe src / anchor href
+// on article pages, which would allow stored XSS. See security bounty GF658.
+const ALLOWED_PROTOCOLS = ["http:", "https:", "mailto:", "tel:"]
+
 // FIXME: cannot use typescript in yarn task commands
 export const sanitizeLink = urlString => {
   let url
@@ -12,6 +17,9 @@ export const sanitizeLink = urlString => {
     url = new URL(urlString)
   } catch (_e) {
     url = new URL(`https://${urlString}`)
+  }
+  if (!ALLOWED_PROTOCOLS.includes(url.protocol)) {
+    return
   }
   if (url.hostname === "artsy.net") {
     url.hostname = "www.artsy.net"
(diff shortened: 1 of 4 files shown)


narikazu:
Tagged Daimond team to this PR since the team is responsible for this project.

brainbicycle (approved):
this looks reasonable to me, thanks!

rabidpaul:
Thank you @narikazu !

anandaroop (approved):
Thanks!

anandaroop:
@narikazu ok if we merge this? 

I'd like to clear the deck of pending prod changes, so that I can use staging to test out another high-priority fix.

narikazu:
@anandaroop 

I pushed a new commit after your approval, so this needs another quick look before merge. If the PR looks good to you, you can merge this into the main. 

Why: Before pushing, I ran the new sanitization against the production article dataset (~35k articles) as a pre-deploy safety check. The original approach — requiring a strict YouTube/Vimeo id shape (`[A-Za-z0-9_-] / [0-9]+`) — would have stripped ~23 legitimate video URLs on their next save, i.e. real data loss. Examples it wrongly caught:

- `youtu.be/<id>?t=4, ?si=…, ?list=…` — valid YouTube links with query params
- `vimeo.com/<id>/<hash>` — valid private/unlisted Vimeo links
- `vimeo.com/<user>/<slug>` — valid Vimeo showcase/portfolio links

These contain no breakout characters; the strict validator just mis-extracts their id (grabs a hash/slug, or misses the path id when a query string is present) and rejected them.

What changed (commit ff0f19fb): sanitizeEmbedUrl now rejects a video URL only when the extracted id contains an iframe-src breakout character (", ', <, >, backtick, whitespace) instead of demanding a strict id shape. The actual stored-XSS still needs a " (via the decoded YouTube v param) to escape src="…", so this keeps the security property while preserving legitimate videos.

Verification:
- Re-ran the production check: all 18 previously-flagged legitimate video URLs are now preserved; the only remaining strips are genuinely-malicious javascript: hrefs.
- All XSS payloads (?v="…onload…", javascript:, data:) still stripped.
- Added regression tests for the exact production URL shapes above; full suite + type-check green.

Sorry for the re-review — I'd rather catch the data-loss risk now than in prod. 🙏

anandaroop:
Thanks for the manual verification and resulting fix. It seems good to me 👍🏽 

I'll go ahead and merge.

anandaroop (approved):

anandaroop:
_Seems_ like saving an article with a YouTube video embed is working as expected ✅ 

Screenshot 2026-07-09 at 4 21 38 PM

mc-jones:
@narikazu - I'm never going to argue against sanitizing inputs, but I'm wondering if this should have been a valid security bounty?

Wouldn't this only be allowed by those with writer access and therefore our own employees?

narikazu:
@mc-jones 

Fair question, and thank you for your input! I'd separate it into validity vs severity.

Validity: it's stored cross-user XSS, not self-XSS. A writer plants the payload and other users or staff trigger it, so the attacker and the victim are different people. Writer access lowers the likelihood but doesn't remove the flaw: a compromised writer account puts an external attacker on the inside, and since the repo is public, they can read the code and craft a guaranteed payload. So I'd say it's a valid entry.

Severity: this is where "internal only" counts. If this were external facing it would be a P1, but the writer-access requirement drops it to P3, which is what I've rated it and what I'm paying out at.

Does that make sense to you?

mc-jones:
Makes sense, but I don't agree it raises to `p3`. But your bounty, your call.

PR #3320, opened 2026-08-25
fix: Add the new filter param to Joi schema list

jpotts244:
Fast follow from: https://github.com/artsy/positron/commit/04ecc4cd1fd9491e4443db88fc1c683ba501a00c

Not having the new filter param in this allow list was removing it before we made it to the database
`Line 32: { stripUnknown: true }` — Joi validates the input against querySchema and strips any unknown fields before passing to toQuery.


jpotts244:
I'm here actively testing! Will revert if this isnt the proper fix but seeing lots of `fix` PRs of the past with this type of hot fix

jpotts244:
The filter works!! ✅

PR #3321, opened 2026-08-27
feat: Sends published_at timestamp in article publish event payload

jpotts244:
feat: Sends `published_at` timestamp in article publish event payload

Addressing: https://artsyproduct.atlassian.net/browse/TOP-403

Adding the `published_at` timestamp of the article being published over the Gravity as part of the event payload

We want to ingest this timestamp on Gravity's side vs infer it when the event gets processed

- https://github.com/artsy/gravity/pull/20480

### DIFF
--- a/src/api/apps/articles/test/model/index/persistence.spec.ts
+++ b/src/api/apps/articles/test/model/index/persistence.spec.ts
@@ -298,12 +298,14 @@ describe("Article Persistence", () => {
       ))
 
     it("sends RabbitMQ event when publishing", done => {
+      const published_at = new Date("2024-01-15T12:00:00.000Z")
       Article.save(
         {
           title: "Top Ten Shows",
           thumbnail_title: "Ten Shows",
           author_id: "5086df098523e60002000018",
           published: true,
+          published_at,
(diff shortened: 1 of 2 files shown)


src/api/apps/articles/test/model/index/persistence.spec.ts
```diff
@@ -321,6 +321,7 @@ describe("Article Persistence", () => {
               title: "Top Ten Shows",
               featured_artist_ids: [new ObjectId("52868347b202a37bb000072a")],
               href: "/article/undefined-ten-shows",
+              published_at: sinon.match.instanceOf(Date),
```
anandaroop: **thought (non-blocking):** Minor, but could maybe assert for a specific date based on the `published_at` attribute, rather than just any ol date here.

anandaroop (approved):
Works for me, one lil non-blocking thought…

src/api/apps/articles/test/model/index/persistence.spec.ts
jpotts244: yes that is much better 😄

